PR #4, opened 2018-05-16
Migrate cmd/ifqld  to platform

aanthony1243:
lots of file movement, minor code changes to resolve errors.  cmd/ifqld/main.go  still has many errors.


jsternberg:
This looks good to me so far.

query/query_test/query_test_pkg.go~
```diff
@@ -0,0 +1 @@
+package query
```
nathanielc: Can you remove this file?

query/query_test/query_test_pkg.go
```diff
@@ -0,0 +1,2 @@
+// no executable code in this directory, which contains tests for the query package
+package query
```
nathanielc: What is the purpose of this file?

aanthony1243: @desa   requested this change due to a build requirement that all directories contain at least one non-test go file.   An alternative would be to re-org the test dir so that it's one level up

jsternberg: I agree with moving this package, but I think we should delay anything that involves refactoring or reorganizing the directory structure until this is merged.

Gopkg.lock
```diff
@@ -1,24 +1,115 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
```
nathanielc: What is this a dep of?

Gopkg.lock
```diff
@@ -1,24 +1,115 @@
...
+  name = "github.com/RoaringBitmap/roaring"
+  packages = ["."]
+  revision = "1a28a7fa985680f9f4e1644c0a857ec359a444b0"
+  version = "v0.4.7"
+
+[[projects]]
+  name = "github.com/Shopify/sarama"
```
nathanielc: This doesn't look right. We should not have a dep on Kafka in this project.

Gopkg.lock
```diff
@@ -1,24 +1,115 @@
...
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
+[[projects]]
+  name = "github.com/cespare/xxhash"
+  packages = ["."]
+  revision = "5c37fe3735342a2e0d01c87a907579987c8936cc"
+  version = "v1.0.0"
+
+[[projects]]
+  name = "github.com/coreos/etcd"
```
nathanielc: This also looks suspicious. We should not have a dep on etcd in this project.

cmd/ifqld/README.md
```diff
@@ -0,0 +1,26 @@
+# IFQL
+
+This process provides an implementation of the idpe.Query interface via the network.
+
+# Exposed Metrics
+
+The `ifqlde` process exposes a Prometheus endpoint on port `8093` by default.
```
nathanielc: Update name here

cmd/ifqld/main.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+var (
+	bindAddr         string
+	concurrencyQuota int
+	memoryBytesQuota int
+)
+
+func init() {
+	viper.SetEnvPrefix("IFQLDE")
```
nathanielc: Update ifqlde -> ifqd throughout this file. Or should we make these changes in a separate commit?

cmd/ifqld/main.go
```diff
@@ -0,0 +1,225 @@
...
+		//	Hosts:   hosts,
+		//	Timeout: etcdTimeout,
+		//}
+		//o := etcd.NewOrganizationCache(etcdConfig)
+		//err = o.Open()
+		//if err != nil {
+		//	logger.Error("failed to open etc organization cache", zap.Error(err))
+		//	os.Exit(1)
+		//}
+		//orgSvc = o
```
nathanielc: Let's remove the bits about etcd here and just leave the first line of the comment.

cmd/ifqld/main.go
```diff
@@ -0,0 +1,225 @@
...
+			},
+		}
+		hosts, err := bucketDiscover.Hosts()
+		logger.Info("connecting to bucket etcd hosts", zap.Strings("hosts", hosts))
+		if err != nil {
+			return errors.Wrap(err, "failed to discover bucket etcd hosts")
+		}
+
+		// TODO(adam): figure correct bucket service
+		//etcdConfig := etcd.Config{
```
nathanielc: Same here

query/query_test/query_test_pkg.go
nathanielc: yes, lets delay the change until after this is merged.

query/query_test/query_test_pkg.go~
nathanielc: This still needs to be removed.

cmd/ifqld/main.go
nathanielc: I am fine if we make these rename changes in a follow up commit.

nathanielc:
Also, lets make sure to rebase and squash this commit so its a clean commit without the intermediate commits with the extra deps etc.

cmd/ifqld/main.go
aanthony1243: only came up in these 2 places so I made the change.

PR #5, opened 2018-05-16
line protocol encoder

docmerlin:
An encoder for the line protocol

### DIFF
--- a/line-protocol/metric.go
+++ b/line-protocol/metric.go
@@ -1,0 +1,73 @@
+package protocol
+
+import "time"
+
+// Tag holds the keys and values for a bunch of Tag k/v pairs
+type Tag struct {
+	Key   string
+	Value string
+}
+
+// Field holds the keys and values for a bunch of Metric Field k/v pairs
+type Field struct {
+	Key   string
+	Value interface{}
+}
+
+// Metric is the interface for marshaling, if you implement this interface you can be marshalled into the line protocol.  Woot!
+type Metric interface {
+	Time() time.Time
+	Name() string
+	TagList() []*Tag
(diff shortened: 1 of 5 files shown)


line-protocol/metric.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// Field holds the keys and values for a bunch of Metric Field k/v pairs
+type Field struct {
+	Key   string
+	Value interface{}
+}
+
+// Metric is the interface for marshaling, if you implement this interface you can be marshalled into the line protocol.  Woot!
+type Metric interface {
```
desa: Would be great if I didn't have to implement this interface to be able to write metrics.

desa: Maybe something using struct tags?

docmerlin: We were trying to avoid having to do a lot of reflection, which is why we went with this.

docmerlin: Here's the original PR.
https://github.com/influxdata/line-protocol/pull/1

line-protocol/metric.go
```diff
@@ -0,0 +1,73 @@
...
+// Tag holds the keys and values for a bunch of Tag k/v pairs
+type Tag struct {
+	Key   string
+	Value string
+}
+
+// Field holds the keys and values for a bunch of Metric Field k/v pairs
+type Field struct {
+	Key   string
+	Value interface{}
```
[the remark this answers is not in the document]
docmerlin: It would be pretty easy to make one as a wrapper around the Encoder.

docmerlin:
Going to close this.

PR #15, opened 2018-05-22
toHTTP function moved from github.com/influxdata/ifql

docmerlin:
toHTTP function moved from github.com/influxdata/ifql
https://github.com/influxdata/ifql/pull/362

### DIFF
--- a/query/functions/to_http.go
+++ b/query/functions/to_http.go
@@ -1,0 +1,430 @@
+package functions
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"runtime"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/influxdata/line-protocol"
+	"github.com/influxdata/platform/query"
+	"github.com/influxdata/platform/query/execute"
+	"github.com/influxdata/platform/query/plan"
+	"github.com/influxdata/platform/query/semantic"
+	"github.com/pkg/errors"
+)
+
+const (
+	ToHTTPKind           = "toHTTP"
+	DefaultToHTTPTimeout = 1 * time.Second
+)
+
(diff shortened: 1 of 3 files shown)


pauldix:
I'd prefer this be renamed to simply `to`. It's hitting the InfluxDB API so it's not just a general HTTP endpoint.

From a user standpoint, I'd prefer that `to` has `addr` as an optional. If it's not provided then the to should be this InfluxDB server or cluster.

Also, would prefer that param name be `host` since `addr` is a Go idiom that I haven't seen elsewhere.

docmerlin:
We were thinking multiple `toFoo` functions, and a `to` function that wrapped them and picked a sane one for the situation when a person just wanted to send data back to their own influx server, instead of doing something else with it.

query/functions/to_http.go
```diff
@@ -0,0 +1,430 @@
...
+	isTag := make([]bool, len(colMetadatas))
+	isValue := make([]bool, len(colMetadatas))
+
+	for i, col := range colMetadatas {
+		isValue[i] = sort.SearchStrings(t.spec.Spec.ValueColumns, col.Label) < len(t.spec.Spec.ValueColumns) && t.spec.Spec.ValueColumns[sort.SearchStrings(t.spec.Spec.ValueColumns, col.Label)] == col.Label
+		isTag[i] = sort.SearchStrings(t.spec.Spec.TagColumns, col.Label) < len(t.spec.Spec.TagColumns) && t.spec.Spec.TagColumns[sort.SearchStrings(t.spec.Spec.TagColumns, col.Label)] == col.Label
+	}
+
+	var err error
+	go func() {
```
nathanielc: To your comment above.

A block is only valid for the duration of the Process call. It is fine to read it asynchronously with making the HTTP request. But you need to guarantee that it is not read once the Process function returns.

docmerlin: Thanks.  Added a wg.

PR #17, opened 2018-05-22
chore: Add Makefile

nathanielc:
Adds top level Makefile which calls into query Makefiles.

 This Makefile follows a few conventions:

* All cmds must be added to this top level Makefile.
* All binaries are placed in ./bin, its recommended to add this directory to your PATH.
* Each package that has a need to run go generate, must have its own Makefile for that purpose.
* All recursive Makefiles must support the targets: all and clean.

Examples:

```
make # build all binaries ensuring go generate is run as needed
make test # build all binaries ensuring go generate is run as needed, then run all tests
make test-race # same as above except with race detection
make clean # clean all artifacts of the build
```


query/parser/ifql.go
```diff
@@ -6961,6 +6961,10 @@ var (
 	// errNoRule is returned when the grammar to parse has no rule.
 	errNoRule = errors.New("grammar has no rule")
 
+	// errInvalidEntrypoint is returned when the specified entrypoint rule
+	// does not exit.
+	errInvalidEntrypoint = errors.New("invalid entrypoint")
```
goller: PR aside for another time... these errors could go into the top-level errors package.  Those errors have a string and a reference code.  In the future, our docs folks can write extended documentation based on those reference codes.

nathanielc: This is autogenerated code, the move from the ifql repo inadvertently update the pigeon command dep so this generated code got update once I fixed the Makefiles.

Gopkg.toml
```diff
@@ -33,6 +35,11 @@
   name = "github.com/influxdata/influxdb"
   branch = "master"
 
+
+[[override]]
```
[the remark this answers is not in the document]
nathanielc: Will do.

PR #19, opened 2018-05-22
fix(query): Add _stop, _start to partition key

nathanielc:


pauldix:
Should this be `_start` and `_end`? Should be consistent with the `range` function arguments.

nathanielc:
Range function arguments are `_start` `_stop`.

PR #23, opened 2018-05-22
Create nightly builds of ifqld

nathanielc:
The builds are really anytime a commit is made to master not nightly.
Builds use goreleaser with the snapshot setting.


nathanielc:
I tested that the deploy is working from within CircleCI. Now deploy will only run on commits to master.

PR #25, opened 2018-05-22
fix and enable vet

docmerlin:
Fix and enable vet.


.circleci/config.yml
```diff
@@ -9,10 +9,8 @@ jobs:
     steps:
       - checkout
       - run: make test
-
+      - run: go vet -v ./...
```
nathanielc: Do you want to add a `make vet` target that does this and call `make vet`? It will probably be just like `make test`

This way the Makefile remains the source of truth on what is runs in the CI system and can easily be replicated locally.

docmerlin: sure

Makefile
```diff
@@ -87,6 +87,9 @@ test: all
 test-race: all
 	$(GO_TEST) -race ./...
 
+vet: all
+	go vet -v ./...
```
nathanielc: Can you define a `GO_VET` var like `GO_TEST`?

PR #28, opened 2018-05-22
feat(http): add common code for running an http server

jsternberg:
It handles graceful termination by listening to signals which will be
particularly useful for commands where the sole purpose is to run an
http server.

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -1,0 +1,105 @@
+package http
+
+import (
+	"context"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+)
+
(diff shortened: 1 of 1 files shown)


http/server.go
```diff
@@ -0,0 +1,99 @@
...
+			errCh <- err
+		}
+		close(errCh)
+	}()
+	return errCh
+}
+
+func (s *Server) shutdown() error {
+	// The shutdown needs to succeed in 20 seconds or less.
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
goller: What do you think about putting the timeout duration as a public field in the struct?  The NewServer function could set the default to 20 seconds

jsternberg: Done. We can also update this struct as we need to with additional features. It'll be useful though to set a baseline with useful functionality though so we want to update this struct.

PR #31, opened 2018-05-23
feat(http): add a /ping endpoint to the transpiler

### LINKED ISSUE #123: Make transpilerd support the /ping endpoint

jsternberg:
This will match the /ping endpoint from the influxdb 1.x httpd service.
This endpoint is used by many of the clients that use the query service
and it will also be helpful for testing that the transpiler service is
up and working.

In the future, we will add the appropriate health endpoint that is
standard for 2.0. But, for the best compatibility, we have to return
something viable for 1.x clients.

Fixes #123.

### DIFF
--- a/http/transpiler_handler.go
+++ b/http/transpiler_handler.go
@@ -28,8 +28,16 @@ func NewTranspilerQueryHandler(orgID platform.ID) *TranspilerQueryHandler {
 		Logger: zap.NewNop(),
 	}
 
+	h.HandlerFunc("GET", "/ping", h.servePing)
 	h.HandlerFunc("POST", "/query", h.handlePostQuery)
 	return h
+}
+
+// servePing returns a simple response to let the client know the server is running.
+// This handler is only available for 1.x compatibility and is not part of the 2.0 platform.
+func (h *TranspilerQueryHandler) servePing(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusNoContent)
 }
 
 // handlePostInfluxQL handles query requests mirroring the 1.x influxdb API.


pauldix:
This should actually follow what other services in the k8s environment do. Should have a health endpoint and a readiness endpoint. @gianarb or @goller, do we have the standard for this documented?

jsternberg:
@pauldix the transpiler will probably be different. We will need a `/ping` endpoint. If we add additional endpoints for the k8s environment, that's fine. But I think we need to implement as a frontend everything that the `influx` client from 1.x believes should exist for a smooth transition.

I'm not sure if I'm understanding that correctly since that would probably involve also implementing the `/write` endpoint at some point (although we could declare that not to work), but the `/ping` endpoint is probably needed.

Are we doing the health endpoints for platform? I don't think those have been migrated over yet.

goller:
The healthz stuff is here: https://github.com/influxdata/idpe/blob/e149594e81bbbca28c8c0f9eda3e9f36d3c170a8/kit/healthz/healthz.go

and... gian has a blog that would need to be added to docs: https://gianarb.it/blog/services-and-common-parts

goller:
@jsternberg chronograf uses it here https://github.com/influxdata/chronograf/blob/e8f80a15acbcbc8555831064b84e66c4708b1570/influx/influx.go#L186

... specifically to find the version of influx by grabbing `X-Influxdb-Build` and if that doesn't exist `X-Influxdb-Version`

pauldix:
I really don't think this is necessary. It's a health check to see if the system is up. Any proper health checking setup can be pointed elsewhere.

pauldix:
@goller isn't that really just to see if the enterprise functionality should be open? Chronograf 2.0 will have completely different Enterprise endpoints that it hits so that stuff will have to change. Should use a 2.0 endpoint for checking that version. The InfluxQL endpoint is the same regardless of which your'e hitting so I don't think it needs a version from that.

jsternberg:
I've updated this and clarified that this is for 1.x compatibility in the PR. It is not meant to serve as the 2.0 health endpoint which will be added later when we have migrated that stuff to the open source repository.

PR #32, opened 2018-05-23
feat(http): perform error handling in the transpiler and the query service

jsternberg:
The query service now handles some errors (not all) from the ifqld
service. This way, the transpiler, which uses the query service, can
report when it encounters an error from ifqld.

The transpiler will now correctly return errors. In the future, it
should adapt the error to be returned in the same way as queries are
returned from influxdb 1.x.


http/query_service.go
```diff
@@ -190,6 +190,26 @@ func (s *QueryService) QueryWithCompile(ctx context.Context, orgID platform.ID,
...
 func (s *QueryService) processResponse(resp *http.Response) (query.ResultIterator, error) {
+	// Check for the status code. If we have a 5xx response, read the error code and
+	// return it.
+	switch resp.StatusCode / 100 {
+	case 4:
+		return nil, errors.Wrap(errors.New(resp.Status), "client error")
+	case 5:
+		// If we have a 5xx response, then read the error code from the header
+		// and return it if possible.
+		errMsg := resp.Header.Get("X-Influx-Error")
```
goller: We have helper functions here: https://github.com/influxdata/platform/blob/master/kit/errors/http.go

http/query_service.go
```diff
@@ -190,6 +190,26 @@ func (s *QueryService) QueryWithCompile(ctx context.Context, orgID platform.ID,
...
+		if errMsg != "" {
+			return nil, errors.New(errMsg)
+		}
+
+		// Otherwise, we need to report that we encountered an unknown server error.
+		// Maybe this isn't an ifql server or we encountered a panic?
+		return nil, errors.Wrap(errors.New(resp.Status), "unknown server error")
```
goller: for future PR these errors should also be metrics

jsternberg:
I created #36 and then rebased this on top of it so this uses the same error handling code as the other http services.

http/transpiler_handler.go
```diff
@@ -121,9 +121,12 @@ var influxqlCE = crossExecute{
 }
 
 func encodeResult(w http.ResponseWriter, results query.ResultIterator, contentType string, encoder query.MultiResultEncoder) error {
+	// Set the headers for the response. We do not have to write the ok header here.
+	// It will automatically be written when the encoder performs its first
+	// write to the io.Writer instance. If encoding fails, the encoder should not
+	// write anything to the io.Writer and we can then write our own error message.
```
nathanielc: > If encoding fails, the encoder should not write anything to the io.Writer

I don't think we can make this requirement of encoders. It is possible for an error to occur mid stream.

As such the only thing we can do is write an explicit `http.StatusOK` and then specify what happens when an error occurs mid stream. That is currently specified [here](https://github.com/influxdata/platform/blob/master/query/docs/SPEC.md#errors)

nathanielc: Also that part of the spec is not implemented yet :(

jsternberg: I can modify this to my real intention. If an encoder _returns an error_ then it has committed to writing a message with a success status and needs to report its error as part of its encoding format.

The point is that once it writes, it will be an `http.StatusOK` and it can't go back. An encoder could also check if its a response writer and issue the status ok by itself, but I didn't want to change the interface from `io.Writer` since using `io.Writer` will be useful for tests I assume and we can always do a type cast.

nathanielc: Yeah, we do not what the encoder writing http status codes.

jsternberg: I think I should clarify the main issue. If we write the status code here before calling the encoder, you get a double write of the status and the second status never gets written when an error happens in the encoder. So we need to only write the status ok when we know an error isn't going to happen. That's the reason why I removed it here.

I looked into it and the `http.ResponseWriter` is documented that when a write occurs without a `WriteHeader` call, it automatically writes `200 OK`. I figured that was the best way to do this without requiring the encoder to be HTTP-aware.

Another potential way is to have the encoder write to a buffer and then write the buffer if there is no error while encoding. That would result in doubling the amount of memory used though because it would have to maintain a buffer of the output. This would be a problem with the JSON encoder though because the JSON encoder already has to write out the entire response all at once because that's how the JSON gets encoded and then it would have to copy from that buffer to the response writer which would use up more memory.

But, that may be better for API and I'm not sure the memory implications are that bad. Either way, we can't call `WriteHeader` before `Encode` is called.

http/transpiler_handler.go
```diff
@@ -100,9 +103,10 @@ func (h *TranspilerQueryHandler) handlePostInfluxQL(w http.ResponseWriter, r *ht
 		return
 	}
 
-	err = encodeResult(w, results, ce.contentType, ce.encoder)
-	if err != nil {
-		kerrors.EncodeHTTP(ctx, err, w)
+	// Once we have reached this stage, it is the encoder's job to encode any
+	// errors in the encoded format.
+	if err := encodeResult(w, results, ce.contentType, ce.encoder); err != nil {
+		h.Logger.Info("Unable to encode result")
```
nathanielc: Should this be a logger.Error with the error passed in?

query/influxql/result.go
```diff
@@ -28,106 +26,103 @@ func (e *MultiResultEncoder) Encode(w io.Writer, results query.ResultIterator) e
...
+					case query.TBool:
+						for j, v := range cr.Bools(idx) {
+							values[j][i] = v
+						}
+					case query.TTime:
+						for j, v := range cr.Times(idx) {
+							values[j][i] = v.Time().Format(time.RFC3339)
+						}
+					default:
+						// Unknown type. Leave it as nil.
```
nathanielc: This should return an error. It will be the transpiler's responsibility to ensure that the results only have columns that this encoder supports.

query/influxql/result.go
```diff
@@ -28,106 +26,103 @@ func (e *MultiResultEncoder) Encode(w io.Writer, results query.ResultIterator) e
...
 
-		err = blocks.Do(func(b query.Block) error {
-			r := NewRow()
+		if err := blocks.Do(func(b query.Block) error {
+			var r Row
 
-			fieldName := ""
-			measurementVaries := -1
-			for j, c := range b.Key().Cols() {
+			for i, c := range b.Key().Cols() {
```
nathanielc: This is a small nit, but since this kind of Go code uses index integers more that typical code the convention is to use `i` for row iteration and `j` for column iteration. It helps keep all the indexes straight.

query/influxql/result.go
```diff
@@ -28,106 +26,103 @@ func (e *MultiResultEncoder) Encode(w io.Writer, results query.ResultIterator) e
...
 	for results.More() {
-		name, r := results.Next()
+		_, r := results.Next()
 
 		blocks := r.Blocks()
-		nameInt, err := strconv.Atoi(name)
-		if err != nil {
-			return fmt.Errorf("error converting result name to integer: %s", name)
-		}
-		result := Result{StatementID: nameInt}
```
nathanielc: Why is the StatementID not being set anymore?

query/influxql/result_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/influxdata/platform/query/csv"
+	"github.com/influxdata/platform/query/influxql"
+)
+
+func TestMultiResultEncoder_Encode(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		in   string
```
nathanielc: I'd prefer not to rely on the csv implementation to ensure the influxql implementation is correct.

If the input data is a `query.ResultIterator` you can supply an erroring implementation. Look at this https://github.com/influxdata/platform/pull/42/files#diff-0b2c0b2c989e5288ceb7d6900199aa05R672

and 
https://github.com/influxdata/platform/pull/42/files#diff-ab5711a64e017c73139b7d3f3da647a1R22

query/influxql/result.go
jsternberg: I'll reverse it.

query/influxql/result.go
jsternberg: We don't have a meaningful spec for how to set the statement id so the previous code was just speculation. I'll put in a TODO so we know to come back to this when we know what the response should look like.

http/transpiler_handler.go
jsternberg: The error needs to be passed in. But, this isn't an `Error` level because it's unlikely to be actionable or indicate that the system is failing in some unrecoverable way. The only time this would fail is if the JSON failed to encode (very unlikely) or if it was unable to write to the response writer. The second would be a temporary error so this likely qualifies as info level.

query/influxql/result.go
```diff
@@ -9,129 +9,128 @@ import (
...
 
 // Encode writes a collection of results to the influxdb 1.X http response format.
-// Expectations/Assumptions:
-//  1.  Each result will be published as a 'statement' in the top-level list of results.  The 'staementID'
-//      will be interpreted as an integer, and will return an error otherwise.
-//  2.  If the _field name is present in the tags, and a _value column is present, the _value column will
-//      be renamed to the value of the _field tag
-//  3.  If the _measurement name is present in the tags, it will be used as the row.Name for all rows.
-//      Otherwise, we'll use the column value, which _must_ be present in that case.
-
```
nathanielc: Why is this comment being removed?

query/influxql/result.go
nathanielc: This is not fixed yet.

query/influxql/result.go
```diff
@@ -9,129 +9,128 @@ import (
...
-						if j == measurementVaries {
-							if c.Type != query.TString {
-								return errors.New("unexpected type, _measurement is not a string")
-							}
-							r.Name = cr.Strings(j)[i]
-							continue
-						}
+			if err := b.Do(func(cr query.ColReader) error {
+				var values [][]interface{}
+				j := 0
```
nathanielc: This should be `i` and the one above should be `j`.

query/influxql/result.go
```diff
@@ -9,129 +9,128 @@ import (
...
+				for idx, c := range b.Cols() {
```
nathanielc: And here `idx` should be `j`.

query/influxql/result.go
jsternberg: The comment itself wasn't correct. I'll update it instead of removing it.

query/influxql/result.go
jsternberg: `idx` and `j` aren't synonyms. The `idx` here refers to the real column id that's used to access values while `j` is equal to the column we are currently filling. This is because the partition key values aren't returned as part of the values so we have to skip over those columns which makes it so that `j` and `idx` become misaligned.

But `idx` isn't a very good name. Any recommendations? Maybe `off` for offset or `colIndex` for the column index?

query/influxql/result.go
jsternberg: This is the column index. There is no `i` variable in this scope yet.

nathanielc:
FWIW, I filed this issue to document these nits instead of waiting to bringing them up during a PR review. https://github.com/influxdata/platform/issues/45

PR #36, opened 2018-05-23
feat(http): add a utility function for reading http errors

jsternberg:
This also modifies the http error encoder to not include the reference
code in the `X-Influx-Error` message so it only includes the text.


desa:
I really like these changes :)

PR #43, opened 2018-05-24
fix(query): Utilize improvements storage RPC API

stuartcarnie:
* clarifies grouping behavior in data types, matching RPC
* distinct handles missing column
* reader operates in two modes, depending on whether it is using GroupFrame to determine partition keys
    * this could be refactored into a state machine
* fixes https://github.com/influxdata/ifql/issues/372 (issue in old ifql repo)


query/execute/format.go
```diff
@@ -117,7 +117,7 @@ func (f *Formatter) WriteTo(out io.Writer) (int64, error) {
 
 	// Write rows
 	r := 0
-	f.b.Do(func(cr query.ColReader) error {
+	w.err = f.b.Do(func(cr query.ColReader) error {
```
stuartcarnie: This is to ensure block errors propagate back (such as data type changed between blocks)

query/functions/group.go
```diff
@@ -233,14 +293,14 @@ func (t *groupTransformation) RetractBlock(id execute.DatasetID, key query.Parti
 func (t *groupTransformation) Process(id execute.DatasetID, b query.Block) error {
 	cols := b.Cols()
 	on := make(map[string]bool, len(cols))
-	if len(t.keys) > 0 {
+	if t.mode == GroupModeBy && len(t.keys) > 0 {
 		for _, k := range t.keys {
 			on[k] = true
 		}
-	} else if len(t.except) > 0 {
+	} else if t.mode == GroupModeExcept && len(t.keys) > 0 {
```
nathanielc: Where are the cases for All and None?

query/functions/group.go
```diff
@@ -69,9 +93,41 @@ func (s *GroupOpSpec) Kind() query.OperationKind {
...
+	return mode
+}
+
 type GroupProcedureSpec struct {
-	By     []string
-	Except []string
+	By        []string
+	Except    []string
+	GroupMode GroupMode
+	GroupKeys []string
```
nathanielc: What's the difference between By and GroupKeys?

stuartcarnie: I meant to remove it from that structure; will update

query/functions/group.go
stuartcarnie: These changes only replicate the previous behavior of 

* `len(t.keys) > 0` -> `t.mode == GroupModeBy && len(t.keys) > 0` AND
* `len(t.except) > 0` to `t.mode == GroupModeExcept && len(t.keys) > 0`

Are you suggesting there is an existing bug that should be addressed?

nathanielc: >Are you suggesting there is an existing bug that should be addressed?

Yes, but it can wait.

query/functions/group.go
nathanielc: Should we call the Except field `GroupExcept`?

FWIW, I prefer the non redundant names By, Except and Mode.

nathanielc: The latest changes make sense. We only have one list of keys, the group mode contains their meaning. +1

PR #49, opened 2018-05-25
add swagger for tasks API

jademcgough:


http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
...
+      responses:
+        '200':
+          description: a list of task results
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  results:
+                    type: array
```
goller: I think this should be an object like this:

```json
{
   "results": []
}

```

http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
...
+            type: string
+          required: true
+          description: ID of task to get logs for
+      responses:
+        '200':
+          description: all logs for a task
+          content:
+            application/json:
+              schema:
+                type: array
```
goller: Probably 
```json
{
   "logs": []
}
```

http/swagger.yml
```diff
@@ -316,6 +472,27 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/Bucket"
+    Link:
+      type: object
+      readOnly: true
+      description: URI of resource.
+      properties:
+        rel:
```
goller: I've gone back and forth on this link style over time.  In this case `rel` would be `next`, but, because it is in the links object as a key, I don't think it need to be in the `Link` object.

Following on, without a rel, I think the `Link` itself could just be a string with url format.

However, we could go the other way too.  `Links` could be an array rather than an object.  Each link would then need rel and href.

What do you think would be nicer to program against?

http/swagger.yml
```diff
@@ -326,6 +503,56 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/Organization"
+    Task:
+      properties:
+        id:
+          type: string
```
goller: I think this should be a readOnly property

http/swagger.yml
```diff
@@ -326,6 +503,56 @@ components:
...
+        name:
+          type: string
+        state:
+          type: string
```
goller: (same ... description and a description about what the updates do such as kill the currently running task)

http/swagger.yml
```diff
@@ -326,6 +503,56 @@ components:
...
+        startTime:
+          type: string
+          format: dateTime
+        endTime:
+          type: string
+          format: dateTime
+        error:
+          $ref: "#/components/schemas/Error"
+        log:
+          type: string
```
goller: I think of this being a link to a specific endpoint for this results logs.

http/swagger.yml
```diff
@@ -326,6 +503,56 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/Organization"
+    Task:
+      properties:
+        id:
+          type: string
+        name:
```
goller: Probably add a description and mention that this can be updated.

http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /tasks:
+    get:
```
goller: Would you add the `post` in here as well?

http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
...
+                      $ref: "#/components/schemas/TaskRecord"
+                  links:
+                    $ref: "#/components/schemas/Links"
+  '/tasks/{taskId}/logs':
+    get:
+      tags:
+        - Tasks
+      summary: Retrieve all logs for a task
+      parameters:
+        - in: path
```
goller: Brainstorming... What do you think about an optional parameter that specifies result id filter for the logs?

http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /tasks:
+    get:
+      tags:
+        - Tasks
+      summary: List all tasks
```
pauldix: Should have pagination by default. Return 100 tasks and then have the `Tasks` object have resource pointers for `previous` and `next`?

http/swagger.yml
```diff
@@ -201,6 +201,162 @@ paths:
...
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+  '/tasks/{taskId}/results':
+    get:
+      tags:
+        - Tasks
+      summary: Retrieve list of results for a task
```
pauldix: Results will also need a method for pagination. Also probably need an API endpoint to get the results in a specific time range. Like when investigating why the data seems off from yesterday at noon or something like that.

http/swagger.yml
jademcgough: yeah, I think I'll just drop rel from here

http/swagger.yml
jademcgough: Yep, I'm going to change all endpoints to be like this, for extensibility purposes.

http/swagger.yml

http/swagger.yml
jademcgough: Yep, this is how I envisioned it being used as well... I'll add a description for clarity.

http/swagger.yml
jademcgough: oh, I completely forgot to add that :|

http/swagger.yml
jademcgough: (chris and I discussed his comments earlier; I'll be adding this as well as some more filter options, like by user or by org)

http/swagger.yml
jademcgough: Yep, that's very reasonable.

jademcgough:
@goller can you take another look at this?

goller:
Looks great ! I think we need one last filter option for logs were one can specify the run id.  

I think the log link in the run object could be the link schema.

jademcgough:
Got it, thanks. I'll make those changes then squash/rebase/merge.

PR #55, opened 2018-05-25
fix(http): keep the http response body open until the result has been read

jsternberg:
This modifies the `MultiResultDecoder` interface to accept an
`io.ReadCloser` so the `ResultIterator` can close the `io.Reader`
instead of doing it through a defer call. It then makes it so the
`Cancel()` method will close the reader or the reader will be
automatically closed when `More()` returns false.

### DIFF
--- a/query/csv/result.go
+++ b/query/csv/result.go
@@ -75,7 +75,7 @@ func (d *ResultDecoder) Decode(r io.Reader) (query.Result, error) {
 // Results are delimited by an empty line.
 type MultiResultDecoder struct {
 	c ResultDecoderConfig
-	r io.Reader
+	r io.ReadCloser
 }
 
 // NewMultiResultDecoder creates a new MultiResultDecoder.
@@ -88,7 +88,7 @@ func NewMultiResultDecoder(c ResultDecoderConfig) *MultiResultDecoder {
 	}
 }
 
-func (d *MultiResultDecoder) Decode(r io.Reader) (query.ResultIterator, error) {
+func (d *MultiResultDecoder) Decode(r io.ReadCloser) (query.ResultIterator, error) {
 	return &resultIterator{
 		c: d.c,
 		r: r,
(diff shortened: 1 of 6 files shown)


query/csv/result.go
```diff
@@ -119,7 +126,8 @@ func (r *resultIterator) Next() query.Result {
 	return r.next
 }
 
-func (r *resultIterator) Cancel() {
+func (r *resultIterator) Cancel() error {
+	return r.r.Close()
```
nathanielc: We can't assume it is safe to call close twice, this should check it was already closed.

query/query.go
```diff
@@ -27,7 +27,7 @@ type ResultIterator interface {
 
 	// Cancel discards the remaining results.
 	// If not all results are going to be read, Cancel must be called to free resources.
-	Cancel()
+	Cancel() error
```
nathanielc: Let's not change cancel to return an error. 
It the action of canceling causes an error, we can report it with the `Err` method.

query/influxql/response.go
```diff
@@ -13,7 +13,10 @@ type Response struct {
 
 func (r *Response) error(err error) {
 	r.Results = nil
-	r.Err = err.Error()
+	if r.Err == "" {
```
nathanielc: How is it possible to get multiple errors?

jsternberg: It's possible for there to be an error and then `Cancel()` to report another error. I've moved this section of the code to the calling site instead since it makes more sense there.

query/query.go
jsternberg: Done.

query/csv/result.go
```diff
@@ -120,6 +125,9 @@ func (r *resultIterator) Next() query.Result {
 }
 
 func (r *resultIterator) Cancel() {
+	if err := r.r.Close(); err != nil && r.err == nil {
```
nathanielc: The user way still call cancel after More returns false, so we need to protect the call to Close

query/csv/result.go
jsternberg: Done.

query/csv/result.go
jsternberg: Sorry, I hadn't pushed that yet. I've fixed this section of code.

PR #56, opened 2018-05-25
feat(query/querytest): first batch of tests ready for review 

aanthony1243:
let's push more tests here, and merge when it's green


query/querytest/query_test.go
```diff
@@ -13,6 +13,14 @@ import (
 	"github.com/andreyvit/diff"
 )
 
+var skipTests = map[string]string{
+	"derivative":                "derivative not supported by influxql",
+	"filter_by_tags":            "arbitrary filtering not supported by influxql",
+	"group_ungroup":             "influxql/ifql disagreement on keycols",
+	"window":                    "ordering of results differs between queries",
+	"window_group_mean_ungroup": "error in influxql: failed to run query: timeValue column \"_start\" does not exist",
```
nathanielc: This looks great. Can you add issues for each of these failing tests? Then we can reference the issue here.

PR #60, opened 2018-05-29
feat(errors): add error lib

kelwang:
https://github.com/influxdata/idpe/issues/488

This is a global errors library. Ideally, we could make all of our generated errors have type and location, 
and wrap all the third party errors with references.

 Due the complexity of our current code base. We decide to separate this project to small PRs. After the library been pushed. We could use this lib for all the new errors, then gradually update our existing code structure.
- [x] Add the library
- [ ] Update the http and idp services, make sure they all return the typed error
- [ ] Update end user services
- [ ] Maybe add https://github.com/kisielk/errcheck to CI

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,0 +1,151 @@
+package platform
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Some error code constant, ideally we want define common platform codes here
+// projects on use platform's error, should have their own central place like this.
+const (
+	EInternal   = "internal error"
+	ENotFound   = "not found"
+	EConflict   = "conflict" // action cannot be performed
+	EInvalid    = "invalid"  // validation failed
+	EEmptyValue = "empty value"
+)
+
+// Error is the error struct of platform.
+//
+// Errors may have error codes, human-readable messages,
+// and a logical stack trace.
+//
+// The Code targets automated handlers so that recovery can occur.
+// Msg is used by the system operator to help diagnose and fix the problem.
+// Op and Err chain errors together in a logical stack trace to
+// further help operators.
+//
+// To create a simple error,
+//     &Error{
+//         Code:ENotFound,
+//     }
+// To show where the error happens, add Op.
+//     &Error{
+//         Code: ENotFound,
+//         Op: "bolt.FindUserByID"
+//     }
+// To show an error with a unpredictable value, add the value in Msg.
+//     &Error{
+//        Code: EConflict,
+//        Message: fmt.Sprintf("organization with name %s already exist", aName),
+//     }
+// To show an error wrapped with another error.
+//     &Error{
+//         Code:EInternal,
+//         Err: err,
+//     }.
+type Error struct {
+	Code string `json:"code"`          // Code is the machine-readable error code.
+	Msg  string `json:"msg,omitempty"` // Msg is a human-readable message.
+	Op   string `json:"op,omitempty"`  // Op describes the logical code operation during error.
+	Err  error  `json:"err,omitempty"` // Err is a stack of additional errors.
+}
+
+// Error implement the error interface by outputing the Code and Err.
+func (e Error) Error() string {
+	var b strings.Builder
+
+	// Print the current operation in our stack, if any.
+	if e.Op != "" {
+		fmt.Fprintf(&b, "%s: ", e.Op)
+	}
+
+	// If wrapping an error, print its Error() message.
+	// Otherwise print the error code & message.
+	if e.Err != nil {
+		b.WriteString(e.Err.Error())
+	} else {
+		if e.Code != "" {
+			fmt.Fprintf(&b, "<%s>", e.Code)
+			if e.Msg != "" {
+				b.WriteString(" ")
+			}
+		}
+		b.WriteString(e.Msg)
+	}
+	return b.String()
+}
+
+// ErrorCode returns the code of the root error, if available; otherwise returns EINTERNAL.
+func ErrorCode(err error) string {
+	if err == nil {
+		return ""
+	} else if e, ok := err.(*Error); ok && e.Code != "" {
+		return e.Code
+	}
+	return EInternal
+}
+
+// ErrorMessage returns the human-readable message of the error, if available.
+// Otherwise returns a generic error message.
+func ErrorMessage(err error) string {
+	if err == nil {
+		return ""
+	} else if e, ok := err.(*Error); ok && e.Msg != "" {
+		return e.Msg
+	} else if ok && e.Err != nil {
+		return ErrorMessage(e.Err)
+	}
+	return "An internal error has occurred."
+}
+
+// errEncode an JSON encoding helper that is needed to handle the recursive stack of errors.
+type errEncode struct {
+	Code string `json:"code"`          // Code is the machine-readable error code.
+	Msg  string `json:"msg,omitempty"` // Msg is a human-readable message.
+	Op   string `json:"op,omitempty"`  // Op describes the logical code operation during error.
+	Err  string `json:"err,omitempty"` // Err is a stack of additional errors.
+}
+
+// MarshalJSON recursively marshals the stack of Err.
+func (e *Error) MarshalJSON() (result []byte, err error) {
+	ee := errEncode{
+		Code: e.Code,
+		Msg:  e.Msg,
+		Op:   e.Op,
+	}
+	if e.Err != nil {
+		if _, ok := e.Err.(*Error); ok {
+			b, err := e.Err.(*Error).MarshalJSON()
+			if err != nil {
+				return result, err
+			}
+			ee.Err = string(b)
+		} else {
+			ee.Err = e.Err.Error()
+		}
+	}
+	return json.Marshal(ee)
+}
+
+// UnmarshalJSON recursively unmarshals the error stack.
+func (e *Error) UnmarshalJSON(b []byte) (err error) {
+	ee := new(errEncode)
+	err = json.Unmarshal(b, ee)
+	e.Code = ee.Code
+	e.Msg = ee.Msg
+	e.Op = ee.Op
+	if ee.Err != "" {
+		var innerErr error
+		innerResult := new(Error)
+		innerErr = innerResult.UnmarshalJSON([]byte(ee.Err))
+		if innerErr != nil {
+			e.Err = errors.New(ee.Err)
+			return err
+		}
+		e.Err = innerResult
+	}
+	return err
+}
--- a/errors_test.go
+++ b/errors_test.go
@@ -1,0 +1,206 @@
+package platform
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"testing"
+)
+
+const EFailedToGetStorageHost = "failed to get the storage host"
+
+func TestErrorMsg(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		msg  string
+	}{
+		{
+			name: "simple error",
+			err:  &Error{Code: ENotFound},
+			msg:  "<not found>",
+		},
+		{
+			name: "with op",
+			err: &Error{
+				Code: ENotFound,
+				Op:   "bolt.FindAuthorizationByID",
+			},
+			msg: "bolt.FindAuthorizationByID: <not found>",
+		},
+		{
+			name: "with op and value",
+			err: &Error{
+				Code: ENotFound,
+				Op:   "bolt.FindAuthorizationByID",
+				Msg:  fmt.Sprintf("with ID %d", 323),
+			},
+			msg: "bolt.FindAuthorizationByID: <not found> with ID 323",
+		},
+		{
+			name: "with a third party error",
+			err: &Error{
+				Code: EFailedToGetStorageHost,
+				Op:   "cmd/fluxd.injectDeps",
+				Err:  errors.New("empty value"),
+			},
+			msg: "cmd/fluxd.injectDeps: empty value",
+		},
+		{
+			name: "with a internal error",
+			err: &Error{
+				Code: EFailedToGetStorageHost,
+				Op:   "cmd/fluxd.injectDeps",
+				Err:  &Error{Code: EEmptyValue, Op: "cmd/fluxd.getStrList"},
+			},
+			msg: "cmd/fluxd.injectDeps: cmd/fluxd.getStrList: <empty value>",
+		},
+	}
+	for _, c := range cases {
+		if c.msg != c.err.Error() {
+			t.Fatalf("%s failed, want %s, got %s", c.name, c.msg, c.err.Error())
+		}
+	}
+}
+
+func TestErrorMessage(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want string
+	}{
+		{
+			name: "nil error",
+		},
+		{
+			name: "simple error",
+			err:  &Error{Msg: "simple error"},
+			want: "simple error",
+		},
+		{
+			name: "embeded error",
+			err:  &Error{Err: &Error{Msg: "embeded error"}},
+			want: "embeded error",
+		},
+		{
+			name: "default error",
+			err:  errors.New("s"),
+			want: "An internal error has occurred.",
+		},
+	}
+	for _, c := range cases {
+		if result := ErrorMessage(c.err); c.want != result {
+			t.Fatalf("%s failed, want %s, got %s", c.name, c.want, result)
+		}
+	}
+}
+func TestErrorCode(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want string
+	}{
+		{
+			name: "nil error",
+		},
+		{
+			name: "simple error",
+			err:  &Error{Code: ENotFound},
+			want: ENotFound,
+		},
+		{
+			name: "embeded error",
+			err:  &Error{Code: ENotFound, Err: &Error{Code: EInvalid}},
+			want: ENotFound,
+		},
+		{
+			name: "default error",
+			err:  errors.New("s"),
+			want: EInternal,
+		},
+	}
+	for _, c := range cases {
+		if result := ErrorCode(c.err); c.want != result {
+			t.Fatalf("%s failed, want %s, got %s", c.name, c.want, result)
+		}
+	}
+}
+
+func TestJSON(t *testing.T) {
+	cases := []struct {
+		name string
+		err  *Error
+		json string
+	}{
+		{
+			name: "simple error",
+			err:  &Error{Code: ENotFound},
+		},
+		{
+			name: "with op",
+			err: &Error{
+				Code: ENotFound,
+				Op:   "bolt.FindAuthorizationByID",
+			},
+		},
+		{
+			name: "with op and value",
+			err: &Error{
+				Code: ENotFound,
+				Op:   "bolt.FindAuthorizationByID",
+				Msg:  fmt.Sprintf("with ID %d", 323),
+			},
+		},
+		{
+			name: "with a third party error",
+			err: &Error{
+				Code: EFailedToGetStorageHost,
+				Op:   "cmd/fluxd.injectDeps",
+				Err:  errors.New("empty value"),
+			},
+		},
+		{
+			name: "with a internal error",
+			err: &Error{
+				Code: EFailedToGetStorageHost,
+				Op:   "cmd/fluxd.injectDeps",
+				Err:  &Error{Code: EEmptyValue, Op: "cmd/fluxd.getStrList"},
+			},
+		},
+	}
+	for _, c := range cases {
+		result, err := json.Marshal(c.err)
+		// encode testing
+		if err != nil {
+			t.Fatalf("%s encode failed, want err: %v, should be nil", c.name, err)
+		}
+		// decode testing
+		got := new(Error)
+		err = json.Unmarshal([]byte(result), got)
+		if err != nil {
+			t.Fatalf("%s decode failed, want err: %v, should be nil", c.name, err)
+		}
+		decodeEqual(t, c.err, got, "decode: "+c.name)
+	}
+}
+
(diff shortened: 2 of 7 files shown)


errors/error.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+)
+
+// constant errors.
+const (
+	// not found
+	AuthorizationNotFound       = Error("authorization not found")
+	AuthorizationNotFoundConext = Error("authorization not found on context")
```
jsternberg: This is misspelled.

errors/error.go
```diff
@@ -0,0 +1,58 @@
...
+	return string(e)
+}
+
+// errWithValue is a generic func to wrap an value to error
+// example:
+// err := fmt.Errorf("user with name %s already exists", name)
+// could be
+// UserNameAlreadyExist = errWithValue("user with name %s already exists")
+// err := UserNameAlreadyExist(name)
+func errWithValue(f string) func(args ...interface{}) error {
```
jsternberg: I think this is a cool trick, but I also don't think this is something we should do. Consider this example: https://play.golang.org/p/jh8lQF0dl5T

In the example, I use the same trick, but then I call the error function incorrectly with too many arguments. Because of the indirect way of calling `fmt.Errorf`, `go vet` doesn't catch this and tells me there isn't a problem.

I know it uses up more space in the library, but these should be top level functions so the type checker can do extra work. It's not like they are real types anyway.

errors/typed.go
```diff
@@ -0,0 +1,104 @@
...
+func (t Type) String() string {
+	return typStr[t]
+}
+
+// HTTPCode returns the http status code.
+func (t Type) HTTPCode() int {
+	return typCode[t]
+}
+
+// TypedError wraps error with a immutable type
```
jsternberg: > an immutable type.

errors/typed.go
```diff
@@ -0,0 +1,104 @@
...
+		return typedErr{
+			typ: typ,
+			err: err,
+			loc: make([]string, 0),
+		}
+	}
+}
+
+// funcs to create a new typed error
+// example tyedErr := InternalErrorf(err)
```
jsternberg:
```go
typedErr := InternalErrorf(err)
```

errors/typed.go
```diff
@@ -0,0 +1,104 @@
...
+var (
+	InternalErrorf = newTypedError(InternalError)
```
jsternberg: Same thing as above. While this doesn't have the same issue because it doesn't use the `Errorf` function, I'm still not a fan of the convention. I think it would be better to do something like this:

```
func newTypeError(err error) error {
    // does the common stuff that you are using the anonymous function for.
}

func InvalidDataf(err error) error { return newTypeError(InvalidData) }
```

Something like that maybe.

errors/typed.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (e typedErr) WrapLocation(loc string) {
+	e.loc = append(e.loc, loc)
+}
+
+// newTypedError is the generic func to wrap an error into TypedError
+func newTypedError(typ Type) func(err error) TypedError {
+	return func(err error) TypedError {
+		if err == nil {
```
jsternberg: Is there going to be some time when the error passed in is expected to be nil?

errors/http.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// HandleHTTP sets the X-Influx-Error and X-Influx-Reference headers on the response,
+func HandleHTTP(ctx context.Context, e TypedError, w http.ResponseWriter) {
+	if e == nil {
```
jsternberg: This seems wrong. This implies that it is ok for the error to be nil, but then it doesn't do anything with that knowledge so I don't see how this logic is useful for the caller or the callee.

errors/typed.go
```diff
@@ -0,0 +1,104 @@
...
+	// NotFound indicates a correct operation returns no results.
+	NotFound
+)
+
+var typCode = []int{
+	http.StatusInternalServerError,
+	http.StatusBadRequest,
+	http.StatusUnprocessableEntity,
+	http.StatusForbidden,
+	http.StatusUnprocessableEntity,
```
nathanielc: Should this be `http.StatusNotFound`?

kelwang: Yeah.

errors/typed.go
kelwang: Yeah, in fact, when I tried to apply the library. A lot of times, we can just wrap it with the typed error without check if err!=nil. 
For example: 
```go
    err := errors.StatusInternalServerError(c.db.View(func(tx *bolt.Tx) error {
          result = ....
    })
    return result, err
```

jsternberg: I'm not a fan of this pattern and would prefer we didn't use it. I find it confusing and I think it would be easier to use the standard go error handling pattern.

kelwang: Sure, let me double think about this. I think what you said makes sense. One concern would be if we can force certain pattern, we might also want to develop a tool to check that pattern.

kelwang:
Sure, great idea

errors/error.go
kelwang: :+1:

errors/const.go
```diff
@@ -0,0 +1,51 @@
...
+	EmptyValue
+)
+
+// common phases
+const (
+	notFound        = " not found"
+	notFoundContext = " not found on context"
+)
+
+var constStr = []string{
```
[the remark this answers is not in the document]
kelwang: Yeah, @goller mentioned same thing as well. Um, I feel like we might ended up with hundreds of errors eventually. Go's map is actually a hash map. implemented in https://golang.org/src/runtime/hashmap.go . I agree using map will increase readability, but compare to an array, there is also a lot of runtime memory overhead when this package is growing.

errors/errors_test.go
```diff
@@ -0,0 +1,183 @@
+package errors
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
kelwang: haha, no actually, we can talk about that. I just saw a lot of our vendors using this already.

errors/http.go
```diff
@@ -0,0 +1,98 @@
...
+		return e.Err.Error()
+	}
+	return fmt.Sprintf("%s: %v", e.Typ.Reference(), e.Err)
+}
+
+func (e httpError) HTTPCode() int {
+	return httpCode[e.HTTPTyp-baseHTTP]
+}
+
+func newHTTP(typ Type) func(e error) HTTPError {
```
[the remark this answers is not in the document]
kelwang: Sure, any suggestions?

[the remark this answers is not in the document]
kelwang: NewHTTPErrorGenerator

errors/const.go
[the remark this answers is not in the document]
kelwang: Correction, I just write some bench mark tests, seems the memory alloc are same compare array and map. Maybe because it's enum map. I will just do the map version, since it's easier to update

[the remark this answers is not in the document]
kelwang: 👍

errors/const.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Code implement the TypedError interface.
+func (e ConstError) Code() Type {
+	return Type(e)
+}
+
+// const errors
+const (
+	// AuthorizationNotFound indicate an error when authorization is not found.
+	AuthorizationNotFound ConstError = baseConst + iota
```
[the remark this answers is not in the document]
goller: I agree with @desa I think we should just start with a number; it's easier to lookup the number in the source code once an error is sent.

errors/const.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// common phases
+const (
+	notFound        = " not found"
+	notFoundContext = " not found on context"
+)
+
+var constStrMap = map[ConstError]string{
+	AuthorizationNotFound:        "authorization" + notFound,
```
goller: No need to break out common phrases like this.  

Imagine that you are looking through our source code for a specific error... you want the sentence easily searchable.

errors/const.go
```diff
@@ -0,0 +1,58 @@
...
+	OrganizationNotFound:         "organization" + notFound,
+	UserNotFound:                 "user" + notFound,
+	TokenNotFoundContext:         "token" + notFoundContext,
+	URLMissingID:                 "url missing id",
+	EmptyValue:                   "empty value",
+}
+
+// MarshalJSON implements json.Marshaler interface.
+func (e ConstError) MarshalJSON() ([]byte, error) {
+	return []byte("{}"), nil
```
goller: Is this correct?

errors/const.go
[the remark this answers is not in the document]
kelwang: 👍

errors/http.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// HTTPError has an additional HTTPCode method compare to TypedError.
+type HTTPError interface {
+	TypedError
+	HTTPCode() int
+}
+
+// HandleHTTP sets the X-Influx-Error and X-Influx-Reference headers on the response,
+func HandleHTTP(ctx context.Context, e HTTPError, w http.ResponseWriter) {
```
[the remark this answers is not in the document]
kelwang: 👍

errors/const.go
kelwang: 👍

errors/const.go
kelwang: 👍  forgot to remove that part

jsternberg:
Also, the current method makes it very hard to identify problems in the code using static typing. If you used a concrete type for each one, you gain a few advantages:

1. It is possible to do a type switch to identify the specific type of error and access attributes inside of it.
2. You can ensure it follows the interface by assigning it to a blank variable like this:

```
var _ TypedError = &MyError{}
```

The current method has no way to have the compiler error if you forget to update the maps that convert a type to a string so forgetting to do that can just end up in the code as a hard-to-find bug later on. This is especially true for an error that doesn't happen very frequently, but needs to exist.

Gopkg.toml
```diff
@@ -92,3 +92,7 @@ required = [
 [[constraint]]
   name = "gopkg.in/robfig/cron.v2"
   branch = "v2"
+
+[[constraint]]
```
goller: Make sure to update your `dep` and rerun dep ensure.

errors.go
```diff
@@ -0,0 +1,170 @@
...
+			ee.Err = string(b)
+		} else {
+			ee.Err = e.Err.Error()
+		}
+	}
+	return json.Marshal(ee)
+}
+
+// DecodeErrorJSON will decode the
+func DecodeErrorJSON(b []byte) (result *Error, err error) {
```
mark-rushakoff: Why not implement json.Marshaler and json.Unmarshaler on Error?

errors_test.go
```diff
@@ -0,0 +1,182 @@
...
+func TestErrorMsg(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		msg  string
+	}{
+		{
+			name: "simple error",
+			err:  &Error{Code: EAuthorizationNotFound},
+			msg:  "<authorization not found> ",
```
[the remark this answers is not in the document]
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,170 @@
...
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+type errCode string
+
+// errCode consts, all the err code must be defined here
```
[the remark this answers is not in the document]
goller: I agree with @mark-rushakoff I think this should be public and include the most broad error types like "not found"

I suspect each package could add their own codes in their own `errors.go` perhaps, but, as this was inspired by @benbjohnson 's post, what do you think @benbjohnson ?

errors.go
kelwang: 👍

errors.go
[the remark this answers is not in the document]
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,170 @@
...
+	} else if e, ok := err.(*Error); ok && e.Code != "" {
+		return string(e.Code)
+	} else if eHTTP, ok := err.(*HTTPError); ok && eHTTP.Err.Code != "" {
+		return string(eHTTP.Err.Code)
+	}
+	return string(EInternal)
+}
+
+// HandleHTTPErr sets the X-Influx-Error and X-Influx-Reference headers on the response,
+func HandleHTTPErr(ctx context.Context, e error, w http.ResponseWriter) {
```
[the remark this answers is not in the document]
kelwang: 👍 removed from this file. I will make another PR when we working on converting errors in http package, which could be a much bigger scope. Currently http package is using EncodeError function, which does exactly same thing

errors_test.go
```diff
@@ -0,0 +1,146 @@
...
+func TestErrorCode(t *testing.T) {
+	assert.Equal(t, "", ErrorCode(nil))
+	assert.Equal(t, string(EAuthorizationNotFound), ErrorCode(&Error{Code: EAuthorizationNotFound}))
+	assert.Equal(t, string(EAuthorizationNotFound), ErrorCode(&HTTPError{Err: Error{Code: EAuthorizationNotFound}}))
+	assert.Equal(t, string(EAuthorizationNotFound), ErrorCode(&Error{Code: EAuthorizationNotFound}))
+	assert.Equal(t, string(EAuthorizationNotFound), ErrorCode(&Error{Code: EAuthorizationNotFound, Err: &Error{Code: EURLMissingID}}))
+	assert.Equal(t, string(EInternal), ErrorCode(errors.New("s")))
+}
+
+func TestJSON(t *testing.T) {
```
e-dard: In general there is little upside to testing JSON explicitly. It's likely to lead to flaky tests down the road...

A JSON object is not strictly ordered. Aside from a possible Go runtime change to re-order things, with the existing Go implementation if you move the fields in your struct around, e.g., for struct packing or refactoring reasons, all your test cases will change.

A better way to test your marshaling/unmarshaling code is as follows:

 1) Setup base object;
 2) marshal base to JSON;
 3) provide result object and unmarshal JSON into result object;
 4) `reflect.DeepEqual` the base and result object.

errors_test.go
```diff
@@ -0,0 +1,146 @@
...
+				Op:   "cmd/fluxd.injectDeps",
+				Err:  &Error{Code: EEmptyValue, Op: "cmd/fluxd.getStrList"},
+			},
+			json: "{\"code\":\"failed to get the storage host\",\"op\":\"cmd/fluxd.injectDeps\",\"err\":\"{\\\"code\\\":\\\"empty value\\\",\\\"op\\\":\\\"cmd/fluxd.getStrList\\\"}\"}",
+		},
+	}
+	for _, c := range cases {
+		result, err := json.Marshal(c.err)
+		// encode testing
+		assert.Nil(t, err, "enode: "+c.name)
```
e-dard: I'm not seeing any value at all here for pulling in testify to be honest. These are all simple checks, and on-liners if you just write a small function./

Gopkg.toml
```diff
@@ -92,3 +92,7 @@ required = [
 [[constraint]]
   name = "gopkg.in/robfig/cron.v2"
   branch = "v2"
+
+[[constraint]]  
+  name = "github.com/stretchr/testify"
+  version = "=1.2.1"
```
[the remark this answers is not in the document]
e-dard: I'm 👎 on pulling in testing libraries.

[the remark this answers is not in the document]
jsternberg: 👎 to the testing library from me too.

bolt/source.go
```diff
@@ -96,16 +96,17 @@ func (c *Client) FindSourceByID(ctx context.Context, id platform.ID) (*platform.
 	return s, nil
 }
 
-func (c *Client) findSourceByID(ctx context.Context, tx *bolt.Tx, id platform.ID) (*platform.Source, error) {
+func (c *Client) findSourceByID(ctx context.Context, tx *bolt.Tx, id platform.ID) (*platform.Source, *platform.Error) {
```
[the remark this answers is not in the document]
kelwang: 👍

bolt/source.go
```diff
@@ -37,11 +37,11 @@ func (c *Client) initializeSources(ctx context.Context, tx *bolt.Tx) error {
 	}
 
 	_, err := c.findSourceByID(ctx, tx, DefaultSource.ID)
-	if err != nil && err != platform.ErrSourceNotFound {
+	if err != nil && err.Code != platform.ESourceNotFound {
```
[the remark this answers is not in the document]
kelwang: 👍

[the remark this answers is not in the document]
kelwang: 👍

bolt/source.go
```diff
@@ -96,16 +96,17 @@ func (c *Client) FindSourceByID(ctx context.Context, id platform.ID) (*platform.
 	return s, nil
 }
 
-func (c *Client) findSourceByID(ctx context.Context, tx *bolt.Tx, id platform.ID) (*platform.Source, error) {
+func (c *Client) findSourceByID(ctx context.Context, tx *bolt.Tx, id platform.ID) (*platform.Source, *platform.Error) {
+	op := platform.ErrOpCode("bolt.findSourceByID")
```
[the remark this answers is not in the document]
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,158 @@
...
+			if e.Msg != "" {
+				buf.WriteString(" ")
+			}
+		}
+		buf.WriteString(e.Msg)
+	}
+	return buf.String()
+}
+
+type errEncode struct {
```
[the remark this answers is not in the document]
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,158 @@
...
+	Msg string `json:"msg,omitempty"`
+	// Logical operation and nested error.
+	Op ErrOpCode `json:"op,omitempty"`
+	// the embed error
+	Err error `json:"err,omitempty"`
+}
+
+// Error implement the error interface
+func (e *Error) Error() string {
+	var buf bytes.Buffer
```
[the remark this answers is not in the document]
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,157 @@
...
+	Msg string `json:"msg,omitempty"`
+	// Logical operation and nested error.
+	Op ErrOpCode `json:"op,omitempty"`
+	// the embed error
+	Err error `json:"err,omitempty"`
+}
+
+// Error implement the error interface
+func (e Error) Error() string {
+	var buf bytes.Buffer
```
mark-rushakoff: [strings.Builder](https://golang.org/pkg/strings/#Builder) is probably more appropriate and more memory-efficient for this use case.

errors.go
```diff
@@ -0,0 +1,157 @@
...
+
+// ErrOpCode is the string of logical operation.
+// for example: bolt.UserCreate.
+// OpCode should be defined inside each package.
+type ErrOpCode string
+
+// Error is the error struct of idpe.
+// To create a simple error, &Error{Code:EUserNotFound}
+// To show where the error happens, add Op.
+// &Error{
```
mark-rushakoff: Code examples in comments should be indented so that they render properly in godoc: https://blog.golang.org/godoc-documenting-go-code

errors.go
kelwang: 👍

Gopkg.toml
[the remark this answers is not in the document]
benbjohnson: 👎

bolt/source.go
```diff
@@ -17,6 +17,11 @@ var (
 	sourceBucket = []byte("sourcesv1")
 )
 
+// ops
+const (
+	opFindSourceByID = platform.ErrOpCode("bolt.findSourceByID")
+)
+
```
benbjohnson: I find it easier just to have the op constant in the function itself. What's the benefit to making it global?

e.g.

```go
func findSourceByID() {
	const op = platform.ErrOpCode("bolt.findSourceByID")
	...
}
```

errors.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// ErrCode is a constant string to identify the type of of an error
+type ErrCode string
```
benbjohnson: I don't personally find it useful to make specific types for the code & op. Upspin uses them but that's only because they have a wonky builder function that requires specific types.

kelwang: 👍

bolt/source.go
kelwang: 👍

errors.go
kelwang: 👍

bolt/source.go
kelwang: 👍

errors.go
kelwang: 👍

errors.go
```diff
@@ -0,0 +1,158 @@
...
+	EFailedToGetBucketName        ErrCode = "failed to get the bucket name"
+	EOrganizationNameAlreadyExist ErrCode = "organization already exists"
+)
+
+// ErrOpCode is the string of logical operation.
+// for example: bolt.UserCreate.
+// OpCode should be defined inside each package.
+type ErrOpCode string
+
+// Error is the error struct of idpe.
```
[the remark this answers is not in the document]
kelwang: I think most time we just do a return &Error{}, and the main func will use error interface to return

errors_test.go
kelwang: 👍

errors_test.go
kelwang: Seems go lib has an example like that https://golang.org/src/encoding/json/encode_test.go#L213

And I never really experience any issues before

Gopkg.toml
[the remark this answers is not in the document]
kelwang: 👍

PR #64, opened 2018-05-29
fix(http): remove the orgID parameter from the transpiler handler

jsternberg:
The /query endpoint didn't have an orgID parameter so we can't have one
in the transpiler endpoint. This removes the parameter and replaces it
with a command line parameter that sets which organization we are using
for querying ifql.

Related to #65.

### DIFF
--- a/cmd/transpilerd/main.go
+++ b/cmd/transpilerd/main.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 
 	influxlogger "github.com/influxdata/influxdb/logger"
+	"github.com/influxdata/platform"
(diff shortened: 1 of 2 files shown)


pauldix:
Having this as a command line argument won't work. This needs to be pulled from the request. If something receives the request and translates it first, this can be a query parameter. If it's just a simple proxy then it will need to pull it from somewhere else like the requesting subdomain. I added a PR to the requirements doc about compatibility. See here for my writeup on translating orgs: https://github.com/influxdata/idpe/pull/674/files#diff-4f35dc3e36c0f06d0b482232fdd5451fR274

pauldix:
Ah, this might make more sense for the OSS version, that makes sense. Wasn't thinking about OSS vs Enterprise.

Issue #65, opened 2018-05-29
Make transpiler query handler accept the same parameters as the influxdb 1.x httpd service


jsternberg:
The transpiler query handler needs to expose the same interface as the influxdb 1.x httpd service.

pauldix:
Please make sure to read through: https://github.com/influxdata/idpe/pull/674 to make sure those requirements are captured.

jsternberg:
Split this into #122, #123, and #124.

There's also the chunking parameters, but we're going to punt on those for now since they aren't essential to making the `influx` CLI work from 1.x.

PR #68, opened 2018-05-29
fix(http): fix the transpiler handler to use an interface to retrieve an id

jsternberg:
The organization is retrieved in a customized way from the http request.
The open source version uses a static value from the command line while
enterprise will likely use something else.

This also removes the `--org` parameter because we can't properly decode
the name into an organization id without access to the idp daemon. We
will either have to open access to that or not have that feature.

Related to #65 and fixes #64.


http/transpiler_handler.go
```diff
@@ -12,19 +13,22 @@ import (
...
 
+type TranspilerRetrieveOrgFunc func(req *http.Request) (platform.ID, error)
+
 type TranspilerQueryHandler struct {
 	*httprouter.Router
 
 	Logger       *zap.Logger
 	QueryService query.QueryService
-	OrgID        platform.ID
+	OrgFunc      TranspilerRetrieveOrgFunc
```
nathanielc: If the transpiler needs to lookup Organization IDs from names then it should consume an instance of the `OrganizationService`. The implementation of that service can vary based on the context of the running transpilerd process.

jsternberg:
I figured that the current layout we have for OSS would work well. Give it a command line argument or we could build in another method in the future. I figured how OSS would retrieve the organization id was still a bit of an open question and I was fine with just using a command line argument to enable testing until we had nailed it down a bit more.

http/transpiler_handler.go
jsternberg: I figure that the main function, which will be the one to provide this function, can use a closure to keep an organization service if it needs it. I'm also sure that the inputs will likely change at some point, but the only inputs we know we need are the HTTP request to read headers at the moment.

Since an `OrganizationService` would never change getting passed in to the function here, I figure that we don't need to pass it in as an argument.

nathanielc:
OK, so for now the OSS version will use a static value. That is fine. We can update code when that changes. 

I think it is best of we use the `OrganziationService` as the interface instead of function closure. You can provide a static implementation of the OrganizationService in the OSS version. See `ifqld` in this repo that does just that.

jsternberg:
I'm not really sure how to use the `OrganizationService` as we don't need to do anything with the id other than pass it along. It would be nice if it was available to the command line so a name could be used and translated, but the `OrganizationService` does something fundamentally different than what this function is attempting to do.

Maybe it would be good to make this into an interface instead so that it's more explicit what the function is for?

http/transpiler_handler.go
```diff
@@ -12,19 +13,34 @@ import (
 	"go.uber.org/zap"
 )
 
+// TranspilerOrganizationProxy will inspect an http.Request to determine
+// which organization a transpiler request should be proxied to.
+type TranspilerOrganizationProxy interface {
+	OrganizationFromRequest(req *http.Request) (platform.ID, error)
```
nathanielc: This interface seems like an odd coupling of concepts to me. Specifically coupling http.Request with looking up an org ID. The http handler function should be responsible for reading the org name or ID off the request. And then consume the `OrganizationService` in order lookup an org ID from the name if necessary.

jsternberg: This PR is to customize how the organization is retrieved from the HTTP handler. Since the open source and enterprise will have different implementations of how it reads the http request, there needs to be an interface to customize this behavior.

The section of the code that is relevant is here: https://github.com/influxdata/platform/blob/master/http/query_service.go#L48

This implementation will be different in open source vs enterprise. Enterprise uses the request to determine which organization id it will use while open source uses a static value.

I don't need to lookup an organization id for this PR so the `OrganizationService` isn't the proper interface for what I need. It might be useful to have an `OrganizationService` implementation in the main function for translating the `--org` parameter, but I removed that so we didn't have to consider whether to add a reference to that implementation for this PR.

nathanielc: OK, I misunderstood the purpose of this PR.

>Since the open source and enterprise will have different implementations of how it reads the http request, there needs to be an interface to customize this behavior.

If the enterprise and OSS versions differ in how they consume an HTTP request then they are different http handlers. As such the enterprise server will not share the http handler code with this version. If there is non http specific code that you do want shared, then that code should be broken out into its own interface/function. 

We do not want to delegate the http specific behavior of the http handler to an outside interface.

PR #71, opened 2018-05-29
update test case for window functions

nathanielc:
Once #70 is merged this test will pass.


query/querytest/execute.go
```diff
@@ -58,20 +58,12 @@ func GetQueryEncodedResults(qs query.QueryService, spec *query.Spec, inputFile s
 	if err != nil {
 		return "", err
 	}
-	enc := csv.NewResultEncoder(csv.DefaultEncoderConfig())
+	enc := csv.NewMultiResultEncoder(csv.DefaultEncoderConfig())
```
aanthony1243: thanks, not sure how I missed this type.

PR #84, opened 2018-05-30
fix(cmd/transpilerd): fix transpilerd to use only --org-id

jsternberg:
This removes the `--org` parameter because we can't properly decode the
name into an organization id without access to the idp daemon. We will
either have to open access to that or not have that feature.

### DIFF
--- a/cmd/transpilerd/main.go
+++ b/cmd/transpilerd/main.go
@@ -49,11 +49,8 @@ func init() {
 	viper.BindEnv("IFQLD_HOSTS")
 	viper.BindPFlag("IFQLD_HOSTS", transpileCmd.PersistentFlags().Lookup("ifqld-hosts"))
 
-	transpileCmd.PersistentFlags().StringP("org", "o", "", "name of the organization that owns the bucket")
-	viper.BindEnv("ORG")
-	viper.BindPFlag("ORG", transpileCmd.PersistentFlags().Lookup("org"))
-
-	transpileCmd.PersistentFlags().StringP("org-id", "", "", "id of the organization that owns the bucket")
+	// TODO(jsternberg): Determine how we are going to identify the organization id in open source.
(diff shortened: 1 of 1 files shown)


cmd/transpilerd/main.go
```diff
@@ -49,13 +49,10 @@ func init() {
...
 	viper.BindPFlag("IFQLD_HOSTS", transpileCmd.PersistentFlags().Lookup("ifqld-hosts"))
 
-	transpileCmd.PersistentFlags().StringP("org", "o", "", "name of the organization that owns the bucket")
-	viper.BindEnv("ORG")
-	viper.BindPFlag("ORG", transpileCmd.PersistentFlags().Lookup("org"))
-
 	transpileCmd.PersistentFlags().StringP("org-id", "", "", "id of the organization that owns the bucket")
 	viper.BindEnv("ORG_ID")
 	viper.BindPFlag("ORG_ID", transpileCmd.PersistentFlags().Lookup("org-id"))
+	transpileCmd.MarkPersistentFlagRequired("org-id")
```
nathanielc: Does it make sense to give this a default value instead of requiring it? The org-id has no meaning right now, so any use of transpilerd will just have to pass a fake ID anyways.

jsternberg: I don't think so. It's easy enough to get a fake organization id and I would rather not revisit this code when the id does something.

nathanielc: Is this code supposed to be long lived? I expected this code to be removed once we had a way of passing the org ID along with each request. Having the org-id tied to a single process won't scale.

jsternberg: I don't really know. This is only for open source anyway.

nathanielc: In that case lets give it a default, we did the same thing  for ifqld, which means we can fix both at the same time when we are ready. In the meantime the transpilerd will be able to run without having to educate everyone to make up a fake id.

jsternberg: Done.

PR #88, opened 2018-05-31
feat(query/influxql): encode influxql json results correctly

jsternberg:
This fixes the encoder so that it will encode the response correctly to
a JSON blob using the outputs of the transpiler. The transpiler has also
been modified to pass through the correct values so the map function is
correctly constructed and the aggregate function is also correctly
constructed.

This removes the group function temporarily because it does not seem to
be working.


jsternberg:
I'm able to transpile the following query with this PR using the available open source tools. InfluxDB as the storage layer, ifqld as the query engine, transpilerd as the transpiler, and telegraf as the thing that provides data.

```
$ curl -v 'http://localhost:8098/query' --data-urlencode 'q=SELECT mean(total) FROM telegraf..processes WHERE time >= now() - 5m AND time < now()'
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8098 (#0)
> POST /query HTTP/1.1
> Host: localhost:8098
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 131
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 131 out of 131 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 31 May 2018 15:14:16 GMT
< Transfer-Encoding: chunked
<
{"results":[{"statement_id":0,"series":[{"name":"processes","columns":["mean","time"],"values":[[461.58620689655174,"2018-05-31T15:09:16Z"]]}]}]}
* Connection #0 to host localhost left intact
```

The group by functionality doesn't seem to be working so I need to inspect that to see if it's me making a mistake or if that functionality is currently broken.

query/influxql/result.go
```diff
@@ -69,27 +69,30 @@ func (e *MultiResultEncoder) Encode(w io.Writer, results query.ResultIterator) e
...
 
 			if err := b.Do(func(cr query.ColReader) error {
+				// Preallocate the number of rows for the response to make this section
+				// of code easier to read. Find a time column which should exist
+				// in the output.
 				var values [][]interface{}
-				j := 0
-				for idx, c := range b.Cols() {
-					if cr.Key().HasCol(c.Label) {
+				for j, c := range b.Cols() {
```
nathanielc: the ColReader type has a `Len` method that indicates the number of rows. This can be simplified to:

```
values = make([][]interface{}, cr.Len())
for j := range values {
	values[j] = make([]interface{}, len(r.Columns))
}
```

query/querytest/test_cases/simple_max.influxql
```diff
@@ -1 +1 @@
-select max(f1) from test..m1
\ No newline at end of file
+select max(f1) from test..m1 where time >= now() - 5m and time < now()
```
nathanielc: How does `now` work for this test when the times are static within the input file?

jsternberg: I think `fromCSV` just ignores the time range at the moment. I mostly changed it to this because the ifql code does this same thing and I wanted them to be in-sync with each other.

nathanielc: OK, sounds good. We can work out relative details later.

PR #92, opened 2018-05-31
fix(query/influxql): add group spec back to the transpiler

jsternberg:
The latest version of influxdb has fixed whatever was wrong with the
group function so this works properly again.


jsternberg:
This was a change to influxdb which is compiled separately. I made a mistake by using an old version so this code shouldn't have been removed to begin with. It's just because I was working with an old version while the new one was in flux.

Pun not intended.

PR #100, opened 2018-06-04
docs(query/influxql): update the transpiler readme

jsternberg:
Update the transpiler docs to account for a change in the query
language. A slightly different method is now used with some
clarifications previously existing problems.

The readme is also indexed with a table of contents for easier reading.

### DIFF
--- a/query/influxql/README.md
+++ b/query/influxql/README.md
@@ -2,91 +2,120 @@
 
 The InfluxQL Transpiler exists to rewrite an InfluxQL query into its equivalent query in IFQL. The transpiler works off of a few simple rules that match with the equivalent method of constructing queries in InfluxDB.
 
-## Identify the variables
+**NOTE:** The transpiler code is not finished and may not necessarily reflect what is in this document. When they conflict, this document is considered to be the correct way to do it. If you wish to change how the transpiler works, modify this file first.
 
-The InfluxQL query engine works by filling in variables and evaluating the query for the values in each row. The first step of transforming a query is identifying the variables so we can figure out how to fill them correctly. A variable is any point in the query that has a **variable or a function call**. Math functions do not count as function calls and are handled in the eval phase.
+1. [Identify the cursors](#identify-cursors)
+2. [Identify the query type](#identify-query-type)
+3. [Group the cursors](#group-cursors)
+4. [Create the cursors for each group](#group-cursors)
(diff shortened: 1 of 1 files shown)


query/influxql/README.md
```diff
@@ -2,91 +2,120 @@
...
-As an example:
+### <a name="identify-variables"></a> Identify the variables
 
-    // TODO(jsternberg): Figure out the real command that makes this work
-    // since it doesn't exist yet.
-    result = left_join(tables: {val1: max, val2: usage_system}, by: "val1", fn: (tables) => {val1: val1, val2: val2})
+Each of the variables in the group are identified. This involves inspecting the condition to collect the common variables in the expression while also retrieving the variables for each expression within the group. For a function call, this retrieves the variable used as a function argument rather than the function itself.
 
-The join is skipped when there is only one cursor. Replace `left_join` with `outer_join` if this is an aggregate or raw query.
+If a wildcard is identified, then the schema must be consulted for all of the fields and dimensions. If there is a wildcard in the dimensions, then the dimensions are excluded. If there is a specific listing of dimensions in the grouping, then those specific tags are excluded.
```
aanthony1243: what is a dimension?

query/influxql/README.md
```diff
@@ -2,91 +2,120 @@
...
+After creating the base cursors, each of them is joined into a single stream using an `inner_join`.
+
+**TODO(jsternberg):** Raw queries need to evaluate `fill()` at this stage while selectors and aggregates should not.
+
+    > SELECT usage_user, usage_system FROM telegraf..cpu WHERE time >= now() - 5m
+    val1 = create_cursor(db: "telegraf", start: -5m, m: "cpu", f: "usage_user")
+    val1 = create_cursor(db: "telegraf", start: -5m, m: "cpu", f: "usage_system")
+    inner_join(tables: {val1: val1, val2: val2}, except: ["_field"], fn: (tables) => {val1: tables.val1, val2: tables.val2})
+
+If there is only one cursor, then we need to exclude the `_field` column from the table.
```
aanthony1243: why?

jsternberg: Mostly because the field tag will be different and it needs to join on everything except the field name (the full series key). And because later in the algorithm assumes that `_field` has been stripped, we need to strip it here too.

PR #102, opened 2018-06-04
fix(http): fix auth header

kelwang:
Moved from 
https://github.com/influxdata/idpe/pull/558

https://github.com/influxdata/idpe/issues/550

I will create another PR in idpe, once this merged

### DIFF
--- a/http/platform_handler.go
+++ b/http/platform_handler.go
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"context"
 	nethttp "net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


http/platform_handler.go
```diff
@@ -63,3 +68,11 @@ func (h *PlatformHandler) ServeHTTP(w nethttp.ResponseWriter, r *nethttp.Request
 
 	nethttp.NotFound(w, r)
 }
+
+func parseAuthHeaderNSetContext(ctx context.Context, r *nethttp.Request) (context.Context, error) {
```
desa: maybe rename to `extractAuthorization`?

kelwang: 👍

Issue #24, opened 2018-05-22
Add options to query language for Tasks


pauldix:
The tasks API will need a three separate concepts represented: how frequently to run the task, a cron style schedule (more flexible version of frequency), and the representation of what `now` means for the IFQL script as it is executing. There are three things I'd like to discuss and address in this issue:

1.  Where should this information be kept? In the IFQL script or as task API metadata?
1.  How should we set the execution context of `now` time in IFQL? In the script? Part of API?
1.  What should these concepts be named?

I'll lay out an example to illustrate the different issues. Say you have the following query that you want run periodically:

```javascript
from(bucket:"foo")
  |> range(start: -1h)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> min()
  |> to(bucket:"bar")
```

So you're calculating the min value every one hour and outputting those results to the `bar` bucket. If the tasks scheduler simply kicked off this query every hour, we'd need to set the `now` time within that context so that the relative range is correct. Really, we'd want `now` to return whatever `now()` was at the start of the query, rounded to the nearest hour.

Another thing we want to expose to the user is the idea of having something run periodically, but to have it wait for a little bit before executing. For example, with our query above if we wanted it to wait until five minutes past the hour to run, but to act as though it is being run on the hour exactly, to facilitate some window of time for late data arrival. This could be `every_offset` or something like that.

In our initial iteration of tasks we should do the simple case for handling late arriving data, which is to just wait. For historical backfills or to fix something when they go over that window of time, we should have an API that allows the user to request a recalculation of the task for the historical window.

For example, if I fill data from yesterday, I want to tell the API to run the above script for every hour of yesterday (even while it is still going to run for the next hour of today when it rolls over).

Here are three different options I can think of to represent this:

## Option 1 - in IFQL

This data could be options at the top of the IFQL script. Here's a swag at it:

```
period = 1h
option every = period
option every offset = 5m
// option cron = "..." // this and every options would be mutually exclusive

// this changes the behavior of now() in the rest of the script to return the 
// time. round(...) could also be a date time literal. This also changes any
// relative time arguments to functions in IFQL
option now = round(time:now(), 1h)

from(bucket:"foo")
  |> range(start: -1h)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> min()
  |> to(bucket:"bar")
```

## Option 2 - Tasks API & now as query request argument

[...]

PR #105, opened 2018-06-05
feat(http): add http service for tasks

jademcgough:


http/task_service.go
```diff
@@ -0,0 +1,456 @@
...
+
+type getLogsRequest struct {
+	filter platform.LogFilter
+}
+
+func decodeGetLogsRequest(ctx context.Context, r *http.Request) (*getLogsRequest, error) {
+	qp := r.URL.Query()
+	req := &getLogsRequest{}
+
+	if id := qp.Get("task"); id != "" {
```
goller: I think this should be `tid`

http/task_service.go
```diff
@@ -0,0 +1,456 @@
...
+		req.filter.Task = &platform.ID{}
+		if err := req.filter.Task.DecodeFromString(id); err != nil {
+			return nil, err
+		}
+	}
+
+	if id := qp.Get("run"); id != "" {
```
goller: I think this should be `rid`

http/task_service.go
```diff
@@ -0,0 +1,456 @@
...
+
+type getRunsRequest struct {
+	filter platform.RunFilter
+}
+
+func decodeGetRunsRequest(ctx context.Context, r *http.Request) (*getRunsRequest, error) {
+	qp := r.URL.Query()
+	req := &getRunsRequest{}
+
+	if id := qp.Get("task"); id != "" {
```
goller: should this be tid?

http/task_service.go
```diff
@@ -0,0 +1,456 @@
...
+		req.filter.Limit = i
+	}
+
+	if time := qp.Get("afterTime"); time != "" {
+		// TODO (jm): verify format is correct, once we've decided on a format
+		req.filter.AfterTime = time
+	}
+
+	if time := qp.Get("beforeTime"); time != "" {
+		// TODO (jm): verify format is correct, once we've decided on a format
```
goller: I think this should be `rfc3339`

task.go
```diff
@@ -0,0 +1,84 @@
...
+
+import "context"
+
+// Task is a task. 🎊
+type Task struct {
+	ID     ID     `json:"id,omitempty"`
+	Name   string `json:"name"`
+	Status string `json:"status"`
+	Owner  User   `json:"owner"`
+	IFQL   string `json:"ifql"`
```
goller: probably `query` now (and in the swagger)

task.go
```diff
@@ -0,0 +1,84 @@
...
+	// Removes a task by ID and purges all associated data and scheduled runs
+	DeleteTask(ctx context.Context, id ID) error
+
+	// Returns logs for a run.
+	FindLogs(ctx context.Context, filter LogFilter) ([]*Log, int, error)
+
+	// Returns a list of runs that match a filter and the total count of returned runs.
+	FindRuns(ctx context.Context, filter RunFilter) ([]*Run, int, error)
+
+	FindRunByID(ctx context.Context, id ID) (*Run, error)
```
goller: need a comment here

http/task_service.go
jademcgough: yeah, this is what I was thinking... `rfc3339` just feels weird as an http get param for some reason, but I'll probably just use it everywhere

task.go
jademcgough: I thought Paul or someone pushed back against using `query`. I can change it though.

http/task_service.go
jademcgough: good catch... this should be pulled from the path

http/task_service.go
goller: is the difference `/tasks/{tid}?run=1` vs `/tasks/{tid}/runs/{rid}` ? 

if so, I would rather scope in the path.

task.go
goller: I think it should be query and in the future we can support more dialects with a `type` parameter.

http/swagger.yml
```diff
@@ -688,16 +727,16 @@ components:
           ]
         owner:
           $ref: "#/components/schemas/User"
-        ifql:
-          description: The ifql to run for this task.
+        query:
```
mark-rushakoff: We're using the noun `script` internally to tasks. I'd prefer to see script here for consistency. Other considered words included query, code, and program.

task.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Returns a single run
+	FindRunByID(ctx context.Context, id ID) (*Run, error)
+
+	// Creates and returns a new run (which is a retry of another run)
+	RetryRun(ctx context.Context, id ID) (*Run, error)
+}
+
+// TaskUpdate represents updates to a task
+type TaskUpdate struct {
```
mark-rushakoff: I believe the only thing that can be updated on a task is the script. The name and other details should be specified in the options: https://github.com/influxdata/platform/issues/24

I don't think we have official language support for those options yet, so the workaround for now in tasks is a special comment at start of the script like `//! every=5s;name=request metrics rollup`.

http/swagger.yml
jademcgough: done

task.go
jademcgough: 👍

PR #112, opened 2018-06-08
feat(query/influxql): implement filter conditions for tags in the transpiler

### LINKED ISSUE #111: Implement condition processing for the transpiler
The transpiler needs to process conditions beyond the time range. The first implementation of this should not try to figure out if the variable is a field or a tag so it should not attempt to perform a join on any conditions that would require a join.

jsternberg:
Fixes #111.

### DIFF
--- a/query/influxql/group.go
+++ b/query/influxql/group.go
@@ -4,6 +4,8 @@ import (
 	"github.com/influxdata/influxql"
 	"github.com/influxdata/platform/query"
 	"github.com/influxdata/platform/query/functions"
+	"github.com/influxdata/platform/query/semantic"
+	"github.com/pkg/errors"
 )
 
 type groupInfo struct {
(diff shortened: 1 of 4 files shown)


query/influxql/group.go
```diff
@@ -82,21 +85,93 @@ func (gr *groupInfo) createCursor(t *transpilerState) (cursor, error) {
...
+	if len(tags) > 0 {
+		cur = &tagsCursor{cursor: cur, tags: tags}
+	}
 
-	// TODO(jsternberg): Handle conditions, function calls, multiple variable references, and basically
-	// everything that needs to be done to create a cursor for a single group.
+	// Evaluate the conditional and insert a filter if a condition exists.
+	if cond != nil {
+		// Generate a filter expression by evaluating the condition and wrapping it in a filter op.
+		expr, err := t.mapField(cond, cur)
```
aanthony1243: I don't think this needs to be changed urgently, but I wonder how sustainable it is to have one function `mapField` that returns a variety of weakly related types.  It makes the code a bit hard to trace/follow.

query/influxql/group.go
```diff
@@ -82,21 +85,93 @@ func (gr *groupInfo) createCursor(t *transpilerState) (cursor, error) {
 		cursors = append(cursors, cur)
 	}
 
+	// TODO(jsternberg): Establish which variables in the condition are tags and which are fields.
```
aanthony1243: looks like this is done?

jsternberg: No. This is the next step of implementing a schema to determine which ones are fields (require a cursor) or tags (no cursor required).

This step will be hard which is why I didn't do it here. It assumes everything is a tag. I did build in the code that would handle fields _if_ they were ever created.

PR #126, opened 2018-06-12
docs(http): add swagger documentation for cells/dashboards API

desa:
This PR introduces a two new resources, `Cells` and `Dashboards`, with their associated APIs.

TODO:
- [ ] Come up with a better name than `Cell` (current suggestions are `Viz` or `Vis`)

In the separating of cells from a `Dashboard`, we introduced the concept of a `DashboardCell` which is a `Cell` that has positional information relevant to a dashboard.

To view the swagger, I'd recommend using https://editor.swagger.io/

 It's my intention for this review to be a discussion about naming for these items.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -5,6 +5,284 @@ info:
 servers:
   - url: /v1
 paths:
+  /cells:
+    post:
+      tags:
+        - Cells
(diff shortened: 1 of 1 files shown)


desa:
closing PR as it is outdated

PR #127, opened 2018-06-12
feat(query/influxql): support for windowing

### LINKED ISSUE #98: windowing an aggregate fails in transpiler
query: `select mean(io_time) from testdb..diskio group by time(1s)`
error:  `failed to run query: timeValue column "_start" does not exist`

jsternberg:
The transpiler now supports basic windowing. The window offsets are not
supported yet at all.

For windowing, we use the window function to split the points, perform
the aggregate/selector operation, and then we put them back into the
same window so they are within the same table as they originally were
located in. This is now reflected in the spec and the code.

Fixes #98.


query/influxql/group.go
```diff
@@ -189,9 +211,70 @@ type groupCursor struct {
...
+	// Perform the grouping by the tags we found. There is always a group by because
+	// there is always something to group in influxql.
+	// TODO(jsternberg): A wildcard will skip this step.
 	id := t.op("group", &functions.GroupOpSpec{
-		By: []string{"_measurement"},
+		By: tags,
 	}, in.ID())
+
+	if windowEvery > 0 {
+		id = t.op("window", &functions.WindowOpSpec{
```
aanthony1243: as discussed offline, we can add some logic here to update the range()  part of the spec to be aligned on the hour so that this will run more similarly to influxql's windowing.

jsternberg:
As an update to this, I've been having difficulty getting the unit tests for this working because of the time boundaries. So I'm a bit stalled on the unit tests portion of this at the moment and I'm taking a minor break to clear my head before continuing.

query/querytest/test_cases/window.out.csv
```diff
@@ -1,17 +1,11 @@
...
-,,7,2018-05-22T19:53:56Z,2018-05-22T19:53:57Z,io_time,diskio,host.local,disk2,2018-05-22T19:53:56Z,648
-,,8,2018-05-22T19:54:06Z,2018-05-22T19:54:07Z,io_time,diskio,host.local,disk0,2018-05-22T19:54:06Z,15205499
-,,9,2018-05-22T19:54:06Z,2018-05-22T19:54:07Z,io_time,diskio,host.local,disk2,2018-05-22T19:54:06Z,648
-,,10,2018-05-22T19:54:16Z,2018-05-22T19:54:17Z,io_time,diskio,host.local,disk0,2018-05-22T19:54:16Z,15205755
-,,11,2018-05-22T19:54:16Z,2018-05-22T19:54:17Z,io_time,diskio,host.local,disk2,2018-05-22T19:54:16Z,648
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,string,dateTime:RFC3339,double
+#partition,false,false,true,true,true,false,false
+#default,0,,,,,,
+,result,table,_start,_stop,_measurement,_time,mean
+,,0,1970-01-01T00:00:00Z,2262-04-11T23:47:16.854775807Z,diskio,2018-05-22T19:53:26Z,7602668
```
nathanielc: Why does this work? I would expect start to be the date in 1655 or whatever corresponds to `-math.MaxInt64` not 1970. We want the 1970 date, but I'd like to understand where its coming from in the code.

jsternberg: I found it. It's in the same package.

In the future, we're going to need to hide the window moving function inside of an interface most likely and not use a random constant. It'll end up working a lot better in the code and will also work well when we add time zone support into the windowing. For now, this should work.

PR #133, opened 2018-06-14
Linking to Flux from README.md

mbbroberg:
Making it easier for interested parties to find Flux. Also linking to the best resource so far, which is @pauldix's presentation.

A couple other suggestions for discoverability (independent from the PR): 

- Once you have a blog/page on the website, link users there instead of directly to Paul's slides (always nice to land on the company page)
- you might want to setup another pointer repository at github.com/influxdata/flux for visibility. I couldn't find Flux without real sleuthing before realizing it was in here. Side note: you could end up migrating there as the project grows. In the meantime, you can turn off Issues on the repo and point people into here

Either way, I hope this helps. Congratulations on the release! 🎉


pauldix:
Awesome, thanks!

PR #148, opened 2018-06-19
feat(http): add swagger definitions for influxql query

goller:
This is documentation for the 1.0 compatibility query path.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -134,10 +134,10 @@ paths:
           description: bucket is the query and write destination store
         - in: query
           name: precision
-          description: preciscion rounds the timestamp to the nearest unit
+          description: precision rounds the timestamp to the nearest unit
           schema:
             type: string
-            description: acceptable epochs define the time units returned in the query
+            description: precision define the time units returned in the query
             enum: [
               "n",
               "u",
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -103,6 +103,138 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /query:
+   servers:
+      - url: /
+   get:
+    tags:
+      - Query
+    summary: Executes InfluxQL to retrieve time-series data
+    parameters:
+        - in: query
+          name: q
+          required: true
+          schema:
+            type: string
+          description: URL-encoded InfluxQL query to execute
+        - in: query
+          name: db
+          schema:
+            type: string
+          description: the database has been renamed to bucket.
+        - in: query
+          name: rp
+          schema:
+            type: string
+          description: optional retention policy to query (> 1.6.0)
+        - in: query
+          name: bucket
+          schema:
+            type: string
+          description: bucket is the query and write destination store
+        - in: query
+          name: precision
+          description: preciscion rounds the timestamp to the nearest unit
+          schema:
+            type: string
+            description: acceptable epochs define the time units returned in the query
+            enum: [
+              "n",
+              "u",
+              "ms",
+              "s",
+              "m",
+              "h"
+            ]
+        - in: query
+          name: epoch
+          description: override default RFC3339 time format with unix epoch at a precision; epoch rounds the timestamp to the nearest unit
+          schema:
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -5,6 +5,138 @@ info:
 servers:
   - url: /v1
 paths:
+  /query:
+   servers:
```
jademcgough: Is this supposed to be here? It's not an HTTP verb.

http/swagger.yml
```diff
@@ -5,6 +5,138 @@ info:
...
+            type: string
+          description: optional retention policy to query (> 1.6.0)
+        - in: query
+          name: bucket
+          schema:
+            type: string
+          description: bucket is the query and write destination store
+        - in: query
+          name: precision
+          description: preciscion rounds the timestamp to the nearest unit
```
jademcgough: precision

http/swagger.yml
```diff
@@ -5,6 +5,138 @@ info:
...
+          schema:
+            type: string
+            description: acceptable epochs define the time units returned in the query
```
jademcgough: looks like a copy-paste of the descriptions from the epoch schema

http/swagger.yml
```diff
@@ -5,6 +5,138 @@ info:
...
+              "m",
+              "h"
+            ]
+        - in: query
+          name: epoch
+          description: override default RFC3339 time format with unix epoch at a precision; epoch rounds the timestamp to the nearest unit
+          schema:
+            type: string
+            description: acceptable epochs define the time units returned in the query
+            enum: [
```
jademcgough: Why make this be an enum? I'd prefer boolean, and use `precision` to determine the precision.

jademcgough: Alternatively, we could have a `format` param if we expect to support more than unix epoch and rfc3339 (though this seems unnecessary for now).

http/swagger.yml
```diff
@@ -5,6 +5,138 @@ info:
...
+            enum:
+             - gzip
+        - in: header
+          name: X-Request-Id
+          description: specifies correlating requests across services
+          schema:
+            type: string
+            format: uuid
+    responses:
+      '200':
```
nhaugo: 204 is used in, at least, the ping endpoint

nhaugo: And writes

http/swagger.yml
```diff
@@ -774,3 +906,72 @@ components:
           readOnly: true
           type: string
       required: [code, message]
+    InfluxQLResults:
+      properties:
+        error:
```
jademcgough: fyi we have an `Error` component that could be used here (just mentioning, I have no preference other than eventual consistency in our API).

http/swagger.yml
goller: @jademcgough yup, it overrides /v1 to / so, in theory the path is `/query`

http/swagger.yml
goller: @jademcgough , yeah, as far as I can tell from the code, one could use either parameter for some rason.

http/swagger.yml

http/swagger.yml
goller: @jademcgough good point.  What do you think of me putting a `oneOf` in the error response schema on the endpoint for this error and our new style error?  

I put it here just for backwards compat.

http/swagger.yml
jademcgough: That's fine, I just mean that the description should be `acceptable precisions` instead of `acceptable epochs`.

jademcgough:
Feel free to ignore most of my comments, I hadn't realized that this is documentation of existing code.

http/swagger.yml
goller: Fixed in ae1b03e

goller:
@jademcgough I'll do a follow-on PR for the new-school query

http/swagger.yml
jademcgough: don't worry about this, then :3

PR #156, opened 2018-06-20
Feature/flux swagger

goller:
Some of this PR is aspiration and some of it is documentation.  There are differences between the SPEC.md and the actual code at the moment, but, I figure we can check the endpoint code after getting this PR in.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -1036,27 +1036,32 @@ components:
             header:
               description: if true, the results will contain a header row
               type: boolean
+              default: true
             delimiter:
               description: separator between cells; the default is ,
               type: string
               default: ","
+              maxLength: 1
             quoteChar:
-              description: set the quoting char for strings
+              description: set the quoting char for strings; cannot be the same as delimiter
               type: string
               default: \"
+              maxLength: 1
             annotations:
               description: https://www.w3.org/TR/2015/REC-tabular-data-model-20151217/#columns
               type: array
               items:
                 type: string
                 enum:
-                  - partition
+                  - group
                   - datatype
                   - default
+                uniqueItems: true
             commentPrefix:
               description: character prefixed to comment strings
               type: string
               default: \#
+              maxLength: 1
             dateTimeFormat:
               description: format of timestamps
               type: string
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -109,6 +109,7 @@ paths:
    get:
     tags:
       - Query
+      - InfluxQL
     summary: Executes InfluxQL to retrieve time-series data
     parameters:
         - in: query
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -235,6 +236,79 @@ paths:
...
+          type: string
+    requestBody:
+        description: flux query or specification to execute
+        content:
+          application/json:
+            schema:
+              oneOf:
+                -  $ref: "#/components/schemas/Query"
+                -  $ref: "#/components/schemas/QuerySpecification"
+    responses:
```
jademcgough: should probably have a default response as well

http/swagger.yml
```diff
@@ -860,6 +934,114 @@ paths:
...
+                  default: 0
+        dialect:
+          $ref: "#/components/schemas/Dialect"
+    Dialect:
+          description: dialect are options to change the default CSV output format; https://www.w3.org/TR/2015/REC-tabular-metadata-20151217/#dialect-descriptions
+          type: object
+          properties:
+            header:
+              description: if true, the results will contain a header row
+              type: boolean
```
nathanielc: `default: true`

http/swagger.yml
```diff
@@ -860,6 +934,114 @@ paths:
...
+            delimiter:
+              description: separator between cells; the default is ,
```
nathanielc: Must be a single character.

http/swagger.yml
```diff
@@ -860,6 +934,114 @@ paths:
...
+              type: string
+              default: ","
+            quoteChar:
+              description: set the quoting char for strings
```
nathanielc: Must be a single character and different from delimiter.

http/swagger.yml
```diff
@@ -860,6 +934,114 @@ paths:
...
+              type: string
+              default: \"
+            annotations:
+              description: https://www.w3.org/TR/2015/REC-tabular-data-model-20151217/#columns
+              type: array
+              items:
+                type: string
+                enum:
+                  - partition
```
nathanielc: Rename partition -> group. This is a name change we have planned.

http/swagger.yml
```diff
@@ -860,6 +934,114 @@ paths:
...
+              description: separator between cells; the default is ,
+              type: string
+              default: ","
+            quoteChar:
+              description: set the quoting char for strings
+              type: string
+              default: \"
+            annotations:
+              description: https://www.w3.org/TR/2015/REC-tabular-data-model-20151217/#columns
+              type: array
```
nathanielc: Currently we have a default set of annotations which is `["group", "datatype", "default"]`. Do we want that to continue to be the default? If not should we state that the default is the empty list?

http/swagger.yml
```diff
@@ -235,6 +236,79 @@ paths:
...
+              schema:
+                type: string
+                example: >
+                  result,table,_start,_stop,_time,region,host,_value
+                  mean,0,2018-05-08T20:50:00Z,2018-05-08T20:51:00Z,2018-05-08T20:50:00Z,east,A,15.43
+                  mean,0,2018-05-08T20:50:00Z,2018-05-08T20:51:00Z,2018-05-08T20:50:20Z,east,B,59.25
+                  mean,0,2018-05-08T20:50:00Z,2018-05-08T20:51:00Z,2018-05-08T20:50:40Z,east,C,52.62
+            application/vnd.influx.arrow:
+              schema:
+                type: string
```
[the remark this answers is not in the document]
jademcgough: Yes: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types

http/swagger.yml
```diff
@@ -235,6 +236,79 @@ paths:
...
+            - application/vnd.influx.arrow
+      - in: header
+        name: Content-Type
+        schema:
+          type: string
+          enum:
+            - application/json
+      - in: query
+        name: query
+        description: flux query string to execute; used if there is no POST body.
```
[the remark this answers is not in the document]
nathanielc: What do you mean by prefer? Meaning if both query parameters and POST body exist use the body? I think that case is an error, not a preference.

My question is how do we now which value to read. Currently the Content-Type controls that code path. If you set Content-Type to application/json we expect a POST body. If Content-Type is empty we expect a query parameter. So the question is: should we change the logic to be driven by the existence of the POST body instead of the Content-Type header?

http/swagger.yml
goller: I think the default should be empty array right now.  That way client code can opt into using those features.

http/swagger.yml
```diff
@@ -235,6 +236,79 @@ paths:
...
+      - in: header
+        name: Accept
+        description: specifies the return content format. Each response content type will have its own dialect options.
+        schema:
+          type: string
+          description: return format of either CSV or Arrow buffers
+          default: text/csv
+          enum:
+            - text/csv
+            - application/vnd.influx.arrow
```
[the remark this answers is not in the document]
goller: Yup, exactly. I have this in the vendor tree from: https://tools.ietf.org/html/rfc6838#section-3.2

http/swagger.yml
goller: Fixed in 8434ee7; good catch

http/swagger.yml
goller: Fixed in b2916c1

http/swagger.yml
goller: Fixed in b2916c1

http/swagger.yml
goller: Fixed in b2916c1

http/swagger.yml
goller: Fixed in b2916c1

http/swagger.yml
goller: So, I added the default empty array in 448a6cb

What do you think?

nathanielc: Yes, I agree an empty array as default makes sense.

http/swagger.yml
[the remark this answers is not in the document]
nathanielc: I don't have a strong feeling either way. My initial thoughts is that using Content-Type provides a nice ways to explicitly control the intent of the request. But I see the value in having a default Content-Type of application/json to make it easier to use from the CLI.

The only concern I have is around how easy it is to reliably check if the POST body exists. A quick read of the docs shows that it will be a combination of checking the ContentLength header and doing an initial read of the Body to see if it returns any data or an EOF.

PR #157, opened 2018-06-20
add syntax for query metadata to Flux grammar

### LINKED ISSUE #24: Add options to query language for Tasks

jlapacik:
This commit closes #24 . It adds the syntax for optional query metadata to the Flux grammar. The Flux grammar defines this syntax strictly as variable declarations preceded by the "option" keyword.

### DIFF
--- a/query/parser/parser_test.go
+++ b/query/parser/parser_test.go
@@ -19,6 +19,92 @@ func TestParse(t *testing.T) {
 		wantErr bool
 	}{
 		{
+			name: "optional query metadata",
+			raw: `option task = {
+				name: "foo",
+				every: 1h,
+				delay: 10m,
+				cron: "0 2 * * *",
+				retry: 5,
+			  }`,
+			want: &ast.Program{
+				Body: []ast.Statement{
+					&ast.OptionStatement{
+						Declaration: &ast.VariableDeclarator{
+							ID: &ast.Identifier{Name: "task"},
+							Init: &ast.ObjectExpression{
+								Properties: []*ast.Property{
+									{
+										Key:   &ast.Identifier{Name: "name"},
+										Value: &ast.StringLiteral{Value: "foo"},
+									},
+									{
+										Key:   &ast.Identifier{Name: "every"},
+										Value: &ast.DurationLiteral{Value: 1 * time.Hour},
+									},
+									{
+										Key:   &ast.Identifier{Name: "delay"},
+										Value: &ast.DurationLiteral{Value: 10 * time.Minute},
+									},
+									{
+										Key:   &ast.Identifier{Name: "cron"},
+										Value: &ast.StringLiteral{Value: "0 2 * * *"},
+									},
+									{
+										Key:   &ast.Identifier{Name: "retry"},
+										Value: &ast.IntegerLiteral{Value: 5},
+									},
+								},
+							},
+						},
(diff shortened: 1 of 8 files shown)


query/parser/parser_test.go
```diff
@@ -18,6 +18,92 @@ func TestParse(t *testing.T) {
...
+			raw: `option task = {
+				name: "foo",
+				every: 1h,
+				delay: 10m,
+				cron: "0 2 * * *",
+				retry: 5,
+			  }`,
+			want: &ast.Program{
+				Body: []ast.Statement{
+					&ast.VariableDeclaration{
```
nathanielc: Using the AST how would I know the difference between a VariableDeclaration and an OptionStatement?

PR #164, opened 2018-06-22
add option statement to semantic graph

jlapacik:
This commit adds the option statement as a recognized node in
the semantic graph of a Flux query.

Flux interpreter must visit an option statement node

interpreter parses option statement

### DIFF
--- a/query/interpreter/interpreter.go
+++ b/query/interpreter/interpreter.go
@@ -30,6 +30,10 @@ func (itrp interpreter) eval(program *semantic.Program, scope *Scope) error {
 func (itrp interpreter) doStatement(stmt semantic.Statement, scope *Scope) error {
 	scope.SetReturn(values.InvalidValue)
 	switch s := stmt.(type) {
+	case *semantic.OptionStatement:
+		if err := itrp.doStatement(s.Declaration, scope); err != nil {
+			return err
+		}
 	case *semantic.NativeVariableDeclaration:
 		if err := itrp.doVariableDeclaration(s, scope); err != nil {
 			return err
@@ -584,6 +588,12 @@ func (f function) resolveIdentifiers(n semantic.Node) (semantic.Node, error) {
 			}
 			n.Body[i] = node.(semantic.Statement)
 		}
+	case *semantic.OptionStatement:
+		node, err := f.resolveIdentifiers(n.Declaration)
+		if err != nil {
+			return nil, err
+		}
+		n.Declaration = node.(semantic.VariableDeclaration)
 	case *semantic.ExpressionStatement:
 		node, err := f.resolveIdentifiers(n.Expression)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


query/interpreter/interpreter.go
```diff
@@ -30,6 +30,10 @@ func (itrp interpreter) eval(program *semantic.Program, scope *Scope) error {
 func (itrp interpreter) doStatement(stmt semantic.Statement, scope *Scope) error {
 	scope.SetReturn(values.InvalidValue)
 	switch s := stmt.(type) {
+	case *semantic.OptionStatement:
+		if err := itrp.doStatement(s.Declaration.(*semantic.NativeVariableDeclaration), scope); err != nil {
```
nathanielc: Does the VariableDeclaration implement the Statement interface? If not would it make sense to do so? That way you could pass the s.Declaration directly. Not sure if this is a good approach, but we should explore it.

query/interpreter/interpreter_test.go
```diff
@@ -249,6 +249,22 @@ func TestEval(t *testing.T) {
...
+				name: "foo",
+				every: 1h,
+				cron: "0 2 * * *",
+			}
+
+			f = (r) => {
+                r2 = r * r
+                return (r - r2) / r2
+            }
+			f(r:2.0) == -0.5 or fail()
```
nathanielc: Can you update this test to use a property of the `task` option so we know that it was added to the scope correctly?

Something like `task.name == "foo" or fail()`?

query/interpreter/interpreter.go
jlapacik: The VariableDeclaration interface implements the Node interface, not the Statement interface. However the two concrete types that implement VariableDeclaration (NativeVariableDeclaration and ExternalVariableDeclaration) both implement Statement as well. As it stands right now, an option can only be a NativeVariableDeclaration. By having the VariableDeclaration interface implement Statement, I believe this would allow an option to be an ExternalVariableDeclaration as well.

jlapacik: @nathanielc Is there a reason a variable declaration should not be considered a statement?

nathanielc: Nope, if both concrete types are already statements lets update the VariableDeclaration to be a Statement too.

query/interpreter/interpreter_test.go
jlapacik: Test has been updated.

query/interpreter/interpreter.go
jlapacik: VariableDeclaration now implements the Statement interface.

query/interpreter/interpreter_test.go
```diff
@@ -249,6 +249,23 @@ func TestEval(t *testing.T) {
...
+				every: 1h,
+				cron: "0 2 * * *",
+				const: 2.5,
+			}
+
+			f = (r) => {
+                r2 = r * r * task.const
+                return (r - r2) / r2
+            }
+			f(r:2.0) == -0.8 or fail()
```
nathanielc: Can you simplify this test? I'd like it to test the task option directly and as little as possible of anything else. That way when the test breaks we can be confident we broken something related to options and not other Flux features like functions etc.

jlapacik: It's updated.

PR #173, opened 2018-06-26
Retrieve and set optional query paramaters from global scope

jlapacik:
This commit adds Flux options as a separate entity to the scope, as well as an interface for getting and setting options at the current scope. The interpreter uses this interface to add any external options to the scope before evaluating the program, and to return all options available to the current scope after program evaluation. 

There is still a question of whether getting and setting options should only be valid at the global scope.

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -61,7 +61,7 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
 		return nil, err
 	}
 
-	if err := interpreter.Eval(semProg, interpScope); err != nil {
+	if _, err := interpreter.Eval(semProg, interpScope, nil); err != nil {
 		return nil, err
 	}
 	spec := qd.ToSpec()
@@ -284,11 +284,13 @@ func builtIns(qd *queryDomain) (map[string]values.Value, semantic.DeclarationSco
 			panic(errors.Wrapf(err, "failed to create semantic graph for builtin %q", name))
 		}
 
-		if err := interpreter.Eval(semProg, interpScope); err != nil {
+		if err := interpreter.EvalBuiltIn(semProg, interpScope); err != nil {
 			panic(errors.Wrapf(err, "failed to evaluate builtin %q", name))
 		}
 	}
-	return scope, decls
+	// TODO(Josh): I think we only need the values in
+	// the top-level scope, but not totally sure about this.
+	return interpScope.GetValues(), decls
 }
 
 type Administration struct {
--- a/query/interpreter/interpreter.go
+++ b/query/interpreter/interpreter.go
@@ -10,13 +10,44 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Eval(program *semantic.Program, scope *Scope) error {
-	itrp := interpreter{}
-	return itrp.eval(program, scope)
+// Eval evaluates a Flux program with respect to a global scope.
+// All variables are resolved to a scope below that of the input scope.
+// Option statements are allowed in a Flux program and option parameters
+// are resolved to a scope directly below that of the global(builtin) scope.
+func Eval(program *semantic.Program, scope *Scope, opts map[string]values.Value) (*Scope, error) {
+	// Ensure input scope is global scope
+	if scope == nil || scope.parent != nil {
+		return nil, errors.New("Expecting builtin scope")
+	}
+
+	// Create the options scope
+	optionsScope := scope.Nest()
+
+	// Add external options to options scope
+	optionsScope.SetValues(opts)
+
+	// Create the program scope
+	programScope := optionsScope.Nest()
+
+	// Evaluate program and resolve variables
+	err := interpreter{}.eval(program, programScope)
+
+	// Return the options scope after evaluation
+	return optionsScope, err
 }
 
-type interpreter struct {
+// EvalBuiltIn evaluates built-in programs exclusively.
+// All variables are resolved to the global(builtin) scope.
+// There are assumed to be no option statements defined in a builtin program.
+func EvalBuiltIn(program *semantic.Program, scope *Scope) error {
+	// Ensure input scope is global scope
+	if scope == nil || scope.parent != nil {
+		return errors.New("Expecting builtin scope")
+	}
+	return interpreter{}.eval(program, scope)
 }
+
+type interpreter struct{}
 
 func (itrp interpreter) eval(program *semantic.Program, scope *Scope) error {
 	for _, stmt := range program.Body {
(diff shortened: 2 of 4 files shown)


query/compile.go
```diff
@@ -284,7 +284,7 @@ func builtIns(qd *queryDomain) (map[string]values.Value, semantic.DeclarationSco
 			panic(errors.Wrapf(err, "failed to create semantic graph for builtin %q", name))
 		}
 
-		if err := interpreter.Eval(semProg, interpScope); err != nil {
+		if _, err := interpreter.Eval(semProg, interpScope, map[string]values.Value{}); err != nil {
```
nathanielc: In general is it best to design a function to be able to take a nil value for a map if its not needed. It is safe to read from nil maps in Go, they simply report that the key does not exist.

query/interpreter/interpreter.go
```diff
@@ -329,6 +348,7 @@ func (itrp interpreter) doArguments(args *semantic.ObjectExpression, scope *Scop
 type Scope struct {
 	parent      *Scope
 	values      map[string]values.Value
+	options     map[string]struct{}
```
nathanielc: Since Scopes can be nested, what if we created a nested Scope as the special OptionsScope instead of adding an extra options map to the Scope struct. Thoughts?

Independent of that change I think `map[string]bool` would be a better choice instead of empty struct. A read from a map returns the zero value if the key does not exist. So to use a map as set for testing existence do this:

```
if  options[key] {
 // option exists
} else {
 // option does not exist
}
```

query/interpreter/interpreter.go
```diff
@@ -358,6 +379,50 @@ func (s *Scope) Set(name string, value values.Value) {
...
+	_, exists := s.options[name]
+	if !(ok && exists) {
+		return s.parent.GetOption(name)
+	}
+	return val, exists
+}
+
+// GetAllOptions retrieves all option varialbes starting at
+// the current scope and traveling up to the global scope.
+func (s *Scope) GetAllOptions(options map[string]values.Value) {
```
nathanielc: If we switch to using an explicit Scope as the Option Scope then these Option specific methods are not longer needed on a scope. GetAllOptions would become a function like `GetValues()` and it would return all values for this layer of the scope.

query/interpreter/interpreter.go
jlapacik: Yes I like the idea of having a scope specifically for options. This way there is one and only one scope that will ever define option variables.

query/compile.go
jlapacik: Totally agree.

query/interpreter/interpreter.go
jlapacik: Here's an interesting question. When we evaluate a program w.r.t a specific scope, call it "scope", do we nest the options' scope directly inside of "scope". This has the implicit assumption that "scope" is the global scope. Is this assumption valid? Would it be better to traverse to the global scope and then nest? Are there any situations where this latter method would be undesirable?

nathanielc: The scope passed into the `Eval` method is the "builtin" or global scope. Nesting the options scope below the builtin scope make sense to me.

query/compile.go
jlapacik: This has been updated.

query/interpreter/interpreter.go
jlapacik: This has been updated. Options are now placed in their own scope which is one level below the builtin scope.

query/interpreter/interpreter.go
jlapacik: This has been updated. Scope now has a `GetValues()` method and a `SetValues()` method.

query/interpreter/interpreter.go
```diff
@@ -72,6 +103,30 @@ func (itrp interpreter) doStatement(stmt semantic.Statement, scope *Scope) error
 	return nil
 }
 
+func (itrp interpreter) doOptionDeclaration(declaration semantic.VariableDeclaration, scope *Scope) error {
+	// Ensure scope is grandchild of global scope
```
nathanielc: What if instead of checking the heritage of the scope we simply kept a reference to the special options scope on the interpreter struct? Then we do not need a special doOptionDeclaration but in the `eval` function above you would pass in the `itrp.optionScope` to `doStatement` in the OptionStatment case. Thoughts?

query/interpreter/interpreter.go
```diff
@@ -10,14 +10,45 @@ import (
...
+
+	// Return the options scope after evaluation
+	return optionsScope, err
 }
 
-type interpreter struct {
+// EvalBuiltIn evaluates built-in programs exclusively.
+// All variables are resolved to the global(builtin) scope.
+// There are assumed to be no option statements defined in a builtin program.
+func EvalBuiltIn(program *semantic.Program, scope *Scope) error {
```
nathanielc: Was there code that depended on this behavior? If so what does it look like to update that code to use the new `Eval` function? I would rather have a single endpoint so a consumer of the code does not have to decide which version they need.

query/interpreter/interpreter.go
```diff
@@ -337,12 +392,30 @@ func NewScope() *Scope {
...
 func NewScopeWithValues(values map[string]values.Value) *Scope {
 	return &Scope{
 		values: values,
 	}
 }
 
+// GetValues returns all variable bindings for this scope.
+// Note that the parent scope is not included.
+func (s *Scope) GetValues() map[string]values.Value {
+	scope := make(map[string]values.Value)
```
nathanielc: A small tip, to help reduce the allocations that Go has to do when you are making a copy of something you can use the length of the source to tell Go how large to make the map (or slice) to begin with.

```
scope := make(map[string]values.Value, len(s.values))
```

This way only one allocation needs to be made since all entries are guaranteed to fit before we have to resize it.

query/interpreter/interpreter_test.go
```diff
@@ -284,6 +285,77 @@ func TestEval(t *testing.T) {
 	}
 
 }
+
+func TestGetOptions(t *testing.T) {
```
nathanielc: Can you update this test to not depend on the parsing step or the Eval steps? This means have the test case takes a scope and a map of values.  It allows the test to test only the code around options without having to worry about bugs in the parser or interperter etc causing this test to fail.

In order to test the Options are handled correctly through the Eval function, can you update the `TestEval` function. Add an options field to the test case struct and update the test function to pass in options to the Eval function. The existing test cases will not change as the default value will be a `nil` options. Then you can add new test cases specific to options.

query/interpreter/interpreter.go
jlapacik: Yes, `query.builtIns()` is responsible for constructing the builtin scope. It can't use the current `Eval()` function because `Eval()` assumes the builtin scope already exists and just constructs the `options` scope and the `program` scope. I don't really like that behavior, so I'm currently trying to come up with a way where we can use just the `Eval()` function for everything. Keeping a reference to the `options` scope in the interpreter might help.

query/interpreter/interpreter.go
jlapacik: I like that idea. I think it will make things a little simpler.

PR #175, opened 2018-06-26
update flux grammar spec to include OptionStatement

### LINKED ISSUE #174: Update Flux grammar spec with new option statement

jlapacik:
closes #174

### DIFF
--- a/query/docs/SPEC.md
+++ b/query/docs/SPEC.md
@@ -378,10 +378,11 @@ A _block_ is a possibly empty sequence of statements within matching brace brack
 
 In addition to explicit blocks in the source code, there are implicit blocks:
 
-1. The _universe block_ encompasses all Flux source text.
-2. Each package has a _package block_ containing all Flux source text for that package.
-3. Each file has a _file block_ containing all Flux source text in that file.
-4. Each function literal has its own _function block_ even if not explicitly declared.
+1. The _options block_ is the top-level block for all Flux programs. All option declarations are contained in this block.
+2. The _universe block_ encompasses all Flux source text aside from option statements. It is nested directly inside of the _options block_.
+3. Each package has a _package block_ containing all Flux source text for that package.
+4. Each file has a _file block_ containing all Flux source text in that file.
+5. Each function literal has its own _function block_ even if not explicitly declared.
 
 Blocks nest and influence scoping.
 
@@ -394,14 +395,20 @@ An identifier may change value via assignment within the same block.
 
 Flux is lexically scoped using blocks:
 
-1. The scope of a preassigned identifier is in the universe block.
-2. The scope of an identifier denoting a variable or function at the top level (outside any function) is the package block.
-3. The scope of a package name of an imported package is the file block of the file containing the import declaration.
-4. The scope of an identifier denoting a function argument is the function body.
-5. The scope of a variable assigned inside a function is the innermost containing block.
+1. The scope of an option identifier is the options block.
+2. The scope of a preassigned (non-option) identifier is in the universe block.
+3. The scope of an identifier denoting a variable or function at the top level (outside any function) is the package block.
+4. The scope of a package name of an imported package is the file block of the file containing the import declaration.
+5. The scope of an identifier denoting a function argument is the function body.
+6. The scope of a variable assigned inside a function is the innermost containing block.
 
-An identifier assigned in a block may be reassigned in an inner block.
+An identifier assigned in a block may be reassigned in an inner block with the exception of option identifiers.
 While the identifier of the inner assignment is in scope, it denotes the entity assigned by the inner assignment.
+
+Option identifiers have default assignments that are automatically defined in the _options block_.
+Because the _options block_ is the top-level block of a Flux program, options are visible/available to any and all other blocks.
+However option values may only be reassigned or overridden in the explicit block denoting the main (executable) package.
+Assignment of option identifiers in any non-executable package is strictly prohibited.
 
 The package clause is not a assignment; the package name does not appear in any scope.
 Its purpose is to identify the files belonging to the same package and to specify the default package name for import declarations.
@@ -508,8 +515,54 @@ Examples:
 
 A statement controls execution.
 
-    Statement = VarAssignment | ReturnStatement |
+    Statement = OptionStatement | VarAssignment | ReturnStatement |
                 ExpressionStatement | BlockStatment .
+
+#### Option statements
+
+Options specify a context in which a Flux query is to be run. They define variables
+that describe how to execute a Flux query. For example, the following Flux script sets
+the `task` option to schedule a query to run periodically every hour:
+
+    option task = {
+        name: "mean",
+        every: 1h,
+    }
+
+    from(db:"metrics")
+        |> range(start:-task.every)
+        |> group(by:["level"])
+        |> mean()
+        |> yield(name:"mean")
+
+All options are designed to be completely optional and have default values to be used when not specified.
+Grammatically, an option statement is just a variable assignment preceded by the "option" keyword.
+
+    OptionStatement = "option" VarAssignment
+
+Below is a list of all options that are currently implemented in the Flux language:
+
+* task
+* now
+
+##### task
+
+The `task` option is used by a scheduler to schedule the execution of a Flux query.
+
+    option task = {
+        name: "foo",        // name is required
+        every: 1h,          // task should be run at this interval
+        delay: 10m,         // delay scheduling this task by this duration
+        cron: "0 2 * * *",  // cron is a more sophisticated way to schedule. every and cron are mutually exclusive
+        retry: 5,           // number of times to retry a failed query
+    }
+
+##### now
+
+The `now` option is a function that returns a time value to be used as a proxy for the current system time.
+
+    // Query should execute as if the below time is the current system time
+    option now = () => 2006-01-02T15:04:05Z07:00
 
 #### Return statements
 


query/docs/SPEC.md
```diff
@@ -508,9 +508,55 @@ Examples:
...
-    Statement = VarAssignment | ReturnStatement |
+    Statement = OptionStatement | VarAssignment | ReturnStatement |
                 ExpressionStatement | BlockStatment .
 
+#### Option statements
+
+Options specify a context in which a Flux query is to be run. While they are a part of the formal language definition,
+they are treated as a separate entity from a Flux query with regards to the execution model. For example, the following
+Flux script sets the `task` option to schedule a query to run periodically every hour, while the `now` option implicitly
+specifies the time range (retrospectively) over which the query is to be run:
```
nathanielc: Now does not specify the time range but rather the relative time context for the script. Not sure how to word that...

query/docs/SPEC.md
```diff
@@ -508,9 +508,55 @@ Examples:
...
+
+    option task = {
+        name: "mean",
+        every: 1h,
+    }
+
+    option now = now() - 5m
```
nathanielc: I like the example, this example is not valid but I do not know what the actual syntax will look like yet. So we can leave this in for now but we need to figure out what the correct syntax will be.

The problem with this syntax is that the identifier `now` is both a `time` and a `function`, which is not valid.

query/docs/SPEC.md
```diff
@@ -508,9 +508,55 @@ Examples:
...
+
+    from(db:"metrics")
+        |> range(start:-task.freq)
```
nathanielc: What is `task.freq`?

query/docs/SPEC.md
```diff
@@ -508,9 +508,55 @@ Examples:
...
+        name: "foo",        // name is required
+        every: 1h,          // task should be run at this interval
+        delay: 10m,         // delay scheduling this task by this duration
+        cron: "0 2 * * *",  // cron is a more sophisticated way to schedule. every and cron are mutually exclusive
+        retry: 5,           // number of times to retry a failed query
+    }
+
+##### now
+
+The `now` option sets a time value that is to be used as a proxy for the current time (Go's `time.Now()`).
```
nathanielc: I like the use of `proxy` as a description. Let's not reference Go's docs from within the spec as the user would have to look that up independently. Can you remove the `time.Now` reference and explain that its default value is the current system time?

query/docs/SPEC.md
```diff
@@ -508,9 +508,55 @@ Examples:
...
 A statement controls execution.
 
-    Statement = VarAssignment | ReturnStatement |
+    Statement = OptionStatement | VarAssignment | ReturnStatement |
                 ExpressionStatement | BlockStatment .
 
+#### Option statements
+
+Options specify a context in which a Flux query is to be run. While they are a part of the formal language definition,
+they are treated as a separate entity from a Flux query with regards to the execution model. For example, the following
```
nathanielc: I am not sure what you are referring to by saying they are `separate`? Can you elaborate? 

This spec defines a vocabulary of concepts and it is recommended to use that vocabulary. For example its OK to talk about a `scope` since scopes are explicitly defined in the spec.  In fact the section on scopes should make mentions of the `options` scope.

query/docs/SPEC.md
jlapacik: That's a typo. Should be task.every.

query/docs/SPEC.md
jlapacik: I wasn't defining the `now` option here, I was trying to describe how it affected the behavior of this particular query.  I'll make it clearer.

query/docs/SPEC.md
jlapacik: This is where I defined the now option.

query/docs/SPEC.md
jlapacik: I was trying to say was that while options are part of the Flux grammar, they are treated more like preprocessor directives than program statements. Although I'm sure if that's totally correct. This part needs some work.

query/docs/SPEC.md
```diff
@@ -508,9 +508,57 @@ Examples:
...
+        every: 1h,
+    }
+
+    from(db:"metrics")
+        |> range(start:-task.every)
+        |> group(by:["level"])
+        |> mean()
+        |> yield(name:"mean")
+
+All options are designed to be completely optional and have default values to be used when not specified by the programmer.
```
nathanielc: `when not specified by the ... ` -> `when not specified.`

query/docs/SPEC.md
```diff
@@ -508,9 +508,57 @@ Examples:
...
 A statement controls execution.
 
-    Statement = VarAssignment | ReturnStatement |
+    Statement = OptionStatement | VarAssignment | ReturnStatement |
                 ExpressionStatement | BlockStatment .
 
+#### Option statements
+
+Options specify a context in which a Flux query is to be run. They are similar to preprocessor
+directives in that they describe how to execute a specific Flux query. For example, the following
```
nathanielc: I like the comparison to preprocessor directives. In the Spec we are trying to define everything about the language from first principles. This means we making a comparison to something not already defined in the Spec doesn't work. But it provides a good starting point. 

What is it about preprocessor directives that is similar to options? How do preprocessor directives behave? If we answer those questions it might help us more clearly specify what an option is in Flux.

query/docs/SPEC.md
```diff
@@ -508,9 +508,57 @@ Examples:
...
+    OptionStatement = "option" VarAssignment
+
+Below is a list of all options that are currently implemented in the Flux language:
+
+* task
+* now
+
+##### task
+
+The `task` option is used by a scheduler to schedule the execution of a Flux query. The tentative specification is as follows:
```
nathanielc: Remove `tentative` its OK to write this Spec as if its implemented already. Then we can link to issues to indicate that it is not implemented yet.

query/docs/SPEC.md
```diff
@@ -508,9 +508,57 @@ Examples:
...
+        name: "foo",        // name is required
+        every: 1h,          // task should be run at this interval
+        delay: 10m,         // delay scheduling this task by this duration
+        cron: "0 2 * * *",  // cron is a more sophisticated way to schedule. every and cron are mutually exclusive
+        retry: 5,           // number of times to retry a failed query
+    }
+
+##### now
+
+The `now` option sets a time value that is to be used as a proxy for the current system time. The tentative specification is as follows:
```
nathanielc: Same here, remove `tentative`.

query/docs/SPEC.md
```diff
@@ -508,9 +508,57 @@ Examples:
...
+
+    option now = {
```
nathanielc: Lets specify the now options as a primitive time value instead of an object.

query/docs/SPEC.md
```diff
@@ -406,6 +406,10 @@ While the identifier of the inner assignment is in scope, it denotes the entity
 The package clause is not a assignment; the package name does not appear in any scope.
 Its purpose is to identify the files belonging to the same package and to specify the default package name for import declarations.
 
+An identifier denoting an option (variable preceded by the "option" keyword) has a special scope that does not correspond to a specific block.
+The scope or visibility of an option is defined to be strictly between those of the universe and package blocks. That is, any identifier at the
```
nathanielc: Instead of saying its a special scope between the universe and package blocks, let's define a new `options` block that is defined as being between the universe and the package blocks. 

This brings up a good question too: Can options be defined in packages? If so then the options block would be beneath the package block. But I think options are restricted to being part of the `main` package i.e. the package being executed. I don't think we want to allow packages to define options as they define side effects, and we do not want importing a package to introduce side effects.

pauldix: I would expect options to only be specified in main, never in packages.

jlapacik: Ok so because options need to be visible to all packages, we should keep the options block right below the universe block. Identifiers in the options block will be similar to identifiers in the universe block, in that they cannot overridden in a nested block. The one exception will be the main package. How does this sound?

query/docs/SPEC.md
```diff
@@ -379,9 +379,10 @@ A _block_ is a possibly empty sequence of statements within matching brace brack
 In addition to explicit blocks in the source code, there are implicit blocks:
 
 1. The _universe block_ encompasses all Flux source text.
-2. Each package has a _package block_ containing all Flux source text for that package.
-3. Each file has a _file block_ containing all Flux source text in that file.
-4. Each function literal has its own _function block_ even if not explicitly declared.
+2. The _options block_ encompasses all option declarations (variables preceded by the "option" keyword) as well as all Flux source not preassigned in the universe block.
```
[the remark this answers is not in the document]
jlapacik: I was trying to say that all blocks are contained in the options block aside from the universe block. I.e options block is directly below universe block.

query/docs/SPEC.md
```diff
@@ -395,14 +396,19 @@ An identifier may change value via assignment within the same block.
...
-4. The scope of an identifier denoting a function argument is the function body.
-5. The scope of a variable assigned inside a function is the innermost containing block.
+2. The scope of an option identifier is in the options block.
+3. The scope of an identifier denoting a variable or function at the top level (outside any function) is the package block.
+4. The scope of a package name of an imported package is the file block of the file containing the import declaration.
+5. The scope of an identifier denoting a function argument is the function body.
+6. The scope of a variable assigned inside a function is the innermost containing block.
 
-An identifier assigned in a block may be reassigned in an inner block.
+An identifier assigned in a block may be reassigned in an inner block with the exception of option identifiers.
```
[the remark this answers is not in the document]
jlapacik: Yes.

desa:
@jlapacik is this PR mergable? We've built up a bit of PR cruft in platform and idpe.

PR #185, opened 2018-06-28
fix(errors): Update Fluxd errors

nathanielc:
The goal of this PR is to get useful errors out of the query service HTTP handler. Prior to this change all StatusCodes were 500.

This PR also completes some TODOs about kit/errors and http/errors.


nathanielc:
I am working on a companion IDPE PR as well.

PR #187, opened 2018-06-29
feat(query): result encoders convert truncated storage-layer column labels to human readable values

aanthony1243:
required for an update on idpe: https://github.com/influxdata/idpe/pull/990


nathanielc:
Why do we need this? The `_m` and `_f` names should not make it out of the read service code. Is this for a unit test? If so why not update the test files to use `_measurement` and `_field`?

aanthony1243:
no longer required for 990

PR #191, opened 2018-07-02
feat(query/influxql): modify the encoder to allow more flexibility from the transpiler

jsternberg:
This extra flexibility makes it easier for the transpiler to generate a
specification since the map step can be focused on only generating the
columns related to fields. In particular, it makes it easier to
implement wildcards for tags because the tags will get passed along with
the partition key.


jsternberg:
Accidentally reused the branch name. Going to open a new PR.

Issue #179, opened 2018-06-27
Deduplicate the points while merging in the federated mode


akurilov:
Flux supports the time series aggregation in the federated mode:
https://github.com/influxdata/platform/tree/master/query#federated-mode
However, currently the federated mode is not usable for high availability purposes because the equal points deduplication is not supported. The deduplication functionality should remove the extra equal points (same tags, same timestamps, same values) leaving only unique ones.

akurilov:
PR #195 contains the deduplication functionality as the special no-arg function "dedup()". The dedup function behaves like previously existing "unique" function but compares the time series using all the columns available.

Example demonstrating the difference for the unique/dedup functions on the same data:
- unique:
```bash
curl -XPOST --data-urlencode 'q=from(db: "test_dedup_1") |> range(start: -100h) |> unique()' http://127.0.0.1:8093/v1/query?orgID=00
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string
#partition,false,false,true,true,false,false,true,true
#default,_result,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement
,,0,2018-07-01T04:02:57.742527076Z,2018-07-05T08:02:57.742527076Z,2018-07-05T06:41:52.015548952Z,1,value,x
```
- dedup:
```bash
curl -XPOST --data-urlencode 'q=from(db: "test_dedup_1") |> range(start: -100h) |> dedup()' http://127.0.0.1:8093/v1/query?orgID=00      
#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string
#partition,false,false,true,true,false,false,true,true
#default,_result,,,,,,,
,result,table,_start,_stop,_time,_value,_field,_measurement
,,0,2018-07-01T04:03:06.486571927Z,2018-07-05T08:03:06.486571927Z,2018-07-05T06:41:52.015548952Z,1,value,x
,,0,2018-07-01T04:03:06.486571927Z,2018-07-05T08:03:06.486571927Z,2018-07-05T06:41:55.969062966Z,1,value,x
```

Unit test provided

Issue #171, opened 2018-06-26
Fluxd becomes unresponsive if any of the specified InfluxDB instances is not available


akurilov:
# Steps

1. Deploy several InfluxDB instances w/ configured IFQL
2. Deploy some metrics source, for example Telegraf instance.
3. Deploy nightly Fluxd instance using the following example command:
  `./fluxd --storage-hosts <INFLUXDB_HOST1>:8082,<INFLUXDB_HOST2>:8082,...`
4. Check that Fluxd works normally, querying it with the following example command:
  ```
curl -XPOST --data-urlencode 'q=from(db: "telegraf")   |> range(start: -10m)' http://<FLUXD_HOST>/v1/query?orgID=00
  ```
5. Stop one InfluxDB instance.
6. Repeat the step # 4
7. Start the stopped previously InfluxDB instance.
8. Repeat the step # 4
9. Restart the Fluxd
  ```
  killall fluxd
  ./fluxd --storage-hosts <INFLUXDB_HOST1>:8082,<INFLUXDB_HOST2>:8082,...
  ```
10. Repeat the step # 4

# Expected

4: Fluxd response contains the expected time series data
6: Fluxd response contains the expected time series data
8: Fluxd response contains the expected time series data
10: Fluxd response contains the expected time series data

# Observed

4: Fluxd response contains the expected time series data
6: Fluxd response is empty
8: Fluxd response is empty
10: Fluxd response contains the expected time series data

jsternberg:
This is fixed now since we switched to using grpc.

akurilov:
@jsternberg Does it recover when a storage host becomes back online?

jsternberg:
Yes.

akurilov:
Unfortunately it doesn't work in this way. The observed behaviour remains the same. Should I open new issue either you will reopen this?

jsternberg:
I'll reopen and try it today.

akurilov:
Updated the PR #195 in accordance with GRPC errors in case of a lost connection

jsternberg:
Just to clarify, I'm decently certain that the multi-host functionality is not meant to serve as a high availability mechanism. @nathanielc can confirm, but the expected output should be:

1. Succeeds.
2. Fails.
3. Succeeds.
4. Succeeds.

Does this happen for you?

pauldix:
[...]

PR #195, opened 2018-07-02
Add the deduplication function

akurilov:
Fixed the following issues: #171, #179


.gitignore
```diff
@@ -13,3 +13,5 @@ idpdb.bolt
 /fluxd
 /transpilerd
 /bin
+
+.idea
```
nathanielc: Please add this to your user based .gitignore file and do not include it as part of this repo.

Gopkg.lock
```diff
@@ -373,6 +374,22 @@
...
+
+[[projects]]
+  name = "github.com/influxdata/yarpc"
+  packages = [
+    ".",
+    "codes",
+    "status"
+  ]
+  revision = "b13d74d757bf9ea3e5574a491eb6646bdc4e2a04"
+  version = "v0.0.1"
```
nathanielc: Yamux and yarpc should no longer be used.

query/functions/storage/pb/reader.go
```diff
@@ -124,10 +125,33 @@ func (bi *bockIterator) Do(f func(query.Block) error) error {
...
+				bi.conns[i] = connection {
+					host: h,
+					conn: cc,
+					client: cl,
+				}
+				stream, err = cl.Read(bi.ctx, &req)
+				if err != nil {
+					println("Read retry failed @ " + h + "")
+					continue
+				}
```
nathanielc: I do not follow why this logic is necessary. gRPC should be reconnecting in the background for us, if it is not then we have probably missed a DialOption or something. Can you clarify the situation in which this code was necessary to force a reconnect?

query/functions/dedup.go
```diff
@@ -0,0 +1,198 @@
...
+		d:      d,
+		cache:  cache,
+	}
+}
+
+func (t *dedupTransformation) RetractBlock(id execute.DatasetID, key query.PartitionKey) error {
+	return t.d.RetractBlock(key)
+}
+
+func (t *dedupTransformation) Process(id execute.DatasetID, b query.Block) error {
```
nathanielc: I see in #179 you have shown how this function is different from the `unique` function. The main difference is that the unique function selects rows that are unique considering a single column, while this dedup function checks if a row is unique considering all columns. I think we should update the unique function to accept a list of columns and then it can do what you want without requiring a separate function. You could also add a parameter to the unique function to indicate that all columns should be used. Thoughts?

pauldix: I think I agree with this, but it can be a bit tricky. Because you want the `columns` to actually just be the group key and the `_time` column. You don't want "all columns" because `_value` shouldn't come into play as well as any columns that aren't in the group key. You also wouldn't want to have to specify the `columns` explicitly. Rather just specify that you want to use the group key columns + some other set.

nathanielc: If all you want to do is use the `_time` column to dedup that can be done already using the unique function, `unique(column:"_time")`. The group key columns are all the same by definition.

nathanielc: I recommend we update the unique function to take a list of columns.

akurilov:
Closing, created 2 separated PRs instead of this:
https://github.com/influxdata/platform/pull/365
https://github.com/influxdata/platform/pull/366/files

PR #196, opened 2018-07-03
Evaluate semantic graph and return results of side effects

jlapacik:
evaluate CallExpression to determine if function has side effects
return all side effects
add Equal method to Value interface

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -61,7 +61,7 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
 		return nil, err
 	}
 
-	if err := interpreter.Eval(semProg, interpScope); err != nil {
+	if _, err := interpreter.Eval(semProg, interpScope); err != nil {
 		return nil, err
 	}
 	spec := qd.ToSpec()
@@ -92,11 +92,24 @@ func RegisterBuiltIn(name, script string) {
 // RegisterFunction adds a new builtin top level function.
 func RegisterFunction(name string, c CreateOperationSpec, sig semantic.FunctionSignature) {
 	f := function{
-		t:            semantic.NewFunctionType(sig),
-		name:         name,
-		createOpSpec: c,
+		t:             semantic.NewFunctionType(sig),
+		name:          name,
+		createOpSpec:  c,
+		hasSideEffect: false,
 	}
-	RegisterBuiltInValue(name, f)
+	RegisterBuiltInValue(name, &f)
+}
+
+// RegisterFunctionWithSideEffect adds a new builtin top level function that produces side effects.
+// For example, the builtin functions yield(), toKafka(), and toHTTP() all produce side effects.
+func RegisterFunctionWithSideEffect(name string, c CreateOperationSpec, sig semantic.FunctionSignature) {
+	f := function{
+		t:             semantic.NewFunctionType(sig),
+		name:          name,
+		createOpSpec:  c,
+		hasSideEffect: true,
+	}
+	RegisterBuiltInValue(name, &f)
 }
 
 // RegisterBuiltInValue adds the value to the builtin scope.
@@ -210,6 +223,23 @@ func (t TableObject) Array() values.Array {
 func (t TableObject) Object() values.Object {
 	return t
 }
+func (t TableObject) Equal(rhs values.Value) bool {
+	if t.Type() != rhs.Type() {
+		return false
+	}
+	r := rhs.Object()
+	if t.Len() != r.Len() {
+		return false
+	}
+	for _, k := range t.keys() {
+		val1, ok1 := t.Get(k)
+		val2, ok2 := r.Get(k)
+		if !ok1 || !ok2 || !val1.Equal(val2) {
+			return false
+		}
+	}
+	return true
+}
 func (t TableObject) Function() values.Function {
 	panic(values.UnexpectedKind(semantic.Object, semantic.Function))
 }
@@ -225,6 +255,10 @@ func (t TableObject) Get(name string) (values.Value, bool) {
 	default:
 		return nil, false
 	}
+}
+
+func (t TableObject) keys() []string {
+	return []string{tableIDKey, tableKindKey, tableParentsKey}
 }
 
 func (t TableObject) Set(name string, v values.Value) {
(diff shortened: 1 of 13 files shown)


query/compile.go
```diff
@@ -210,6 +223,31 @@ func (t TableObject) Array() values.Array {
...
+	r2, ok2 := r.Get(tableKindKey)
+	r3, ok3 := r.Get(tableParentsKey)
+
+	if !ok1 || !ok2 || !ok3 {
+		return false
+	}
+
+	t1, _ := r.Get(tableIDKey)
+	t2, _ := r.Get(tableKindKey)
+	t3, _ := r.Get(tableParentsKey)
```
nathanielc: We should not assume that the lhs `t` is valid either. Maybe just loop over the different keys and recurse that way?

query/compile.go
```diff
@@ -409,7 +448,17 @@ func (f function) Object() values.Object {
 	panic(values.UnexpectedKind(semantic.Function, semantic.Object))
 }
 func (f function) Function() values.Function {
```
nathanielc: Change this to a pointer receiver and return `f` directly. Otherwise the pointer returned is always to a copy of `f` which is not what we want.

query/compile.go
```diff
@@ -368,10 +406,11 @@ func (d *queryDomain) ToSpec() *Spec {
...
-	createOpSpec CreateOperationSpec
-	qd           *queryDomain
+	name          string
+	t             semantic.Type
+	createOpSpec  CreateOperationSpec
+	qd            *queryDomain
+	hasSideEffect bool
 }
 
 func (f function) Type() semantic.Type {
```
nathanielc: Change all the methods on `function` to be pointer receivers so that Go knows it does not have to make a copy to call the method. Sometimes Go can inspect the function body and determine that f is not modified and it can skip the copy step, but since we are using pointers now lets just always use pointer receivers.

query/compiler/runtime.go
```diff
@@ -1121,7 +1121,17 @@ func (f functionValue) Object() values.Object {
 	panic(values.UnexpectedKind(semantic.Function, semantic.Object))
 }
 func (f functionValue) Function() values.Function {
```
nathanielc: Same here about pointer receivers.

query/interpreter/interpreter.go
```diff
@@ -10,50 +10,56 @@ import (
...
+			itrp.values = append(itrp.values, val)
+		}
 	}
 	return nil
 }
 
-func (itrp interpreter) doStatement(stmt semantic.Statement, scope *Scope) error {
-	scope.SetReturn(values.InvalidValue)
+func (itrp *interpreter) doStatement(stmt semantic.Statement, scope *Scope) (values.Value, error) {
+	scope.SetReturn(&values.InvalidValue)
```
nathanielc: Why does this need to be a pointer?

query/interpreter/interpreter.go
```diff
@@ -10,50 +10,56 @@ import (
...
 	switch s := stmt.(type) {
 	case *semantic.OptionStatement:
-		if err := itrp.doStatement(s.Declaration, scope); err != nil {
-			return err
-		}
+		return itrp.doStatement(s.Declaration, scope)
 	case *semantic.NativeVariableDeclaration:
 		if err := itrp.doVariableDeclaration(s, scope); err != nil {
-			return err
+			return nil, err
```
nathanielc: I think we need to return the value that was assigned to a var. This way assigning to a var still catches the side effects.

query/interpreter/interpreter.go
```diff
@@ -10,50 +10,56 @@ import (
...
+		if val != nil {
+			itrp.values = append(itrp.values, val)
+		}
 	}
 	return nil
 }
 
-func (itrp interpreter) doStatement(stmt semantic.Statement, scope *Scope) error {
-	scope.SetReturn(values.InvalidValue)
+func (itrp *interpreter) doStatement(stmt semantic.Statement, scope *Scope) (values.Value, error) {
```
nathanielc: We should comment what the return value of doStatement are.

query/interpreter/interpreter.go
```diff
@@ -298,17 +304,23 @@ func (itrp interpreter) doCall(call *semantic.CallExpression, scope *Scope) (val
 	}
 
 	// Check if the function is an interpFunction and rebind it.
-	if af, ok := f.(function); ok {
-		af.itrp = itrp
+	if af, ok := f.(*function); ok {
+		//af.itrp = itrp
```
nathanielc: Does this need to be fixed?

query/interpreter/interpreter.go
```diff
@@ -343,6 +355,20 @@ func NewScopeWithValues(values map[string]values.Value) *Scope {
 	}
 }
 
+func (s *Scope) level() int {
```
nathanielc: Is this code used?

query/interpreter/interpreter.go
```diff
@@ -482,7 +508,19 @@ func (f function) Object() values.Object {
...
+	return &f
+}
+func (f *function) Equal(rhs values.Value) bool {
+	if f.Type() != rhs.Type() {
+		return false
+	}
+	v, ok := rhs.(*function)
+	return ok && (f == v)
+}
+func (f function) HasSideEffect() bool {
```
nathanielc: Pointer receiver.

query/interpreter/interpreter.go
jlapacik: No I'll remove.

query/interpreter/interpreter.go
jlapacik: It doesn't. That was from some debugging. Will remove.

query/compile.go
```diff
@@ -210,6 +223,31 @@ func (t TableObject) Array() values.Array {
...
 }
+func (t TableObject) Equal(rhs values.Value) bool {
+	if t.Type() != rhs.Type() {
+		return false
+	}
+	r := rhs.Object()
+	if t.Len() != r.Len() {
+		return false
+	}
+	r1, ok1 := r.Get(tableIDKey)
```
nathanielc: Can you change this to use a loop like the Equal methods does for Object? That should make it much simpler.

query/compile.go
```diff
@@ -266,9 +304,12 @@ func builtIns(qd *queryDomain) (map[string]values.Value, semantic.DeclarationSco
 	scope := make(map[string]values.Value, len(builtinScope))
 	for k, v := range builtinScope {
 		if v.Type().Kind() == semantic.Function {
-			if f, ok := v.Function().(function); ok {
-				f.qd = qd
-				v = f
+			if f, ok := v.Function().(*function); ok {
+				// TODO(Josh) Cannot use f directly here, need to make a copy.
+				// Otherwise tests in functions package will break. Need to figure out why that is.
```
nathanielc: The tests break because they are modifying the same `function` and so only the last test to run wins the race. By making a copy for each test they can pass.

PR #347, opened 2018-07-03
remove notion of query domain

jlapacik:


query/compile.go
```diff
@@ -304,7 +315,7 @@ func builtIns(qd *queryDomain) (map[string]values.Value, semantic.DeclarationSco
 				// Must make separate copy of f. Otherwise function
 				// tests will modify same f and tests will break.
 				newfunc := f.copy()
-				newfunc.qd = qd
+				newfunc.id = id
```
nathanielc: This whole `if` block should be removed at this point.

The `id` must be assigned in the `buildSpec` method. This way mean that if you call a function twice the operation would have the same ID.

query/compile.go
```diff
@@ -61,17 +60,29 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
...
 
 	if o.verbose {
 		log.Println("Query Spec: ", Formatted(spec, FmtJSON))
 	}
 	return spec, nil
 }
 
+func toSpec(operations []values.Value) *Spec {
+	spec := new(Spec)
+	visited := make(map[OperationID]bool)
```
nathanielc: We need to assign the IDs as we build the spec. It should be possible to change this map to be `map[*TableObject]bool`, then IDs can be assigned by iterating of the visited map or something similar.

query/compile.go
```diff
@@ -371,44 +382,11 @@ func (a *Administration) AddParent(np TableObject) {
...
-	spec := new(Spec)
-	visited := make(map[OperationID]bool)
-	for _, t := range d.operations {
-		t.buildSpec(spec, visited)
-	}
-	return spec
-}
-
 type function struct {
+	id            *int
```
nathanielc: ID is a property of an operation not a function.

query/interpreter/interpreter.go
```diff
@@ -313,7 +313,7 @@ func (itrp *interpreter) doCall(call *semantic.CallExpression, scope *Scope) (va
 	if err != nil {
 		return nil, err
 	}
-	if f.HasSideEffect() {
+	if f.HasSideEffect() && !scope.isTopLevel() {
```
nathanielc: Why was this added back in?

jlapacik: I need to put a comment in. If a side effecty function is called at the top-level block, then we should not explicitly capture it, as all top-level expressions are captured regardless of behavior.

query/compile.go
jlapacik: Ah yes I like that.

query/interpreter/interpreter.go
jlapacik: Of course there might be a better way to do it.

nathanielc: Capturing it twice should not matter, as the `visited` map will detect that and ignore it.

query/compile.go
```diff
@@ -240,37 +317,39 @@ func (t TableObject) Equal(rhs values.Value) bool {
...
 	switch name {
-	case tableIDKey:
-		return values.NewStringValue(string(t.ID)), true
 	case tableKindKey:
 		return values.NewStringValue(string(t.Kind)), true
 	case tableParentsKey:
 		return t.Parents, true
 	default:
-		return nil, false
+		return t.Args.Get(name)
```
nathanielc: What is this used for?

query/compile.go
```diff
@@ -471,29 +497,27 @@ func (f *function) Call(argsObj values.Object) (values.Value, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	if a.parents.Len() > 1 {
 		// Always add parents in a consistent order
 		a.parents.Sort(func(i, j values.Value) bool {
-			return i.(TableObject).ID < j.(TableObject).ID
+			// TODO(Josh): Below comparison might not guarantee consistency
+			return i.(*TableObject).Kind < j.(*TableObject).Kind
```
nathanielc: This should be good enough for now. Once we have consistent order for iterating Flux Objects this won't be an issue.

jlapacik: Tests might fail at random because of this.

query/compile.go
jlapacik: This is used to correctly test for equality between two TableObjects now that TableObjects don't have IDs. The equality check is used to remove duplicate operations before building the spec.

nathanielc: I see, in that case can you make the `Args` field private i.e. `args` since it does not need to be exposed. 

Seems like we need a way to test if operations  specs are equal. Using the args works for now, but its a round about solution. Can you create an issue to track updating the OperationSpec interface to have an Equal method and add a TODO here to remove `args` once it is closed?

jlapacik: #373

Issue #229, opened 2018-07-03
SPEC: Null


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#332  

Is null usable (i.e. a keyword or value) inside IFQL?

PR #349, opened 2018-07-04
Add tests for null keyword behavior

adamperlin:
Adds test cases to address #229. Results: `null` currently has no special meaning in flux and is a valid identifier.


query/querytest/test_cases/null_as_ident.flux
```diff
@@ -0,0 +1,3 @@
+null = "test"
+from(db: null)
```
aanthony1243: This doesn't work for a few reasons, first being that a DB is a data source and a null DB doesn't make much sense.   Second, the test framework replaces this line with the CSV data source so it's never executed on.  

after range, add   a   `|> filter(fn: (r) => r._value == null)` 

I would expect this one to error out, actually.

query/querytest/test_cases/null_as_value.flux
```diff
@@ -0,0 +1,3 @@
+db = null
```
aanthony1243: same thing here, let's put this into a filter function.

adamperlin:
Note: rebased against master

query/querytest/test_cases/null_as_ident.flux
```diff
@@ -0,0 +1,4 @@
+null = 318324641792
```
aanthony1243: This is  demonstrating a current behavior, but tests should match desired behavior and not actual behavior.  The second test effectively shows that null is not a reserved keyword.  What should be happening in this case, is an error saying we can't assign a value to null.   I think it's OK to just drop this test.

adamperlin: ok, sounds good.

PR #350, opened 2018-07-05
refactor(query/influxql): modify the transpiler compilation tests to remove skip

jsternberg:
The transpiler compilation tests will now not allow skip to be
specified. Instead, it must return an error message that starts with
`unimplemented` and then the reason will be used as the skip message.

This way, it will be easier to identify the failing tests in the
transpiler. In the previous method, it was possible for a test to be
marked as skip, but for the transpiler to return the wrong error message
because the test did not differentiate between an unimplemented error
message and an incorrect error message.


jsternberg:
Updated. Yes, I changed some of the code so that it would return error messages or unimplemented. Otherwise, it was returning the wrong error message for some of these which made the test change not work.

Issue #313, opened 2018-07-03
Window bucket start boundary should default to `now`


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#205  

If the user is windowing by 1m and the current time is t=30s off the minute then time boundaries should be at t +1m , t +2m, ...  t + Nm.

https://github.com/influxdata/influxdb/issues/4680

nathanielc:
We have changed this design. By default windows start on even intervals of the `every` value. If a user wants the boundary to start exactly when the query is run they can use the `now` option as the value for the start parameter.

For example:

```
from(db: "test")
    |> range(start:2018-05-22T19:53:00Z)
    |> window(start:now,every: 1m) 
```

See #37 for adding the `now` option.

PR #354, opened 2018-07-06
Add initial window bound offset test cases for flux

adamperlin:
Addresses #313.  Adds a test case to document the behavior of start boundaries with offsets smaller than the `every` interval in the flux `window` function. Essentially, if the `start` parameter to the `window` function is offset from the `every` interval (for example, we're grouping at intervals of 1 minute, but the `start` time is 30s off the minute), then the window boundaries are shifted according to that offset.


query/querytest/test_cases/window_start_bound.in.csv
```diff
@@ -0,0 +1,10 @@
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string
+#partition,false,false,false,false,false,false,true,true,true
+#default,_result,,,,,,,,
+,result,table,_start,_stop,_time,_value,_field,_measurement,host
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:53:26Z,63.053321838378906,used_percent,mem,host.local
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:54:36Z,62.71536350250244,used_percent,mem,host.local
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:55:46Z,62.38760948181152,used_percent,mem,host.local
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:56:56Z,62.74595260620117,used_percent,mem,host.local
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:57:06Z,62.78183460235596,used_percent,mem,host.local
+,,147,2018-05-22T19:53:24.421470485Z,2018-05-22T19:54:24.421470485Z,2018-05-22T19:58:16Z,62.46745586395264,used_percent,mem,host.local
```
nathanielc: The _stop times are not large enough to contain the _time.
 Can you update them so that `_start <= _time < _stop`?

adamperlin: ah yes, forgot to do that when I modified the _time column

adamperlin:
Fixed test data time ranges and added the requested second test. Let me know what you think

PR #360, opened 2018-07-06
Add test to make sure relevant _time values are returned when using top() selector

adamperlin:
InfluxQL used to return Unix zero time instead of corresponding timestamps when using selectors like `top` and grouping by a key other than time. This behavior is documented [this](https://github.com/influxdata/influxdb/issues/4680) influxdb issue. It seems this bug has been fixed in flux, as is demonstrated in the included tests. When selecting with the `top()` function and using `group()`  with a specified tag, the expected timestamps for points are returned.


adamperlin:
Note: ci appears to be failing because of an issue fetching a dependency. Possible to re-run?

query/querytest/test_cases/selector_preserve_time.flux
```diff
@@ -0,0 +1,4 @@
+from(db:"test")
+	|> range(start:-5m)
+	|> top(n:3)
+	|> group(by:["host"])
```
nathanielc: Why does the test need a `group` at the end? It is best if the test is as small as possible to test the behavior of the selector. That way when this test breaks we can be confident we broke selectors and not something unrelated.

adamperlin: I included the group because the original InfluxQl issue was triggered when the query had a `GROUP BY user_tag`. Can definitely be removed, just wanted to try and make the test as similar as possible.

nathanielc: OK yeah, that seems fair.

Issue #352, opened 2018-07-05
Panic when using `max` on strings in flux


adamperlin:
### Stack Trace
```
panic: interface conversion: interface is nil, not execute.DoStringRowSelector
goroutine 21 [running]:
runtime/debug.Stack(0x14bba60, 0x14b9b40, 0xc42034cb80)
	/usr/local/Cellar/go/1.10.3/libexec/src/runtime/debug/stack.go:24 +0xa7
github.com/influxdata/platform/query/execute.(*poolDispatcher).Start.func1.1(0xc420520a50)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/dispatcher.go:67 +0xb9
panic(0x14b9b40, 0xc42034cb80)
	/usr/local/Cellar/go/1.10.3/libexec/src/runtime/panic.go:502 +0x229
github.com/influxdata/platform/query/execute.(*rowSelectorTransformation).Process.func1(0x15bff40, 0xc420218050, 0x1543801, 0xc42034cb40)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/selector.go:183 +0x258
github.com/influxdata/platform/query/execute.(*ColListBlock).Do(0xc420218050, 0xc42034cb40, 0x0, 0x0)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/block.go:602 +0x3c
github.com/influxdata/platform/query/execute.(*rowSelectorTransformation).Process(0xc4200423c0, 0x255fce4fb6a41b39, 0x928afa2d6a496c96, 0x15bc6c0, 0xc420218050, 0x1, 0x2)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/selector.go:172 +0x26b
github.com/influxdata/platform/query/execute.processMessage(0x15bc800, 0xc4200423c0, 0x15b72c0, 0xc4201be260, 0x3, 0x0, 0x0)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/transport.go:199 +0x200
github.com/influxdata/platform/query/execute.(*consecutiveTransport).processMessages(0xc420221500, 0xa)
	/Users/adamperlin/go/src/github.com/influxdata/platform/query/execute/transport.go:156 +0xa1
github.com/influxdata/platform/query/execute.(*consecutiveTransport).(github.com/influxdata/platform/query/execute.processMessages)-fm(0xa)
[...]

PR #363, opened 2018-07-06
Fix panic when unimplemented selector function is used

adamperlin:
This addresses #352. Adds a nil check before running row selector functions to prevent a nil interface panic.  Allows for `max` and `min`  to be called on strings in flux without crashing the executor.


query/execute/selector.go
```diff
@@ -169,6 +170,13 @@ func (t *rowSelectorTransformation) Process(id DatasetID, b query.Block) error {
 		return fmt.Errorf("unsupported selector type %v", valueCol.Type)
 	}
 
+	// if rower has a nil value, this means that the row selector doesn't
+	// yet have an implementation
+
+	if rower == nil {
+		return fmt.Errorf("unimplemented row selector function")
```
nathanielc: Can you add information about the type of the column being selected and the type of selector being used?

adamperlin: yes.

adamperlin:
Alright, I updated the error message

query/querytest/query_test.go
```diff
@@ -19,7 +19,7 @@ var skipTests = map[string]string{
 	"derivative":                "derivative not supported by influxql (https://github.com/influxdata/platform/issues/93)",
 	"filter_by_tags":            "arbitrary filtering not supported by influxql (https://github.com/influxdata/platform/issues/94)",
 	"window_group_mean_ungroup": "error in influxql: failed to run query: timeValue column \"_start\" does not exist (https://github.com/influxdata/platform/issues/97)",
-	"string_max":                "panic because no max() implementation for string (https://github.com/influxdata/platform/issues/352)",
+	"string_max":                "error: unimplemented row selector type",
```
nathanielc: Does this test still pass with the change in the error string?

query/functions/query_test.go
```diff
@@ -112,20 +120,24 @@ func queryTranspileTester(t *testing.T, transpiler query.Transpiler, qs query.Qu
...
-func QueryTestCheckSpec(t *testing.T, qs query.QueryService, spec *query.Spec, inputFile, want string, enc query.MultiResultEncoder) {
+func QueryTestCheckSpec(t *testing.T, qs query.QueryService, spec *query.Spec, inputFile, want string, enc query.MultiResultEncoder) error {
 	t.Helper()
+
 	querytest.ReplaceFromSpec(spec, inputFile)
 
 	got, err := querytest.GetQueryEncodedResults(qs, spec, inputFile, enc)
 	if err != nil {
 		t.Errorf("failed to run query: %v", err)
+		return fmt.Errorf("Error running query")
```
aanthony1243: use `errors.New("Error running query")`  since you aren't using the format-string feature of Errorf

query/functions/query_test.go
```diff
@@ -2,28 +2,29 @@ package functions_test
...
+
+	"github.com/andreyvit/diff"
 )
 
 var skipTests = map[string]string{
 	"derivative":                "derivative not supported by influxql (https://github.com/influxdata/platform/issues/93)",
+	"filter_by_tags":            "arbitrary filtering not supported by influxql (https://github.com/influxdata/platform/issues/94)",
 	"window_group_mean_ungroup": "error in influxql: failed to run query: timeValue column \"_start\" does not exist (https://github.com/influxdata/platform/issues/97)",
-	"string_max":                "panic because no max() implementation for string (https://github.com/influxdata/platform/issues/352)",
+	"string_max":                "error: invalid use of function: *functions.MaxSelector has no implementation for type string",
```
aanthony1243: @nathanielc  technically this test "passes" because it has expected behavior according to the code.  But we want to change this behavior.  

@adamperlin  can we reference a proper issue here, maybe the one that prompted this work that was about sorting/ordering for strings?

nathanielc: This test case looks good now. When I commented the error message was still the old message so I expected it to fail.

Issue #340, opened 2018-07-03
Add ignore to group function


adamperlin:
From ifql created by [pauldix](https://github.com/pauldix) : influxdata/ifql#83  

The ignore parameter should take an array of tag keys. This is to merge the series together by all tag keys except those that appear in the ignore list. The parser should validate that ignore and keys are mutually exclusive.

aanthony1243:
covered by `except` argument

PR #364, opened 2018-07-07
Add group except test for flux

adamperlin:
Adds an additional test case for grouping with the `except` argument. See #340.


adamperlin:
Changes made.

query/querytest/test_cases/group_except.flux
```diff
@@ -0,0 +1,4 @@
+from(db:"test")
+    |> range(start:-5m)
+    |> group(except:["_measurement"])
```
nathanielc: I think we want to add `_time` and `_value` to the except list otherwise every point is in its own group which makes the test hard to follow.

adamperlin: Agreed - done.

PR #365, opened 2018-07-09
Issue #171 resolution - Add the connection recovery function

akurilov:
gRPC doesn't seem to reconnect automatically.


jsternberg:
Sorry for not looking at this for so long. The project has been moving very fast so we missed this. If this is still a problem, can you file an issue? Thanks.

PR #366, opened 2018-07-09
Issue #179 resolution - Add the deduplication function

akurilov:
The separate "dedup" function may be useful if it's not known which columns are returned by the query.
"Unique" function doesn't support a flag forcing to use all returned columns. Also, separate "dedup" function allows to make the changes incremental and not affecting the existing functionality.


nathanielc:
>"Unique" function doesn't support a flag forcing to use all returned columns. 

Can you add a parameter to the unique function to support this behavior? Something like `unique(all:true)`?

desa:
@nathanielc Is this mergable?

jsternberg:
This PR needs to be made against `influxdata/flux` instead now. Sorry for the delay. I'm going to close this because this PR itself isn't actionable.

PR #382, opened 2018-07-10
chore: Rename parition key to group key

nathanielc:


nathanielc:
@goller The swagger doc was already up-to-date since we knew this change was coming

PR #384, opened 2018-07-11
feat(query/influxql): Add group-by time + offset support

aanthony1243:
previous versions only supported the first parameter to time() that set the window size.  This version supports the second parameter, which shifts the offset a fixed amount from the epoch


query/influxql/spectests/aggregates_with_window.go
```diff
@@ -79,6 +79,7 @@ func init() {
 							Spec: &functions.WindowOpSpec{
 								Every:              query.Duration(time.Minute),
 								Period:             query.Duration(time.Minute),
+								Start:              query.Time{},
```
jsternberg: These are unnecessary. The default value already covers them if you use the above. Also, based on my reading of the code, this is causing the test cases to fail currently.

query/influxql/group.go
```diff
@@ -267,6 +267,7 @@ type groupCursor struct {
 
 func (gr *groupInfo) group(t *transpilerState, in cursor) (cursor, error) {
 	var windowEvery time.Duration
+	var windowOffset time.Duration
```
jsternberg: This should be a `query.Time` and then you can set it whenever you get the offset parameter. That way, you can use the zero struct so you can leave the start parameter unassigned easily if there is no window offset. That way, you also don't have to add `query.Time{Absolute: time.Unix(0, 0)}` to `window` calls when the offset isn't set.

query/influxql/group.go
```diff
@@ -299,7 +300,7 @@ func (gr *groupInfo) group(t *transpilerState, in cursor) (cursor, error) {
 				} else {
 					windowEvery = lit.Val
 					if len(expr.Args) == 2 {
-						return nil, errors.New("unimplemented: group by offsets")
+						windowOffset = expr.Args[1].(*influxql.DurationLiteral).Val
```
jsternberg: This needs protection. There's no protection code for if the second call isn't a duration literal and no part of the parser verifies that. In general, remember that influxql will mostly accept anything as long as it fits the syntax. If something is using an interface, assume that it can be anything that fulfills that interface.

Look at the code in influxdb for this. I didn't copy it mostly because it wasn't relevant yet, but it should mostly be duplicated now that we're implementing this functionality: https://github.com/influxdata/influxdb/blob/7dc7efd5014bfba6492801981c21718912fa1bad/query/compile.go#L893

query/influxql/group.go
```diff
@@ -320,9 +321,11 @@ func (gr *groupInfo) group(t *transpilerState, in cursor) (cursor, error) {
 	}, in.ID())
 
 	if windowEvery > 0 {
+		windowStart := time.Unix(0, 0)
```
nathanielc: I guess this works but it feels unintuitive why it works. Perhaps are clearer solution is to use the actual start time of the query truncate it with the group by time and then add the offset?

pauldix:
Just a broader question here, but what is the offset for? When and why would a user need this?

query/influxql/group.go
aanthony1243: @jsternberg  I'm not seeing a direct way to get the query start time via the types available in `group.go`  any tips?

jsternberg: The start time isn't really available here.

nathanielc:
The feature is to change the window boundaries to be aligned to other times besides round intervals of the group by time. This is common for time zone manipulation.

In addition its a commonly used feature in InfluxQL so we are updating the tranpsiler to support it.

PR #388, opened 2018-07-11
set now time at query spec level

jlapacik:
This commit adds a `now` field to a query spec as well as `Now` fields to logical and physical plans. These fields represent the current system time that a query will execute with respect to. This commit is in preparation for adding an interface for getting and setting Flux options - in particular the **now** option.

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -19,6 +19,7 @@ const (
 	TableParameter  = "table"
 	tableKindKey    = "kind"
 	tableParentsKey = "parents"
+	nowOption       = "now"
 	//tableSpecKey    = "spec"
 )
 
@@ -35,7 +36,8 @@ type options struct {
 }
 
 // Compile evaluates a Flux script producing a query Spec.
-func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
+// now parameter must be non-zero, that is the default now time should be set before compiling.
+func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spec, error) {
 	o := new(options)
 	for _, opt := range opts {
 		opt(o)
@@ -49,8 +51,10 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
 	s, _ = opentracing.StartSpanFromContext(ctx, "compile")
 	defer s.Finish()
 
-	scope, decls := builtIns()
-	interpScope := interpreter.NewScopeWithValues(scope)
+	itrp := NewInterpreter()
+	itrp.SetOption(nowOption, nowFunc(now))
+
+	_, decls := builtIns(itrp)
 
 	// Convert AST program to a semantic program
 	semProg, err := semantic.New(astProg, decls)
@@ -58,11 +62,10 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
 		return nil, err
 	}
 
-	operations, err := interpreter.Eval(semProg, interpScope)
-	if err != nil {
+	if err := itrp.Eval(semProg); err != nil {
 		return nil, err
 	}
-	spec := toSpec(operations)
+	spec := toSpec(itrp)
 
 	if o.verbose {
 		log.Println("Query Spec: ", Formatted(spec, FmtJSON))
@@ -70,7 +73,38 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
 	return spec, nil
 }
 
-func toSpec(stmtVals []values.Value) *Spec {
+// NewInterpreter returns an interpreter instance with
+// pre-constructed options and global scopes.
+func NewInterpreter() *interpreter.Interpreter {
+	options := make(map[string]values.Value, len(builtinOptions))
+	globals := make(map[string]values.Value, len(builtinScope))
+
+	for k, v := range builtinScope {
+		globals[k] = v
+	}
+
+	for k, v := range builtinOptions {
+		options[k] = v
+	}
+
+	return interpreter.NewInterpreter(options, globals)
+}
+
+func nowFunc(now time.Time) values.Function {
+	timeVal := values.NewTimeValue(values.ConvertTime(now))
+	ftype := semantic.NewFunctionType(semantic.FunctionSignature{
+		ReturnType: semantic.Time,
+	})
+	call := func(args values.Object) (values.Value, error) {
+		return timeVal, nil
+	}
+	sideEffect := false
+	return values.NewFunction(nowOption, ftype, call, sideEffect)
+}
+
+func toSpec(itrp *interpreter.Interpreter) *Spec {
+	operations := itrp.SideEffects()
+
 	ider := &ider{
 		id:     0,
 		lookup: make(map[*TableObject]OperationID),
@@ -78,9 +112,9 @@ func toSpec(stmtVals []values.Value) *Spec {
 
 	spec := new(Spec)
 	visited := make(map[*TableObject]bool)
-	nodes := make([]*TableObject, 0, len(stmtVals))
+	nodes := make([]*TableObject, 0, len(operations))
 
-	for _, val := range stmtVals {
+	for _, val := range operations {
 		if op, ok := val.(*TableObject); ok {
 			dup := false
 			for _, node := range nodes {
@@ -95,12 +129,23 @@ func toSpec(stmtVals []values.Value) *Spec {
 			}
 		}
 	}
+
+	// now option is Time value
+	nowValue, _ := itrp.Option(nowOption).Function().Call(nil)
+	spec.Now = nowValue.Time().Time()
+
 	return spec
 }
 
 type CreateOperationSpec func(args Arguments, a *Administration) (OperationSpec, error)
 
 var builtinScope = make(map[string]values.Value)
+
+// TODO(Josh): Default option values should be registered similarly to built-in
+// functions. Default options should be registered in their own files
+// (or in a single file) using the RegisterBuiltInOption function which will
+// place the resolved option value in the following map.
+var builtinOptions = make(map[string]values.Value)
 var builtinDeclarations = make(semantic.DeclarationScope)
 
 // list of builtin scripts
@@ -148,6 +193,17 @@ func RegisterBuiltInValue(name string, v values.Value) {
 	}
 	builtinDeclarations[name] = semantic.NewExternalVariableDeclaration(name, v.Type())
 	builtinScope[name] = v
+}
+
+// RegisterBuiltInOption adds the value to the builtin scope.
+func RegisterBuiltInOption(name string, v values.Value) {
+	if finalized {
+		panic(errors.New("already finalized, cannot register builtin option"))
+	}
+	if _, ok := builtinOptions[name]; ok {
+		panic(fmt.Errorf("duplicate registration for builtin option %q", name))
+	}
+	builtinOptions[name] = v
 }
 
 // FinalizeBuiltIns must be called to complete registration.
@@ -373,16 +429,12 @@ func BuiltIns() (map[string]values.Value, semantic.DeclarationScope) {
 	if !finalized {
 		panic("builtins not finalized")
 	}
-	return builtIns()
+	return builtIns(NewInterpreter())
 }
 
-func builtIns() (map[string]values.Value, semantic.DeclarationScope) {
+func builtIns(itrp *interpreter.Interpreter) (map[string]values.Value, semantic.DeclarationScope) {
 	decls := builtinDeclarations.Copy()
-	scope := make(map[string]values.Value, len(builtinScope))
-	for k, v := range builtinScope {
-		scope[k] = v
-	}
-	interpScope := interpreter.NewScopeWithValues(scope)
+
 	for name, script := range builtins {
 		astProg, err := parser.NewAST(script)
 		if err != nil {
@@ -393,11 +445,11 @@ func builtIns() (map[string]values.Value, semantic.DeclarationScope) {
 			panic(errors.Wrapf(err, "failed to create semantic graph for builtin %q", name))
 		}
 
-		if _, err := interpreter.Eval(semProg, interpScope); err != nil {
+		if err := itrp.Eval(semProg); err != nil {
 			panic(errors.Wrapf(err, "failed to evaluate builtin %q", name))
 		}
 	}
-	return scope, decls
+	return itrp.GlobalScope().Values(), decls
 }
 
 type Administration struct {
--- a/query/control/controller.go
+++ b/query/control/controller.go
@@ -114,7 +114,7 @@ func (c *Controller) compileQuery(q *Query, queryStr string) error {
 	if !q.tryCompile() {
 		return errors.New("failed to transition query to compiling state")
 	}
-	spec, err := query.Compile(q.compilingCtx, queryStr, query.Verbose(c.verbose))
+	spec, err := query.Compile(q.compilingCtx, queryStr, q.now, query.Verbose(c.verbose))
 	if err != nil {
 		return errors.Wrap(err, "failed to compile query")
 	}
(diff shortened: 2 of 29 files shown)


jlapacik:
The last commit e939a561fcf081d26fc8ff30c074f270b28f897a introduces an interface for retrieving variable/identifier bindings from a specific block in a Flux program. Also for adding new variable bindings to a specific block of the code after it has already been parsed to an AST and then a semantic program.

This interface allows you to add and retrieve variables from any block of a Flux program, but its intended use case is for getting and setting Flux options. The _options block_ is an implicit block of a Flux program, nested directly inside of the top-level _universe block_. All option variables are bound inside of the _options block_. Therefore you can call the methods `Get` and `Set` directly on this block in order to programmatically get and set options.

`Set` and `SetValues` should be called before interpretation of the program.
`Get` and `GetValues` should be called after interpretation of the program.

As follows:

```
globalScope := interpreter.NewScope()
optionScope := globalScope.Nest()
programScope := optionScope.Nest()

optionScope.Set("task", values.NewIntValue(5))

interpreter.Eval(semanticProgram, programScope, optionScope)

optionSpecifiedInScript := optionScope.Get("some_option")
```

query/spec.go
```diff
@@ -12,6 +13,7 @@ type Spec struct {
 	Edges      []Edge             `json:"edges"`
 	Resources  ResourceManagement `json:"resources"`
 
+	now      time.Time
```
nathanielc: We need to `now` value to be public so it can be set externally and as part of the marshalled JSON.

query/spec.go
```diff
@@ -23,6 +25,19 @@ type Edge struct {
 	Child  OperationID `json:"child"`
 }
 
+// SetNow idempotently sets the current system time for a query
+func (q *Spec) SetNow() {
```
nathanielc: We want to treat the Spec as read only once its been created. So the plan should not call this method, but rather just read the Now field of the Spec. And then if it is Zero change the value in the plan struct.

query/spec.go
```diff
@@ -23,6 +25,19 @@ type Edge struct {
...
+	if q.now.IsZero() {
+		q.now = time.Now().UTC()
+	}
+}
+
+// GetNow retrieves the current system time
+// that a query will execute with respect to.
+func (q *Spec) GetNow() time.Time {
+	return q.now
+}
```
nathanielc: Once Now is public there will be no need for a getter.

query/spec.go
```diff
@@ -73,6 +88,8 @@ func (q *Spec) Parents(id OperationID) []*Operation {
 // prepare populates the internal datastructure needed to quickly navigate the query DAG.
 // As a result the query DAG is validated.
 func (q *Spec) prepare() error {
+	q.SetNow()
```
nathanielc: We do not want to mutate the public parts of the spec in prepare.

query/plan/physical.go
```diff
@@ -14,16 +14,18 @@ import (
...
 	Order      []ProcedureID
-	// Results is a list of datasets that are the result of the plan
-	Results map[string]YieldSpec
 
+	// Results is a list of datasets that are the result of the plan
+	Results   map[string]YieldSpec
 	Resources query.ResourceManagement
+
+	// Now represents the relative current time of the plan.
+	Now time.Time
```
nathanielc: Why did these fields get reordered?

query/plan/physical.go
```diff
@@ -61,13 +63,18 @@ func NewPlanner() Planner {
...
+		return nil, errors.New("Now time is not set")
+	}
+
 	p.plan = &PlanSpec{
-		Now:        now,
 		Procedures: make(map[ProcedureID]*Procedure, len(lp.Procedures)),
 		Order:      make([]ProcedureID, 0, len(lp.Order)),
 		Resources:  lp.Resources,
 		Results:    make(map[string]YieldSpec),
+		Now:        now,
```
nathanielc: Why the change here?

query/plan/logical.go
```diff
@@ -49,6 +51,9 @@ func (p *logicalPlanner) Plan(q *query.Spec) (*LogicalPlanSpec, error) {
 	if err != nil {
 		return nil, err
 	}
+	// GetNow must be called after Walk to ensure
+	// the current system time has been set.
+	p.plan.Now = q.GetNow()
```
nathanielc: This is where the logic to check against Zero should live. 

````
p.plan.Now = q.Now
if p.plan.Now.IsZero() {
 p.plan.Now = time.Now().UTC()
}
```

query/control/controller.go
```diff
@@ -99,7 +99,6 @@ func (c *Controller) createQuery(ctx context.Context, orgID platform.ID) *Query
 		},
 		state:     Created,
 		c:         c,
-		now:       time.Now().UTC(),
```
nathanielc: We should keep this behavior. The controller places the incoming queries in a queue. We want to capture the time closest to when the query is submitted not when it finally leaves the queue.

So before the query is planned, check if the spec and set this `now` time if is zero.

In other words, store the time we received the query, then before planning, if the spec does not specify a Now time, use the time we received the query as the now time.

query/compile.go
```diff
@@ -150,6 +164,17 @@ func RegisterBuiltInValue(name string, v values.Value) {
 	builtinScope[name] = v
 }
 
+// RegisterBuiltInOption adds the value to the builtin scope.
+func RegisterBuiltInOption(name string, v values.Value) {
+	if finalized {
+		panic(errors.New("already finalized, cannot register builtin option"))
+	}
+	if _, ok := builtinScope[name]; ok {
```
nathanielc: This should check against builtinOptions

query/compile.go
```diff
@@ -101,6 +107,14 @@ func toSpec(stmtVals []values.Value) *Spec {
 type CreateOperationSpec func(args Arguments, a *Administration) (OperationSpec, error)
 
 var builtinScope = make(map[string]values.Value)
+
+// TODO(Josh): Default option values should be registered similarly to built-in
+// functions. Default options should be registered in their own files
+// (or in a single file) using the RegisterBuiltInOption function which will
+// place the resolved option value in the following map.
+//
+// Note: This is how one should register all options EXCEPT for the special "now" option.
```
nathanielc: Why can't the now options be specified here?

jlapacik: It might be able to actually. If it's being registered, that means the query has already been submitted and is executing. So assuming we could set the option value to `time.Now().UTC()` then we could register the now option just like any other option.

jlapacik:
Variable scoping is a concept relevant only to the interpreter. So how about we encapsulate the notion of scope completely inside the interpreter? We can then add a `GetOption` and `SetOption` interface to the interpreter itself instead of the scope which makes more sense logically and is also a little cleaner.

query/interpreter/interpreter.go
```diff
@@ -10,19 +10,73 @@ import (
...
+	return itrp.globals
+}
+
+// SetVar adds a variable binding to the global scope
+func (itrp *Interpreter) SetVar(name string, val values.Value) {
+	itrp.globals.Set(name, val)
+}
+
+// GetValues returns the evaluated expressions of a Flux program
+func (itrp *Interpreter) GetValues() []values.Value {
```
nathanielc: Getters in Go should not use the `Get` prefix.

query/compile.go
```diff
@@ -35,7 +35,7 @@ type options struct {
 }
 
 // Compile evaluates a Flux script producing a query Spec.
-func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
+func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spec, error) {
```
nathanielc: Add comment to that `now` must be non zero.

query/interpreter/interpreter.go
nathanielc: Rename `GetValues` to `SideEffects`.

query/compile.go
```diff
@@ -49,27 +49,46 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
...
 		return nil, err
 	}
 
-	operations, err := interpreter.Eval(semProg, interpScope)
-	if err != nil {
+	if err := itrp.Eval(semProg); err != nil {
 		return nil, err
 	}
-	spec := toSpec(operations)
+	spec := toSpec(itrp.GetValues())
```
nathanielc: `toSpec` should access the intepreter and grab the sideeffects and now option itself.

query/compile.go
```diff
@@ -369,20 +405,18 @@ func DefaultFunctionSignature() semantic.FunctionSignature {
 	}
 }
 
-func BuiltIns() (map[string]values.Value, semantic.DeclarationScope) {
+func BuiltIns() (*interpreter.Interpreter, semantic.DeclarationScope) {
```
nathanielc: Why was this API changed? The uses of BuiltIns all use the scope anyways and no other interperter methods.

query/control/controller.go
```diff
@@ -110,10 +110,14 @@ func (c *Controller) compileQuery(q *Query, queryStr string) error {
...
 	}
-	spec, err := query.Compile(q.compilingCtx, queryStr, query.Verbose(c.verbose))
+	spec, err := query.Compile(q.compilingCtx, queryStr, q.now, query.Verbose(c.verbose))
 	if err != nil {
 		return errors.Wrap(err, "failed to compile query")
 	}
+	// Query has already been created so we are guaranteed to have a now time
+	if spec.Now.IsZero() {
+		spec.Now = q.now
+	}
```
nathanielc: This is no longer necessary, we passed in `q.now` so spec.Now is already set.

query/functions/join.go
```diff
@@ -127,7 +127,7 @@ func createJoinOpSpec(args query.Arguments, a *query.Administration) (query.Oper
 				return
 			}
 			p := t.(*query.TableObject)
-			joinParams.add(k/*parameter*/, p/*argument*/)
+			joinParams.add(k /*parameter*/, p /*argument*/)
```
nathanielc: Remove comments here.

query/interpreter/interpreter.go
```diff
@@ -10,19 +10,73 @@ import (
...
+
+// GetOption returns a Flux option by name
+func (itrp *Interpreter) GetOption(name string) values.Value {
+	return itrp.options.Get(name)
+}
+
+// GetOptions returns all Flux options
+func (itrp *Interpreter) GetOptions() map[string]values.Value {
+	return itrp.options.GetValues()
+}
```
nathanielc: Remove unused plural methods.

query/compile.go
```diff
@@ -49,38 +50,70 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
...
+	}
+
+	for k, v := range builtinOptions {
+		options[k] = v
+	}
+
+	return interpreter.NewInterpreter(options, globals)
+}
+
+func nowFunc(now time.Time) values.Value {
```
nathanielc: This should return `values.Function` i.e the more specific interface.

query/compile.go
```diff
@@ -49,38 +50,70 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
...
+	name := "now"
+	ftype := semantic.NewFunctionType(semantic.FunctionSignature{
+		ReturnType: semantic.Time,
+	})
+	call := func(args values.Object) (values.Value, error) {
+		return values.NewTimeValue(values.ConvertTime(now)), nil
```
nathanielc: Let's do this conversion only once and then simply return the converted value from this function. 

```
n := values.NewTimeValue(values.ConvertTime(now))
...
return n, nil
```
Or similar

query/compile.go
```diff
@@ -95,12 +128,23 @@ func toSpec(stmtVals []values.Value) *Spec {
 			}
 		}
 	}
+
+	// now option is Time value
+	nowValue, _ := itrp.Option("now").Function().Call(values.NewObject())
```
nathanielc: Can you pass `nil` as the args object?

query/compile.go
```diff
@@ -49,38 +50,70 @@ func Compile(ctx context.Context, q string, opts ...Option) (*Spec, error) {
...
+
+	for k, v := range builtinOptions {
+		options[k] = v
+	}
+
+	return interpreter.NewInterpreter(options, globals)
+}
+
+func nowFunc(now time.Time) values.Value {
+	name := "now"
```
nathanielc: Lets make `now` a constant string so we can refer to it here and below were we read the option. Maybe call it `nowOption`?

query/interpreter/interpreter_test.go
```diff
@@ -2,27 +2,41 @@ package interpreter_test
...
+	_ "github.com/influxdata/platform/query/options"
 	"github.com/influxdata/platform/query/parser"
 	"github.com/influxdata/platform/query/semantic"
 	"github.com/influxdata/platform/query/semantic/semantictest"
 	"github.com/influxdata/platform/query/values"
 )
 
-var testScope = interpreter.NewScope()
+func init() {
+	query.FinalizeBuiltIns()
```
nathanielc: This test should not include any `query` package specific logic.

query/interpreter/interpreter_test.go
```diff
@@ -418,6 +439,67 @@ func TestResolver(t *testing.T) {
 	}
 }
 
+// ExampleInterpreter_Option demonstrates retrieving an option from the Flux interpreter
+func ExampleInterpreter_Option() {
+	itrp := query.NewInterpreter()
```
nathanielc: These examples look good. But they should exist in the `query` package since they are specific to the way the query interpeter is used.

query/interpreter/interpreter_test.go
```diff
@@ -418,6 +439,67 @@ func TestResolver(t *testing.T) {
...
+	// The now option is a function value whose default behavior is to return
+	// the current system time when called.
+	nowTime, _ := nowFunc.Function().Call(values.NewObject())
+	fmt.Fprintf(os.Stderr, "The current system time (UTC) is: %v\n", nowTime)
+	// Output:
+}
+
+// ExampleInterpreter_SetOption demonstrates setting an option from the Flux interpreter
+func ExampleInterpreter_SetOption() {
+	itrp := query.NewInterpreter()
```
nathanielc: This example should probably just use the `inteperter.Interpreter` type, not the `query` specific one.

query/interpreter/interpreter_test.go
```diff
@@ -418,6 +439,67 @@ func TestResolver(t *testing.T) {
...
+	// Set a new option from the interpreter
+	itrp.SetOption("dummy_option", values.NewIntValue(3))
+
+	fmt.Printf("dummy_option = %d", itrp.Option("dummy_option").Int())
+	// Output: dummy_option = 3
+}
+
+// Example_overrideDefaultOption demonstrates a query using an option
+// that is overridden before the query it evaluated by the interpreter.
+func ExampleInterpreter_overrideDefaultOption() {
```
nathanielc: Rename to `ExampleInterpreter_OverrideDefaultOption`

query/values/function.go
```diff
@@ -0,0 +1,89 @@
+package values
+
+import (
+	"regexp"
+
+	"github.com/influxdata/platform/query/semantic"
+)
+
+// NewFunction returns a new function value
+func NewFunction(name string, typ semantic.Type, call func(args Object) (Value, error), sideEffect bool) Function {
```
nathanielc: Can you move the `Function` interface to this file as well?

query/values/function.go
```diff
@@ -0,0 +1,89 @@
...
+func (f function) Array() Array {
+	panic(UnexpectedKind(semantic.Object, semantic.Function))
+}
+
+func (f function) Object() Object {
+	panic(UnexpectedKind(semantic.Object, semantic.Object))
+}
+
+func (f function) Function() Function {
+	return &f
```
nathanielc: This will create a copy of the function struct each time, which doesn't change the behavior since the struct contains a pointer to a Go function. As a result it is wasteful. Since the implementation requires that a pointer receiver be used on at least one method, a pointer receiver should be used for all methods.

Can you update all method to use a pointer receiver?

query/interpreter/interpreter_test.go
jlapacik: Because the Interpreter does not have a method called OverrideDefaultOption, the Go testing example framework says to keep lowercase. The example will actually throw an error otherwise.

Issue #309, opened 2018-07-03
IFQL Meta queries should return normal tables, allowing for further manipulation


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#209  

IFQL Meta queries return data about the schema of a table. The returned result should be a table itself so it can be further transformed.

See for  some use cases https://github.com/influxdata/influxdb/issues/5548 https://github.com/influxdata/influxdb/issues/5668

PR #390, opened 2018-07-12
Add initial flux meta query tests

adamperlin:
Addresses #309; Adds test cases for all major meta-queries outlined [here](https://gist.githubusercontent.com/stuartcarnie/8a1e0f937e78f13fe195c355986642d2/raw/d37d8667e30746ad1272bb1f836dd09c69989c32/meta-queries.md). Are these results as expected?


query/functions/testdata/meta_query_measurements.out.csv
```diff
@@ -0,0 +1,6 @@
+#datatype,string,long,string,string
+#partition,false,false,false,false
+#default,_result,,,
+,result,table,_measurement,_value
+,,0,disk,disk
```
nathanielc: Can you add input data so we get more than one row of output?

adamperlin: As in add another measurement?

adamperlin: I added additional measurements to the input data

stuartcarnie:
I compared the test coverage to what I had posted in [this gist](https://gist.github.com/stuartcarnie/0e94582dc740ff3a032306d1829c0773) and it looks good 👍

adamperlin:
Great thanks for reviewing!

aanthony1243:
This is OK to merge.

Issue #393, opened 2018-07-12
Flux `difference` function panics if called on empty table


adamperlin:
If `difference` is called on an empty table, it causes a runtime panic. This can easily happen if no data is returned from an earlier part of the query. For example: 
```flux
from(db:"db")
    |> filter(fn: (r) => r._field == "no_exist")
    |> difference()
```
Panic stack trace:
```
panic: runtime error: slice bounds out of range
goroutine 12 [running]:
runtime/debug.Stack(0x14ba440, 0x14b3b00, 0x1868f80)
	/usr/local/Cellar/go/1.10.3/libexec/src/runtime/debug/stack.go:24 +0xa7
github.com/influxdata/platform/query/execute.(*poolDispatcher).Start.func1.1(0xc42034f8b0)
	/Users/User/go/src/github.com/influxdata/platform/query/execute/dispatcher.go:67 +0xb9
panic(0x14b3b00, 0x1868f80)
	/usr/local/Cellar/go/1.10.3/libexec/src/runtime/panic.go:502 +0x229
github.com/influxdata/platform/query/functions.(*differenceTransformation).Process.func1(0x15beac0, 0xc420426d70, 0x1, 0xc420426ff0)
	/Users/User/go/src/github.com/influxdata/platform/query/functions/difference.go:215 +0x859
github.com/influxdata/platform/query/execute.(*ColListBlock).Do(0xc420426d70, 0xc420426ff0, 0x4, 0x5)
	/Users/User/go/src/github.com/influxdata/platform/query/execute/block.go:602 +0x3c
github.com/influxdata/platform/query/functions.(*differenceTransformation).Process(0xc42035bc00, 0x8758767ca5bfbed5, 0x63eb5c4900b059b8, 0x15bb240, 0xc420426d70, 0x1, 0x0)
	/Users/User/go/src/github.com/influxdata/platform/query/functions/difference.go:172 +0x44f
github.com/influxdata/platform/query/execute.processMessage(0x15bb840, 0xc42035bc00, 0x15b5e00, 0xc420524060, 0x0, 0x0, 0x0)
	/Users/User/go/src/github.com/influxdata/platform/query/execute/transport.go:199 +0x200
github.com/influxdata/platform/query/execute.(*consecutiveTransport).processMessages(0xc4202e3b00, 0xa)
	/Users/User/go/src/github.com/influxdata/platform/query/execute/transport.go:156 +0xa1
[...]

PR #398, opened 2018-07-12
Fix for panic when `difference` is called with no data

adamperlin:
Address #393; Small bounds checking fix in `difference` implementation to prevent panic; added flux queries to demonstrate relevant behavior; updates to `difference` unit tests for cases when there are less than two rows.


query/functions/difference.go
```diff
@@ -173,6 +173,10 @@ func (t *differenceTransformation) Process(id execute.DatasetID, b query.Block)
 		l := cr.Len()
 		for j, c := range cols {
 			d := differences[j]
+			if d == nil && l == 0 {
```
aanthony1243: should this be || ? 

or another approach,  if   `l == 0` we can just skip the loop completely, right?

adamperlin: whoops you're right - `l == 0` is all that's needed.

adamperlin: fixed

query/functions/difference.go
```diff
@@ -173,7 +173,7 @@ func (t *differenceTransformation) Process(id execute.DatasetID, b query.Block)
 		l := cr.Len()
 		for j, c := range cols {
 			d := differences[j]
-			if d == nil && l == 0 {
+			if l == 0 {
```
aanthony1243: `l` is unchanged by this loop, so you're looping over cols for no reason.  
you probably want 
```
if l != 0 {
    for j, c := range cols {
     .....
   }
}
```

Issue #307, opened 2018-07-03
What to do about COUNT with missing data?


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#211  

https://github.com/influxdata/influxdb/issues/6412
https://github.com/influxdata/influxdb/issues/6967

Do we create series for all tag in the index? How could we do that in a deterministic manner.

PR #400, opened 2018-07-12
Add flux tests demonstrating counts with no data on some time intervals

adamperlin:
Addresses #307; Adds a test for `count` with data gaps, in addition to a general `count` test. In the included test for data gaps (`count_no_data`), there is no data for the last two windows that the query is requesting a `count` for. The current behavior of `count` is to ignore those windows and only provide counts for windows that contain data. If this   test's output  was in accordance with [this](https://github.com/influxdata/influxdb/issues/6412) feature request (also see https://github.com/influxdata/influxdb/issues/6967), one might expect two more windows in the output, both with `count` 0.  @nathanielc is this behavior desired?

### DIFF
--- a/query/functions/testdata/count_no_data.in.csv
+++ b/query/functions/testdata/count_no_data.in.csv
@@ -1,0 +1,9 @@
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string,string
+#partition,false,false,false,false,false,false,true,true,true,true
+#default,_result,,,,,,,,,
+,result,table,_start,_stop,_time,_value,_field,_measurement,cpu,host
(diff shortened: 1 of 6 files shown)


query/functions/testdata/count_no_data.in.csv
```diff
@@ -0,0 +1,9 @@
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string,string
```
aanthony1243: I don't see where there would be empty windows in this data, given the query.

nathanielc:
Yes, my work on window just specified how this should work. I have opened this issue https://github.com/influxdata/platform/issues/661 to track updating the implementation.

That issue depends on three other issues #659 #658 #657

PR #401, opened 2018-07-12
feat: Use DBRPMappings in 1.x read path

nathanielc:


query/functions/query_test.go
```diff
@@ -29,7 +53,9 @@ var skipTests = map[string]string{
 func Test_QueryEndToEnd(t *testing.T) {
 	qs := querytest.GetQueryServiceBridge()
 
-	influxqlTranspiler := influxql.NewTranspiler()
+	influxqlTranspiler := influxql.NewTranspilerWithConfig(influxql.Config{
```
jsternberg: Can you make the DBRPMapper a required parameter instead of part of the config? I don't think there's a way to run the transpiler without it. Most likely, when I implement the Transpiler, I'm going to make the Schema interface include the DBRPMapper like this:

```
type Schema interface {
    DBRPMapper
}
```

query/influxql/transpiler.go
```diff
@@ -128,13 +130,27 @@ func (t *transpilerState) from(m *influxql.Measurement) (query.OperationID, erro
 	if rp == "" {
 		if t.config.DefaultRetentionPolicy != "" {
 			rp = t.config.DefaultRetentionPolicy
-		} else {
-			rp = "autogen"
 		}
 	}
+	log.Println("dbrp", db, rp)
```
jsternberg: Debugging code that should be removed.

Issue #250, opened 2018-07-03
Parse regular expression literals


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#314  

From the spec

A regular expression literal represents a regular expression pattern, enclosed in forward slashes.
Within the forward slashes, any unicode character may appear except for an unescaped forward slash.
The `\x` hex byte value representation from string literals may also be present.

Regular expression literals support only the following escape sequences:

    \/   U+002f forward slash
    \\   U+005c backslash

    regexp_lit         = "/" { unicode_char | byte_value | regexp_escape_char } "/" .
    regexp_escape_char = `\` (`/` | `\`)

Examples:

    /.*/
    /http:\/\/localhost:9999/
    /^\xe6\x97\xa5\xe6\x9c\xac\xe8\xaa\x9e(ZZ)?$/
    /^日本語(ZZ)?$/ // the above two lines are equivalent
    /\\xZZ/ // this becomes the literal pattern "\xZZ"

The regular expression syntax is defined by [RE2](https://github.com/google/re2/wiki/Syntax).

aanthony1243:
@adamperlin  we have some regex tests.  See if you can put a regex in a variable and use it.

nathanielc:
Regex literals can be parsed. If there is a bug with it we can open a separate issue.

PR #402, opened 2018-07-12
Add ifql test for parsing and storing regexp in variable

adamperlin:
Addresses #250; Adds additional test coverage for regular expressions, including storing a regexp in a variable for use in a filter predicate.


adamperlin:
The second test does not pass - no data is returned. See #408

adamperlin:
@aanthony1243 sounds good.

aanthony1243:
@adamperlin let's rebase with master and merge this back.  you document the problem in #408 and there's a test here too.  Make sure you regenerate the output for all the tests, it may be different now with so many improvements.

PR #403, opened 2018-07-12
feat(query/control): catch panics when executing a query in the controller

jsternberg:


query/control/controller.go
```diff
@@ -198,6 +206,26 @@ func (c *Controller) run() {
 
 // processQuery move the query through the state machine and returns and errors and if the query should be popped.
 func (c *Controller) processQuery(q *Query) (pop bool, err error) {
+	defer func() {
+		if e := recover(); e != nil {
+			// Log that a panic happened.
+			c.logger.Error("Caught panic", zap.Any("error", e))
```
nathanielc: We do not need to log the panic. We are returning the error to the caller which will return the error on up to the query logger. We do not need to handle the error twice.

As a result the controller should not need a logger.

jsternberg: Removed.

PR #406, opened 2018-07-12
chore: Rename block to table

nathanielc:


query/execute/table.go
```diff
@@ -114,30 +114,30 @@ func AddNewCols(b query.Block, builder BlockBuilder) []int {
 	return colMap
 }
 
-// AppendBlock append data from block b onto builder.
-// The colMap is a map of builder column index to block column index.
-func AppendBlock(b query.Block, builder BlockBuilder, colMap []int) {
-	if len(b.Cols()) == 0 {
+// AppendTable append data from table b onto builder.
```
mark-rushakoff: Should be `table t`.

nathanielc: Wow, good catch. Thanks!

PR #411, opened 2018-07-13
docs(http): add swagger documentation for auth via extension

jaredscheib:
A developer developing against our API specification is not currently presented with information about how auth works in conjunction with the API. This PR proposes a step in a direction that could be taken.

We wanted to solicit feedback before continuing down this path, and/or at least put the idea out there for consideration.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -352,6 +352,8 @@ paths:
       tags:
         - Buckets
       summary: List all buckets
+      security:
+        - apiKeyAuth: [read-bucket]
       responses:
         '200':
           description: a list of buckets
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -955,6 +957,29 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 components:
+  securitySchemes:
+    apiKeyAuth:
+      type: apiKey
+      in: header
```
goller: This is not what we are doing right now.  We are using type `http` with a scheme.
Here are the docs about swagger for auth: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#security-scheme-object

http/swagger.yml
```diff
@@ -955,6 +957,29 @@ paths:
...
+        write-bucket:
+          action: write
+          resource: bucket
+        create-bucket:
+          action: create
+          resource: bucket
+        delete-bucket:
+          action: delete
+          resource: bucket
+    cookieAuth:
```
goller: Let's not document anything about cookies at the moment.

jaredscheib:
Next steps:
- [ ] Fill out permissions across all routes based on latest commit
- [ ] Create follow-on issue from authorization PR to change from non-standard Authorization Token to standard Authorization Bearer

PR #418, opened 2018-07-16
feat: Add query statistics

nathanielc:
Introduces the Statisticser interface which ResultIterators may
implement.

The HTTP implementation uses HTTP trailers to preserve the statistics.
This way we do not need to have all encoders and decoders support
statistics.

### DIFF
--- a/http/query_service.go
+++ b/http/query_service.go
@@ -17,6 +17,8 @@ import (
 
 const (
 	queryPath = "/v1/query"
+
+	statsTrailer = "Influx-Query-Statistics"
 )
 
 type QueryHandler struct {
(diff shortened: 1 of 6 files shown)


http/query_service.go
```diff
@@ -208,5 +228,57 @@ func (s *QueryService) processResponse(resp *http.Response) (query.ResultIterato
...
+
+func (s *statsResultIterator) Cancel() {
+	s.result.Cancel()
+	s.readStats()
+}
+
+func (s *statsResultIterator) Err() error {
+	err := s.result.Err()
+	if err != nil {
+		return err
```
aanthony1243: are there any stats we'd want to see in an error case?  I'm thinking at least memory usage, but durations may be useful too if the query ran for a while and then failed.

nathanielc: The stats are still reported in the error case. Returning early here does not prevent the reading of the stats.

PR #420, opened 2018-07-17
feat(docker): create a fluxd nightly docker image

### LINKED ISSUE #167: Publish a nightly build to Quay for Flux
Having a nightly tagged as nightly  -- like Chronograf see below -- for Flux will allow us to integrate this into a "nightly" sandbox.  Essentially, we want to provide an option for folks to use either a stable build or a nightly build of our bits through a standard set of docker containers.

screen shot 2018-06-22 at 9 06 41 am

jsternberg:
Fixes #167.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,9 @@ jobs:
     working_directory: /go/src/github.com/influxdata/platform
     steps:
       - checkout
-      - run: make nightly
+      - run: |
+          docker login -u "$QUAY_USER" -p $QUAY_PASS quay.io
+          make nightly
 
 
 workflows:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,10 @@ jobs:
     working_directory: /go/src/github.com/influxdata/platform
     steps:
       - checkout
-      - run: make nightly
+      - run: |
+          make nightly
+          docker login -u "$QUAY_USER" -p $QUAY_PASS quay.io
+          docker push quay.io/influxdb/flux:nightly
```
nathanielc: Can you add these docker command to the make file. Then we can call the makefile locally if needed.

jsternberg: One potential issue with that is the calls to `docker login` are CI specific. I can do this instead:

```
docker login -u "$QUAY_USER" -p $QUAY_PASS quay.io
make nightly
```

That way the push happens inside of the Makefile.

nathanielc: Yeah that works. Thanks

PR #421, opened 2018-07-17
fix: Defer writing HTTP headers in query service

nathanielc:
This makes it so we can encode errors in HTTP headers so long as the
encoder did not write anything.

This also changes the behavior to encode the stats trailer even in the
event of an error.


http/query_service.go
```diff
@@ -115,22 +120,20 @@ func (h *QueryHandler) handlePostQuery(w http.ResponseWriter, r *http.Request) {
...
 		}
 	}
 
 	if hasStats {
 		data, err := json.Marshal(stats.Statistics())
 		if err != nil {
-			h.Logger.Info("Failed to encode statisitcs",
-				zap.Error(err),
-			)
+			h.Logger.Info("Failed to encode statisitcs", zap.Error(err))
```
goller: statistics

PR #423, opened 2018-07-17
feat(query): split compile into two functions to allow Eval access

lyondhill:


query/compile.go
```diff
@@ -42,35 +42,44 @@ func Compile(ctx context.Context, q string, now time.Time, opts ...Option) (*Spe
...
-	itrp.SetOption(nowOption, nowFunc(now))
+	spec := toSpec(itrp)
+
+	if o.verbose {
+		log.Println("Query Spec: ", Formatted(spec, FmtJSON))
+	}
+	return spec, nil
+}
+
+func Eval(itrp *interpreter.Interpreter, q string) error {
```
nathanielc: Can you add a comment here? And remove the first empty line in the function body?

lyondhill: Will do.

PR #424, opened 2018-07-17
fix(circleci): setup remote docker so that the deploy works for nightlies

jsternberg:


jsternberg:
I need to amend the commit to get rid of the successful deploy, but here is the commit that succeeded: https://circleci.com/gh/influxdata/platform/731

PR #425, opened 2018-07-18
feature(kit/signals): add standard method to capture signals with context

goller:
I'm adding various nice-to-haves in the kit package.

Here is a helper function to capture signals and receive notifications via `<-ctx.Done()`


leodido:
LGTM

PR #428, opened 2018-07-18
chore(kit/signals): removed duplicated check for signals and add an example

fntlnz:

### DIFF
--- a/kit/signals/context_test.go
+++ b/kit/signals/context_test.go
@@ -25,6 +25,25 @@ func ExampleWithSignals() {
 	// finished
 }
 
+func ExampleWithUnregisteredSignals() {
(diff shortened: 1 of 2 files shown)


kit/signals/context_test.go
```diff
@@ -25,6 +25,24 @@ func ExampleWithSignals() {
...
+		time.Sleep(10 * time.Millisecond) // after some time SIGUSR2 is sent
+		// mimicking a signal from the outside, WithSignals will not handle it
+		syscall.Kill(syscall.Getpid(), syscall.SIGUSR2)
+	}()
+
+	select {
+	case <-ctx.Done():
+		fmt.Println("finished")
+	}
+
```
goller: Would you add an example output here?
https://golang.org/pkg/testing/#hdr-Examples

fntlnz:
@goller did that!

PR #438, opened 2018-07-19
WIP Filtering in flux range function when not pushed down

adamperlin:
Adds basic range filtering to flux `range()`, to allow the function to return results when its operations are not delegated to the storage layer (such as when a different source is used, i.e. `fromCSV`). When used with the storage layer, `range` should behave as it did previously. 

Currently, this is a WIP because there are some details to be clarified, namely:
- Is `range()` valid if it is called with two negative, **absolute** start and stop times? What about if the absolute start time is greater than the absolute stop time?
- Does `range()` need to handle `_start` and `_stop` in any way?
- Should `range()` order results by time before returning them? What is the storage layer's protocol for this? 
- Currently, the implementation checks each row against the provided time range. This seems inefficient, but the data tables created from input such as a csv file are not guaranteed to be pre-sorted by time. Is there a better approach?

Other notes: 
All flux and influxql transpiler tests have been updated to work with the new behavior of `range()`.
I rebased over the changes introduced in https://github.com/influxdata/platform/pull/430 because it fixed a deadlock that made testing much more inconvenient. I can remove if necessary.


adamperlin:
Also - ci is failing because of pre-existing vet errors

pauldix:
For negatives, do you mean something like `range(start:-2h, end:-1h)`? If so, that's valid. The start time must not be greater than the end time, that should throw an error on validation.

I don't think it needs to handle `_start` and `_stop` in any way. Although it should take a `col` argument that is default to `_time` similar to how the aggregate functions have that argument that defaults to `_value`.

I'm not sure about the sorting, it's very inefficient to check every record. But it would be more inefficient to sort. We could do something like assume that the input to range is sorted and put that on the user to ensure they sort before the call to range. Could also make `sort` an optional parameter that defaults to false. If we have that then I'd also have `order` be an optional parameter that defaults to `asc`.

adamperlin:
Thanks for the comment! For the negative **absolute** time bounds, I was talking about the rare case where a relative offset might overflow the value of `now()` in the negative direction (I ran into this while testing). If `now` was set to epoch+1h and we call `range(start:-2h)`, we end up with `range(start:1h-2h)`, or `range(start:unix(-3600))`. Should that be supported?

adamperlin:
Added support for `column` parameter (if `range` is not pushed down). Should we add support for this to the storage layer as well?

nathanielc:
>Is range() valid if it is called with two negative, absolute start and stop times? What about if the absolute start time is greater than the absolute stop time?

> If now was set to epoch+1h and we call range(start:-2h), we end up with range(start:1h-2h), or range(start:unix(-3600)). Should that be supported?

Yes this is supported, the times are stored as nanoseconds since the epoch as an int64. So that gives us a couple hundred years before and after the epoch.

> Does range() need to handle _start and _stop in any way?

Yes, it should clip the _start and _stop values if they exist and set them if they do not. The names `_start` and `_stop` should be configurable using the `startCol` and `stopCol` arguments to the function. Look at how window does this.
Also range can both use the columns to be more efficient. If it sees that the _start and _stop columns exist already and are part of the group key and within the range, it can pass the entire table on without any needing to read row by row.

>Should range() order results by time before returning them? What is the storage layer's protocol for this?

No, range should not change the order in any way.

> Currently, the implementation checks each row against the provided time range. This seems inefficient, but the data tables created from input such as a csv file are not guaranteed to be pre-sorted by time. Is there a better approach?

This is fine, the only efficiency gain I can see is to use the existing `_stop` and `_start` cols if they exist and are part of the group key. 


> Added support for column parameter (if range is not pushed down). Should we add support for this to the storage layer as well?

No, if the column specified is not the `_time` column then range cannot be pushed down. You will need to update the PushDownRule for range to ensure that.

query/functions/range.go
```diff
@@ -2,25 +2,31 @@ package functions
...
 )
 
 const RangeKind = "range"
 
 type RangeOpSpec struct {
-	Start query.Time `json:"start"`
-	Stop  query.Time `json:"stop"`
+	Start         query.Time `json:"start"`
+	Stop          query.Time `json:"stop"`
+	Column        string     `json:"column"`
```
aanthony1243: as discussed, let's follow the pattern for Window here, 
`TimeCol            string            `json:"time_col"``

aanthony1243: You might try a trial version of Goland for this task.  It has a refactor tool that can find many of the references you've made in all these 60 files.  

Also, it's a 1 month trial, so it would nearly last the rest of your internship if you wanted to give it a shot.

query/functions/range.go
```diff
@@ -121,27 +168,36 @@ func createRangeTransformation(id execute.DatasetID, mode execute.AccumulationMo
...
-	cache execute.TableBuilderCache
-
-	Start query.Time
-	Stop  query.Time
+	d             execute.Dataset
+	cache         execute.TableBuilderCache
+	pushedDown    bool
+	Start         query.Time
+	Stop          query.Time
+	Adjusted      execute.Bounds
```
aanthony1243: is adjusted the right idea here?  These are the Previous bounds, before this transformation, right?  Consider renaming.

query/functions/range.go
```diff
@@ -121,27 +168,36 @@ func createRangeTransformation(id execute.DatasetID, mode execute.AccumulationMo
...
+	Column        string
```
aanthony1243: rename this when you rename the one in the spec above

query/functions/range.go
```diff
@@ -150,18 +206,137 @@ func (t *rangeTransformation) RetractTable(id execute.DatasetID, key query.Group
...
+
+	forwardTable := false
+	if startKeyColIdx > 0 && stopKeyColIdx > 0 {
+		keyStart := tbl.Key().Value(startKeyColIdx).Time()
+		keyStop := tbl.Key().Value(stopKeyColIdx).Time()
+		keyBounds := execute.Bounds{
+			Start: keyStart,
+			Stop:  keyStop,
+		}
+		forwardTable = t.Adjusted.Overlaps(keyBounds)
```
aanthony1243: This is not an Overlap case, but a Contains case.  You might also need an Overlap test where the ranges have some part in common but one hangs over the other on one side.

query/functions/range.go
```diff
@@ -150,18 +206,137 @@ func (t *rangeTransformation) RetractTable(id execute.DatasetID, key query.Group
...
+					builder.AppendTime(j, start)
+				case t.stopColLabel:
+					var stop = values.Time(0)
+					if !stopAdded {
+						stop = cr.Times(j)[i]
+					}
+					if stop > t.Adjusted.Stop {
+						stop = t.Adjusted.Stop
+					}
+					builder.AppendTime(j, stop)
```
aanthony1243: this logic all looks good.  nice job.

query/functions/range_test.go
```diff
@@ -1,18 +1,25 @@
...
 	"github.com/influxdata/platform/query/execute"
+	"github.com/influxdata/platform/query/execute/executetest"
 	"github.com/influxdata/platform/query/functions"
 	"github.com/influxdata/platform/query/plan"
 	"github.com/influxdata/platform/query/plan/plantest"
 	"github.com/influxdata/platform/query/querytest"
 )
 
 func TestRange_NewQuery(t *testing.T) {
+	const csvFile = "testdata/range.in.csv"
```
aanthony1243: fromCSV has an optional parameter where you can pass in CSV text directly as a string literal.  That would make this test less fragile against moving files around.  take a look at the fromCSV test code.  

oh, but do you actually run this query?  or is this just testing the spec generator?  if that's the case make this csvFile point to something not real.

query/functions/range.go
adamperlin: Thanks for catching that. The overlap test is necessary also because if there is no overlap at all, no further processing is needed.

query/functions/range.go
```diff
@@ -150,18 +209,152 @@ func (t *rangeTransformation) RetractTable(id execute.DatasetID, key query.Group
 }
 
 func (t *rangeTransformation) Process(id execute.DatasetID, tbl query.Table) error {
+	if t.Bounds.Start > t.Bounds.Stop {
```
nathanielc: Where is `t.Bounds` set? I think that might be some dead code.

query/execute/bounds_test.go
```diff
@@ -0,0 +1,131 @@
...
+					Start: execute.Time(10),
+					Stop:  execute.Time(20),
+				},
+			},
+			want: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := execute.Bounds{
```
nathanielc: To make this test easier to follow can you have the test case contain two bounds structs? Instead of using the `fields` struct and creating the bounds each time.

Something like:

```
tests := []struct {
		name   string
		a, b execute.Bounds
		want   bool
	}{ ... }
```

query/execute/executetest/transformation.go
```diff
@@ -27,6 +28,15 @@ func ProcessTestHelper(
 	parentID := RandomDatasetID()
 	for _, b := range data {
 		if err := tx.Process(parentID, b); err != nil {
+			if wantErr != nil && !cmp.Equal(wantErr.Error(), err.Error()) {
```
nathanielc: The `Error` method returns a string, so there is no need to use `cmp.Equal` here. Can you simplify this logic?

query/execute/executetest/transformation.go
```diff
@@ -27,6 +28,15 @@ func ProcessTestHelper(
...
+				t.Errorf("unexpected error -want/+got\n%s", cmp.Diff(err.Error(), wantErr.Error()))
```
nathanielc: Can you change these pairs of `Errorf` and `Fatal` to just one call to `Fatalf`?

query/functions/range.go
```diff
@@ -7,20 +7,25 @@ import (
...
 
 const RangeKind = "range"
 
 type RangeOpSpec struct {
-	Start query.Time `json:"start"`
-	Stop  query.Time `json:"stop"`
+	Start         query.Time `json:"start"`
+	Stop          query.Time `json:"stop"`
+	TimeCol       string     `json:"time_col"`
+	StopColLabel  string     `json:"stop_col_label"`
```
nathanielc: The SPEC defines these as being simply `StopCol` instead of `StopColLabel`. Can you remove all the `Label` bits? 

i.e. `StopColLabel` -> `StopCol`, `stopColLable` -> `stopCol`, and `stop_col_label` -> `stop_col`.

And for `start` as well. 

The window function is being updated like this in a separate PR.

query/functions/range.go
```diff
@@ -121,27 +165,45 @@ func createRangeTransformation(id execute.DatasetID, mode execute.AccumulationMo
...
-	cache execute.TableBuilderCache
-
-	Start query.Time
-	Stop  query.Time
+	d             execute.Dataset
+	cache         execute.TableBuilderCache
+	Bounds        execute.Bounds
+	TimeCol       string
+	StartColLabel string
+	StopColLabel  string
```
nathanielc: Can you make these fields private? There is no need to expose them.

query/functions/range.go
```diff
@@ -150,18 +212,148 @@ func (t *rangeTransformation) RetractTable(id execute.DatasetID, key query.Group
...
+
+		keyStart := tbl.Key().Value(startKeyColIdx).Time()
+		keyStop := tbl.Key().Value(stopKeyColIdx).Time()
+		keyBounds := execute.Bounds{
+			Start: keyStart,
+			Stop:  keyStop,
+		}
+		// If there is no overlap between the bounds in the group key and the bounds in the range transformation,
+		// no further processing is needed.
+
```
nathanielc: nit, can you remove the space here so the comment is relevant for the `if` statement?

query/functions/range.go
```diff
@@ -150,18 +212,148 @@ func (t *rangeTransformation) RetractTable(id execute.DatasetID, key query.Group
...
+		// If [start, stop) (where start <= stop) from the group key is contained in the
+		// range transformation bounds [keyStart, keyStop], we can skip the whole table.
+
+		// Still want to skip if start >= keyStart and t.Bounds.Stop == keyStop
+		forwardTable = t.Bounds.Contains(keyStart) && (t.Bounds.Contains(keyStop) || t.Bounds.Stop == keyStop)
+	}
+
+	// Pass data through if the range node has been pushed down to the storage layer,
+	// or if the group key's start and stop were within the range bounds
```
nathanielc: This comment can now be removed.

adamperlin:
Changes made

query/functions/range.go
```diff
@@ -50,6 +55,30 @@ func createRangeOpSpec(args query.Arguments, a *query.Administration) (query.Ope
...
 
+	if col, ok, err := args.GetString("timeCol"); err != nil {
+		return nil, err
+	} else if ok {
+		spec.TimeCol = col
+	} else {
+		spec.TimeCol = execute.DefaultTimeColLabel
+	}
+
+	if label, ok, err := args.GetString("startColLabel"); err != nil {
```
nathanielc: These arguments names still need to be renamed.

query/functions/range_test.go
```diff
@@ -99,6 +148,308 @@ func TestRange_PushDown(t *testing.T) {
...
+				if err != nil {
+					t.Fatal(err)
+				}
+				t2, err := values.NewValue(values.Time(20*time.Minute.Nanoseconds()), semantic.Time)
+				if err != nil {
+					t.Fatal(err)
+				}
+
+				vs := []values.Value{t1, t2}
+				log.Println("values:", vs)
```
nathanielc: Can you remove this debug line?

adamperlin:
Fixed.

adamperlin:
Good? Can also remove the rest of the commit message.

PR #444, opened 2018-07-20
feat(query/functions) Update percentile function to provide a selector option add transpiler support.

aanthony1243:
Updated transpiler function support to call percentile.   Also updated percentile in flux to provide a selector option.


query/functions/percentile.go
```diff
@@ -272,3 +300,70 @@ func (a *ExactPercentileAgg) ValueFloat() float64 {
...
+func (s *ExactPercentileSelector) NewFloatSelector() execute.DoFloatRowSelector {
+	return &ExactPercentileSelector{Quantile: s.Quantile}
+}
+
+func (s *ExactPercentileSelector) NewStringSelector() execute.DoStringRowSelector {
+	return nil
+}
+
+func (s *ExactPercentileSelector) Rows() []execute.Row {
+	// TODO: sort data, then find the quantile points then return
```
nathanielc: This TODO looks complete

query/functions/percentile.go
```diff
@@ -272,3 +300,70 @@ func (a *ExactPercentileAgg) ValueFloat() float64 {
...
+func (s *ExactPercentileSelector) NewBoolSelector() execute.DoBoolRowSelector {
+	return nil
+}
+
+func (s *ExactPercentileSelector) NewIntSelector() execute.DoIntRowSelector {
+	return nil
+}
+
+func (s *ExactPercentileSelector) NewUIntSelector() execute.DoUIntRowSelector {
+	return new(MinUIntSelector)
```
nathanielc: We should be able to select on `bool`, `int` and `uint` types.

query/functions/percentile.go
```diff
@@ -14,12 +14,19 @@ import (
 )
 
 const PercentileKind = "percentile"
-const ExactPercentileKind = "exact-percentile"
+const ExactPercentileAggKind = "exact-percentile-aggregate"
+const ExactPercentileSelectKind = "exact-percentile-selector"
+
+const (
+	method_estimate_tdigest = "estimate_tdigest"
```
nathanielc: Can you camelCase style names for these constants identifiers not the values?

query/functions/percentile.go
```diff
@@ -113,34 +121,61 @@ func (s *PercentileProcedureSpec) Copy() plan.ProcedureSpec {
...
 	spec, ok := qs.(*PercentileOpSpec)
 	if !ok {
 		return nil, fmt.Errorf("invalid spec type %T", qs)
 	}
-	if spec.Exact {
-		return &ExactPercentileProcedureSpec{
+
+	switch spec.Method {
+	case method_estimate_tdigest:
+		return &PercentileProcedureSpec{
```
nathanielc: Should we rename `PercentileProcedureSpec` to `TDigestPercentileProcedureSpec`?

query/functions/percentile.go
```diff
@@ -113,34 +121,61 @@ func (s *PercentileProcedureSpec) Copy() plan.ProcedureSpec {
...
+		return &ExactPercentileAggProcedureSpec{
+			Percentile:      spec.Percentile,
+			AggregateConfig: spec.AggregateConfig,
+		}, nil
+	case method_exact_selector:
+		return &ExactPercentileSelectProcedureSpec{
+			Percentile: spec.Percentile,
+		}, nil
+
+	default:
```
nathanielc: Since the `default` case is the same as `method_estimate_tdigest` can you make the `method_estimate_tdigest` case fallthrough to the default?

PR #448, opened 2018-07-20
chore: Specify calendar intervals, windowing and time zone behavior

### LINKED ISSUE #447: Write up docs for window function, dealing with incomplete windows

nathanielc:
## Changes

There are a few fundamental changes the PR introduces:

1. A `timeZone` options is added to set the context of the entire script.
2. Durations are not a single integer value, they are a set of three values: nanoseconds, days and months.
3. New types were introduced: The `interval` which is a length of time with specific start and stop dates. And the `generator` types which produce an unknown number of values. Together these concepts can represent a set of intervals of unknown size.
4. The window function accepts a set of intervals (via a generator or otherwise) that dictate how to window data.
5. The idea of `interval comprehensions` are added which allow a user to define arbitrary natural intervals of time. These comprehensions can describe fortnights, quarters, fiscal years that start in July, the hours of 9-5 on Mon-Fri, among many others.

Additionally the defaults of the window function have been updated to make it so that window bounds align with the `now` time of the query so incomplete windows are not returned. This is still configurable, but by default ` range(start:-1h) |> window(every: 1m)` will return exactly 60 windows.

## Implementation notes

This PR contains the changes to the language specification. What follows are some notes on implementation. These are not final and are only here to make some of the ideas more concrete.

### Durations

The current implementation of duration is a int64 as nanoseconds since the epoch.
The proposed change is to represent durations as three independent values: nanoseconds, days and months. These three values are independent, meaning that no amount of nanoseconds is equal to a day and no amount of days is equal to a month. In other words the three values of nanoseconds, days and months are a linearly independent basis for all possible time values (on a Gregorian calendar). So we can treat each duration as a 3 vector in a 3 dimensional space, where the dimensions are nanoseconds, days, months. This means that we can add and subtract arbitrary durations without loss of generality.

Some example durations:
```
type Duration struct {
   Nanoseconds int64
   Days int64
   Months int64
}

var week = Duration{Days: 7} 
var fortnight = Duration{Days: 14} 
week + week == fortnight // true
var year = Duration{Months: 12}
```

See PR content for examples using Flux duration literals.

### Generators

I am thinking the interface for Generators looks like:

```
type Generator interface {
    // Generate produces the next value in the sequence or returns false if there are no more values.
    Generate() (values.Value, bool)
}
```

### Interval Generators

This type will be the implementation behind interval comprehensions
```
type IntervalGenerator struct {
    Start, Stop Time
    Every, Period, Offset Duration
}

func (g *IntervalGenerator) Generate() (values.Value, bool) {
    // .... produce the next interval in the set.
}
```

Closes #447

### DIFF
--- a/query/docs/SPEC.md
+++ b/query/docs/SPEC.md
@@ -166,7 +166,7 @@ It has an integer part and a duration unit part.
 Multiple duration may be specified together and the resulting duration is the sum of each smaller part.
 
     duration_lit        = { int_lit duration_unit } .
-    duration_unit       = "ns" | "u" | "µ" | "ms" | "s" | "m" | "h" | "d" | "w" .
+    duration_unit       = "ns" | "us" | "µs" | "ms" | "s" | "m" | "h" | "d" | "w" | "mo" | "y" .
 
 | Units    | Meaning                                 |
 | -----    | -------                                 |
@@ -176,11 +176,18 @@ Multiple duration may be specified together and the resulting duration is the su
 | s        | second                                  |
 | m        | minute (60 seconds)                     |
 | h        | hour (60 minutes)                       |
-| d        | day (24 hours)                          |
+| d        | day                                     |
 | w        | week (7 days)                           |
+| mo       | month                                   |
+| y        | year (12 months)                        |
 
-Durations represent a fixed length of time.
-They do not change based on time zones or other time related events like daylight savings or leap seconds.
+Durations represent a length of time.
+Durations can be combined via addition and subtraction.
+Durations can be multiplied by an integer value.
+
+Durations track the basic units of seconds, days and months independently.
+No amount of seconds is equal to a day, as days vary in their number of seconds.
+No amount of days is equal to a month, as months vary in their number of days.
 
 Examples:
 
@@ -188,6 +195,8 @@ Examples:
     10d
     1h15m // 1 hour and 15 minutes
     5w
+    1mo5d // 1 month and 5 days
+
 
 [IMPL#311](https://github.com/influxdata/platform/query/issues/311) Parse duration literals
 
@@ -208,7 +217,6 @@ The format follows the RFC 3339 specification.
     second            = decimal_digit decimal_digit .
     fractional_second = "."  { decimal_digit } .
     time_offset       = "Z" | ("+" | "-" ) hour ":" minute .
-
 
 #### String literals
 
@@ -337,6 +345,14 @@ The time type name is `time`.
 A _duration type_ represents a length of time with nanosecond precision.
 The duration type name is `duration`.
 
+Durations can be added to times to produce a new time.
+
+Examples:
+
+    2018-07-01T00:00:00Z + 1mo // 2018-08-01T00:00:00Z
+    2018-07-01T00:00:00Z + 2y  // 2020-07-01T00:00:00Z
+    2018-07-01T00:00:00Z + 5h  // 2018-07-01T05:00:00Z
+
 #### String types
 
 A _string type_ represents a possibly empty sequence of characters.
@@ -368,6 +384,13 @@ A _function type_ represents a set of all functions with the same argument and r
 
 
 [IMPL#315](https://github.com/influxdata/platform/query/issues/315) Specify type inference rules
+
+#### Generator types
+
+A _generator type_ represents a value that produces an unknown number of other values.
+The generated values may be of any other type but must all be the same type.
+
+[IMPL#XXX](https://github.com/influxdata/platform/query/issues/XXX) Implement generators
 
 ### Blocks
 
@@ -542,8 +565,16 @@ Grammatically, an option statement is just a variable assignment preceded by the
 
 Below is a list of all options that are currently implemented in the Flux language:
 
-* task
 * now
+* task
+* location
+
+##### now
+
+The `now` option is a function that returns a time value to be used as a proxy for the current system time.
+
+    // Query should execute as if the below time is the current system time
+    option now = () => 2006-01-02T15:04:05-07:00
 
 ##### task
 
@@ -557,12 +588,14 @@ The `task` option is used by a scheduler to schedule the execution of a Flux que
         retry: 5,           // number of times to retry a failed query
     }
 
-##### now
+##### location
 
-The `now` option is a function that returns a time value to be used as a proxy for the current system time.
+The `location` option is used to set the time zone of all times in the script.
+The location maps the UTC offset in use at that location for a given time.
+The default value is the location of the running process.
 
-    // Query should execute as if the below time is the current system time
-    option now = () => 2006-01-02T15:04:05Z07:00
+    option location = fixedZone(offset:-5h) // set timezone to be 5 hours west of UTC
+    option location = loadLocation(name:"America/Denver") // set location to be America/Denver
 
 #### Return statements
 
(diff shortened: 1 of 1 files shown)


query/docs/SPEC.md
```diff
@@ -210,6 +219,54 @@ The format follows the RFC 3339 specification.
...
+
+    {1h}        // 1 hour intervals
+    {1h:2h}     // 2 hour long intervals every 1 hour
+    {1h:2h:30m} // 2 hour long intervals every 1 hour starting at 30m past the hour
+    {1w:1w:1d}  // 1 week intervals starting on Monday (by default weeks start on Sunday)
+    {1d:-1h}    // the hour from 11PM - 12AM every night
+
+Examples using a predicate:
+
+    {1d:1d (t) => !(weekday(t:t) in {Sunday, Satruday})}     // 1 day intervals excluding weekends
```
nathanielc: Should we require that the function be literals? Or should we allow this:

```
weekends = (t) => !(weekday(t:t) in {Sunday, Satruday})
{1d:1d weekends}
```

I think we should.

query/docs/SPEC.md
```diff
@@ -210,6 +219,54 @@ The format follows the RFC 3339 specification.
...
+The optional second duration literal is called the "period" duration.
+It is the length of each interval and it defaults to the value of the "every" duration.
+The "period" duration can be negative, indicating the start and stop boundaries are reversed.
+The optional third duration literal is called the "offset" duration.
+It is the offset from the time zone epoch time, it defaults to time zone offset of the `now()` time.
+To specify the "offset" duration both the "every" and "period" durations must be specified.
+
+The optional predicate allows for the exclusion of some of the intervals based on the predicate logic.
+
+    IntervalComprehension = "{" duration_lit { ":" duration_lit { ":" duration_lit } } { FunctionLit } "}" .
```
nathanielc: Should we use  a different bracket? Maybe `<>` since `{}` are overloaded already?

query/docs/SPEC.md
nathanielc: We could also allow for multiple types of predicates:

1. (start) => ... // accepts only the start time of the interval in question
2. (start, stop) => ... // accepts both the start and stop times of the interval.

Or optionally we could pass the interval directly as an object.
```
weekends = (interval) => !(weekday(t:interval.start) in {Sunday, Satruday})
```

query/docs/SPEC.md
```diff
@@ -337,6 +394,12 @@ The time type name is `time`.
 A _duration type_ represents a length of time with nanosecond precision.
 The duration type name is `duration`.
 
+#### Interval types
+
+An _interval type_ represents a length of time with explicit start and stop times.
+The start time is inclusive and the stop time is exclusive.
+The interval type name is `interval`.
```
nathanielc: Do we need this as an explicit type? Or can we just pass around objects with `start` and `stop` properties?

query/docs/SPEC.md
```diff
@@ -210,6 +219,54 @@ The format follows the RFC 3339 specification.
...
 
+#### Interval literals
+
+An interval literal is a representation of a length of time with absolute start and stop times.
+It has start and stop times.
+
+    interval_lit = "(" date_time_lit "," date_time_lit ")" .
+
+Intervals represent a length of time with explicit start and stop times.
+The start time is inclusive and the stop time is exclusive.
```
nathanielc: Do we need explicit interval literals? Or can we use objects `{start: ...., stop: ...}`?

query/docs/SPEC.md
```diff
@@ -557,12 +643,19 @@ The `task` option is used by a scheduler to schedule the execution of a Flux que
...
 
-    // Query should execute as if the below time is the current system time
-    option now = () => 2006-01-02T15:04:05Z07:00
+    option timeZone = -5h // set timezone to be 5 hours west of UTC
+    option timeZone = loadTimeZone(name:"America/Denver") // set timezone to be America/Denver
+
+The time zone "epoch" time is defined as, the time zone offset added to the Unix epoch (1970-01-01T00:00:00Z).
+
+QUESTION: Using the offset is simple, but what about daylight savings time, when the offset changes depending on the date?
+Does timeZone need to be a function that return the offset given a date?
```
docmerlin: This will change calculations a lot if you don't do this.  When people want to know the sum of X from July 6th to Dec 4th in their time zone, without accounting with daylight savings time, it will get a difference answer if you do loadTimeZone(name:"America/Denver") vs if you do an offset.  People usually want the actual timezone not the offset.  Some systems don't do this properly (for example Periscope), which causes a lot of confusion in BI circles.

jsternberg:
The generator interface will likely need to output more than one value. It will probably need a start and stop time to be returned which would correspond to the `_start` and `_stop` values. Otherwise, I like the generator interface and we can work out specifics during implementation.

query/docs/SPEC.md
```diff
@@ -1147,25 +1303,45 @@ A single input record will be placed into zero or more output tables, depending
...
 * `stopCol` string
-    Name of the column containing the window stop time. Defaults to `_stop`.
+    Name of the column containing the window stop time.
+    Defaults to `_stop`.
 
-[IMPL#319](https://github.com/influxdata/platform/query/issues/319) Remove concept of Bounds from tables
+Examples:
+
+    window(every:1h) // window the data into 1 hour intervals
+    window(intervals: {1d:8h:9h}) // window the data into 8 hour intervals starting at 9AM every day.
```
pauldix: I'm not a fan of this syntax, it's not clear what's going on. You can't read it and understand that's what it means.

query/docs/SPEC.md
pauldix: This section is a bit confusing to me. Can you show some examples on how interval comprehensions should be used?

pauldix:
@nathanielc We don't care about leap seconds. Let the user handle it. For time zones, the location is important because of the dst and all that. A simple offset won't work. I like to think of this less like durations and more like start and stop times in a given time zone which could be figured as a unix epoch.

I like the default interval sets you have, except I'd probably exclude `workHours` from that list.

query/docs/SPEC.md
```diff
@@ -176,18 +176,53 @@ Multiple duration may be specified together and the resulting duration is the su
...
+    // Addition and subtraction of durations to date times does not commute
+    2018-02-28T00:00:00Z + 1mo + 1d // 2018-03-29T00:00:00Z
+    2018-02-28T00:00:00Z + 1d + 1mo // 2018-04-01T00:00:00Z
+    2018-01-01T00:00:00Z + 3mo - 1d // 2018-02-28T00:00:00Z
+    2018-01-01T00:00:00Z - 1d + 3mo // 2018-03-03T00:00:00Z, December 31st + 3m0 is February 31st which is normalized to March 3 in 2018.
+
+    // Addition and subtraction of durations to date times applies months, days and seconds in that order.
+    2018-02-28T00:00:00Z + 1mo + 1d // 2018-03-29T00:00:00Z
+    2018-02-28T00:00:00Z + 1mo1d    // 2018-03-29T00:00:00Z
+    2018-02-28T00:00:00Z + 1d + 1mo // 2018-04-01T00:00:00Z, explicit left associative add of 1d first changes the result
```
nathanielc: @pauldix Does these examples make sense? I have show where the happy path as well as were this model breaks down.

pauldix: Yep, these make sense!

PR #459, opened 2018-07-23
Add increase to flux as built in function

### LINKED ISSUE #302: Add support for increase function
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#217  

https://github.com/influxdata/influxdb/issues/7076

Probably looks something like this:

```
increase = (table=<-) => 
    difference(table, nonNegative:true)
        |> cumulativeSum()
```

See also https://prometheus.io/docs/querying/functions/#increase.

adamperlin:
Resolves #302. Adds `increase(table=<-, columns=["_value"])` to flux as a built-in function. `increase` flux test cases have been updated to specifically check for behavior referenced in https://github.com/influxdata/ifql/issues/217#issuecomment-379323490.


query/functions/increase.go
```diff
@@ -0,0 +1,18 @@
+package functions
+
+import "github.com/influxdata/platform/query"
+
+func init() {
+	query.RegisterBuiltIn("increase", increaseBuiltIn)
```
aanthony1243: This is really great example code.  @nathanielc  @pauldix  I might be over-reacting but this seems blog-worthy that we can inject flux code as a permanent builtin with such minimal effort.

query/functions/increase.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+var increaseBuiltIn = `
+// Increase returns the total non-negative difference between values in a table. 
+// A main usage case is tracking changes in counter values which may wrap over time when they hit 
+// a threshold or are reset. In the case of a wrap/reset,
+// we can assume that the absolute delta between two points will be at least their non-negative difference.
+
```
nathanielc: nit, can you remove this empty line? Later we will want to associate this comment as the "doc" for the function, an empty line will make that association hard.

query/functions/increase.go
```diff
@@ -0,0 +1,18 @@
...
+increase = (table=<-, columns = ["_value"]) =>
```
nathanielc: nit on style, can you remove the spaces around the `=` operator?

query/functions/increase.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+var increaseBuiltIn = `
+// Increase returns the total non-negative difference between values in a table. 
+// A main usage case is tracking changes in counter values which may wrap over time when they hit 
+// a threshold or are reset. In the case of a wrap/reset,
+// we can assume that the absolute delta between two points will be at least their non-negative difference.
+
+increase = (table=<-, columns = ["_value"]) => 
+    difference(table: table, nonNegative: true, columns:columns)
```
nathanielc: Also a style nit, can you use the `|>` operator even with the `table` identifier?

i.e. `table |> difference(...)`

We want to be consistent, specially in our built in code.

adamperlin:
Formatting changes made. I'll merge this.

PR #460, opened 2018-07-24
Fix initial transpiler grouping so that aggregates will compute

aanthony1243:
Added a test for mean() to confirm that we need to have _start in the default group key in order for aggregates to run without errors.  Spec tests updated according to change.  Compiler tests pass without modification.


jsternberg:
Can you update the transpiler README to make note of this change?

Issue #461, opened 2018-07-24
Design Rename/Drop column function


aanthony1243:
helper function to rename and/or drop columns from a table.

PR #462, opened 2018-07-24
Add spec for flux rename/drop function

adamperlin:
See #461. Adds user-level specification for a combined rename and drop function. This is just a concept for discussion. A high-level note on design: it seems like rename and drop operations are very often used together, so it might be best to join them into one function for user convenience.


adamperlin:
According to this spec, your example would cause an error. You wouldn't be able to mix `dropCols` and `dropOthers`.

adamperlin:
However, if it's too bizarre/unclear to have the drop arguments at all, we could just do two separate functions.

nathanielc:
> You wouldn't be able to mix dropCols and dropOthers

That is kind of my point, if the combination of the functions makes it such that you can only use the function in special ways, then I say we break up the function. Curious what @aanthony1243 thinks here?

aanthony1243:
also, we discussed briefly that internally renaming and dropping columns are closely related operations wrt the table API so that we might implement both functions using the same underlying code.

nathanielc:
> also, we discussed briefly that internally renaming and dropping columns are closely related operations wrt the table API so that we might implement both functions using the same underlying code.

Yes, agreed here, I expect a single transformation to implement both of these.

I dislike the `keepUnlisted` argument to the rename function. Kapacitor has a similar function and it confused many users when a rename like operation was also removing columns.

I do like the two argument version of drop. I think the args should be named `columns` and `except` to follow similar arguments in other functions.

aanthony1243:
I don't love it enough to fight for it.  @adamperlin what do you think about simplifying rename to having just the single map parameter `columns`   and then also adding a section for `drop` as discussed above?

pauldix:
I'm not a fan of the keep unlisted. I'd prefer to see different functions like this:

```
drop(cols: ["foo", "bar"])
drop(match: /some regex/)
keep(cols: ["asdf", "jkl])
keep(cols: [])
keep(match: /foo/)

rename(cols: {"old":"new"})
rename(from: "foo", to:"bar")
```

adamperlin:
I think that's a good idea. Having an additional drop function adds much
more clarity, and rename really only needs one parameter. I like the idea
of the `keep` function as well - it seems like a much clearer and more
explicit way of specifying columns to ignore.

adamperlin:
I'm unclear as to what operation `keep` would actually perform on a table. More of an implementation detail but @pauldix  were you thinking keep would be pushed down into `drop`?

pauldix:
@adamperlin I updated my comment a bit and noticed you responded via email, so you'll probably want to reread it here on GH. On your keep question, my idea is that it would be to keep any of the specific columns. Both `drop` and `keep` have the same arguments, `cols` and `fn`. One and only one of those arguments is required.

drop could be implemented using keep or vice versa. You'd just put a not (`!`) on the return boolean of the passed in `fn`

adamperlin:
Ok. That makes sense. For some reason I was thinking keep would be syntactic sugar for an `except` argument to drop. But having keep be inverse of drop seems like a much better idea.Thanks for clarifying!

pauldix:
Actually, now that I think about this I think it would be better to go with the `select` and `reject` paradigm. The only trick then is that select and reject could also make sense applied to rows rather than columns. Thoughts?

aanthony1243:
@pauldix   I think the same could be said for `drop`  and `keep`.  I think that the names alone are not enough, in a very golang type of way.  `select(columns: ["c1"])`  makes it clear that it's column-wise.  

I think true relational-DB theory experts well-versed in relational algebra will revolt against using `select` to pick columns from a table, but your average SQL user will find it to be a natural thought pattern.  

`reject` is less familiar to me as a keyword in this context. I'd rather have `select`  and `drop` if we went that route.

pauldix:
Let's go with `drop` and `keep` and use `fn` for the function argument. For row operations we already have `filter` which will do everything we need there.

adamperlin:
@aanthony1243 @nathanielc I've updated the spec to reflect what was decided. Let me know what you think.

query/docs/SPEC.md
```diff
@@ -1084,6 +1084,61 @@ Range has the following properties:
...
+Example usage:
+
+Rename a single column: `rename(columns:{host: "server"})`
+
+Rename all columns with `fn` parameter: `rename(fn: (col) => "{col}_new")`
+
+#### Drop 
+
+Drop will exclude specified columns from a returned data table. Columns to exclude can be specified either through a 
+list, or a predicate function.
```
nathanielc: What happens to a column that is part of the group key?

query/docs/SPEC.md
```diff
@@ -1084,6 +1084,61 @@ Range has the following properties:
     Specifies the exclusive newest time to be included in the results.
     Defaults to "now"
 
+#### Rename 
+
+Rename will rename specified columns in a data table. 
+There are two variants: one which takes a map of old column names to new column names,
+ and one which takes a mapping function.
```
nathanielc: What happens to a column that is part of the group key?

query/docs/SPEC.md
```diff
@@ -1084,6 +1084,61 @@ Range has the following properties:
     Specifies the exclusive newest time to be included in the results.
     Defaults to "now"
 
+#### Rename 
+
+Rename will rename specified columns in a data table.
```
nathanielc: nit, lets refer to tables consistently, no where currently in the Spec is a table referred to as a `data table`, which makes the user question if a `data table` is a special kind of table. Let's just call them `tables` without the extra undefined adjective. There are a few cases of this in the change.

query/docs/SPEC.md
```diff
@@ -1084,6 +1084,61 @@ Range has the following properties:
...
+
+Example Usage:
+
+Drop a list of columns: `drop(columns: ["host", "_measurement"])`
+
+Drop all columns matching a predicate: `drop(fn: (col) => col =~ /usage*/)`
+
+#### Keep 
+
+Keep can be thought of as the inverse of drop. It will return a table containing only columns that are specified,
```
nathanielc: nit, `Keep is the inverse of drop.`

PR #467, opened 2018-07-25
fix(query/influxql): influxql transpiler should set the now time

jsternberg:
The now time is stamped by the influxql transpiler and used inside of
the actual query. It will result in more accuracy if we take the
timestamp we have created and send it as part of the spec to queryd
rather than force ourselves to ensure absolute times exist everywhere.


jsternberg:
queryd will now use that time. It's a new thing because of the changes to make the now option work which is really good because we need it.

PR #468, opened 2018-07-25
controller sets Now time before enqueuing a query

jlapacik:
A transpiled query is submitted to the controller but is not produced by the Flux interpreter. Hence it cannot take advantage of the `now` option in order to set its `Now` field. The controller must explicitly check that the Now field of a query spec is set, and if not, must set it before enqueuing it.

### DIFF
--- a/query/control/controller.go
+++ b/query/control/controller.go
@@ -81,12 +81,19 @@ func (c *Controller) QueryWithCompile(ctx context.Context, orgID platform.ID, qu
 	return q, err
 }
 
-// Query submits a query for execution returning immediately.
-// The spec must not be modified while the query is still active.
-// Done must be called on any returned Query objects.
+// Query submits a query for execution returning immediately. Once a query
+// is submitted, that is once a query is enqueued, the query spec must not
+// be modified. Done must be called on any returned Query objects.
 func (c *Controller) Query(ctx context.Context, orgID platform.ID, qSpec *query.Spec) (query.Query, error) {
 	q := c.createQuery(ctx, orgID)
 	q.spec = *qSpec
+
+	// Incoming query spec may have been produced by an entity other than the
+	// Flux interpreter, so we must set the default Now time if not already set.
+	if q.spec.Now.IsZero() {
+		q.spec.Now = q.now
+	}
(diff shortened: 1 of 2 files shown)


jlapacik:
I would rather we made `Now` required and called `Validate` instead of having the planner check if `Now` is set. That just seems unrelated to the planner?

query/spec.go
```diff
@@ -45,6 +45,9 @@ func (q *Spec) Walk(f func(o *Operation) error) error {
 
 // Validate ensures the query is a valid DAG.
 func (q *Spec) Validate() error {
+	if q.Now.IsZero() {
+		return errors.New("query has no Now time")
+	}
```
jlapacik: @jsternberg this enforces that all queries must have the Now field set. A query will be rejected otherwise.

query/control/controller.go
```diff
@@ -81,11 +81,18 @@ func (c *Controller) QueryWithCompile(ctx context.Context, orgID platform.ID, qu
...
+// Query submits a query for execution returning immediately. Once a query
+// is submitted, that is once a query is enqueued, the query spec must not
+// be modified. Done must be called on any returned Query objects.
 func (c *Controller) Query(ctx context.Context, orgID platform.ID, qSpec *query.Spec) (query.Query, error) {
 	q := c.createQuery(ctx, orgID)
+
+	// Incoming query spec may have been produced by an entity other than the
+	// Flux interpreter, so we must set the default Now time if not already set.
+	if qSpec.Now.IsZero() {
+		qSpec.Now = q.now
```
jsternberg: Can you do this after the copy? That way, you aren't modifying the incoming argument.

query/spec.go
jsternberg: A nitpick: Can this error be "now time must be set". I think it's more straightforward and I don't like uppercase letters. It also has the additional bonus that it shows up as `"now"` in the JSON and not `"Now"`, but I don't think a user should ever see this message anyway.

query/control/controller.go
jlapacik: This is now updated.

PR #471, opened 2018-07-25
feat(http): Add /ping endpoint to fluxd

### LINKED ISSUE #470: Ping endpoint for Flux
In Chronograf, we are trying to validate a connection to flux. Right now we are doing it by making request with a query and seeing if it returns anything.

ischolten:
Closes #470 

Problem: We needed a way to verify that a flux connection was valid before updating or creating it on the Chronograf side. 

Solution: Add a ping endpoint for fluxd.

### DIFF
--- a/http/query_service.go
+++ b/http/query_service.go
@@ -39,8 +39,15 @@ func NewQueryHandler() *QueryHandler {
 		csvEncoder: csv.NewMultiResultEncoder(csv.DefaultEncoderConfig()),
 	}
 
+	h.HandlerFunc("GET", "/ping", h.handlePing)
 	h.HandlerFunc("POST", queryPath, h.handlePostQuery)
 	return h
+}
+
+// handlePing returns a simple response to let the client know the server is running.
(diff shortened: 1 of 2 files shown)


http/query_service.go
```diff
@@ -39,10 +39,17 @@ func NewQueryHandler() *QueryHandler {
...
 	}
 
+	h.HandlerFunc("GET", "/ping", h.handlePing)
 	h.HandlerFunc("POST", queryPath, h.handlePostQuery)
 	return h
 }
 
+// handlePing returns a simple response to let the client know the server is running.
+func (h *QueryHandler) handlePing(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
```
nathanielc: Why do we set the content type if there is no content?

ischolten: Some browsers require it.

nathanielc: Can you add a comment to that fact? Otherwise it will probably get removed in the future since it "shouldn't" be necessary.

Issue #454, opened 2018-07-23
Update Join Spec


nathanielc:
Update the query SPEC for the design for Join https://github.com/influxdata/platform/issues/407

mark-rushakoff:
Looks like this was meant to be closed by #475.

PR #475, opened 2018-07-25
new join implementation

jlapacik:
This PR removes the `fn` parameter from the join and constructs the schema of the output table in a consistent fashion. It also allows joins between tables with different group keys. This new join implementation is just for the inner join which replaces the entire functionality of the previous join. New join methods (cross product and such) will be added in a separate commit.


query/functions/join.go
```diff
@@ -282,56 +291,46 @@ func (t *mergeJoinTransformation) Process(id execute.DatasetID, tbl query.Table)
...
-				break
-			}
-		}
-		if !found {
-			u = append(u, a)
-		}
+	// Create the schema for the output table of the join if not already set
+	if tables.left != nil && tables.right != nil && tables.key == nil {
+		tables.constructJoinSchema()
+		t.cache.Data().Set(tables.key, tables)
```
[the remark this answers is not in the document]
nathanielc: Also I think there is a bug here that fixing the test case I mentioned will expose.

The code as it is written can only produce a single output table with a single group key. Adding/Fixing a testcase that produces more than one output table should expose this.

I think the way this could work is to dynamically create new tables when the code discovers a new key.
Then do a new Set each time that happens.

jlapacik:
I made a few changes to the implementation of the join operator the details of which are as follows.

* All data for a join is held by the `mergeJoinCache` https://github.com/influxdata/platform/blob/20493c96432b387afb7dfcddd7016ef2ec06488e/query/functions/join.go#L354

* The main containers for holding incoming data are the `leftStream` and `rightStream` fields https://github.com/influxdata/platform/blob/20493c96432b387afb7dfcddd7016ef2ec06488e/query/functions/join.go#L361

* Once triggered, the records in each of theses left and right caches are sorted according to the join columns, merged and placed into an output cache https://github.com/influxdata/platform/blob/20493c96432b387afb7dfcddd7016ef2ec06488e/query/functions/join.go#L363

* An `execute.Allocator` is passed to each of the builders in the `mergeJoinCache` https://github.com/influxdata/platform/blob/20493c96432b387afb7dfcddd7016ef2ec06488e/query/execute/stream.go#L32

  This way the memory consumed by the both the left and right stream caches can be tracked as well as freed immediately after execution of the join. The memory used by the output stream cache is freed using the `DiscardTable()` and `ExpireTable()` methods https://github.com/influxdata/platform/blob/20493c96432b387afb7dfcddd7016ef2ec06488e/query/functions/join.go#L415

query/functions/join.go
```diff
@@ -382,230 +343,337 @@ func (t *mergeJoinTransformation) Finish(id execute.DatasetID, err error) {
...
+	return c.outStream.Table(key)
 }
 
+// ForEach iterates over each table in the output stream
 func (c *mergeJoinCache) ForEach(f func(query.GroupKey)) {
-	c.data.Range(func(key query.GroupKey, value interface{}) {
-		f(key)
-	})
+	if c.leftStream != nil && c.rightStream != nil && c.leftStream.NRows() > 0 && c.rightStream.NRows() > 0 {
+		c.Join()
```
nathanielc: I don't think this will work. The `Join` method both produces new values and clears the stream cache of any buffered data. We cannot perform those destructive actions just because we want to iterate of the cached data. Right now this works because we only have events for UpdateWatermark that iterate through the cached data, and those events almost always cause the tables to be materialized. (Also I do not think our tests are using the watermarks, instead they are just relying on Finish to flush once all data has been processed, we should probably change that).

I think this is really close to how it could work. Instead of checking here during iteration of the cache, if we proactively checked for data ready to be joined in the `Process` method I think this idea will work.

I am thining something like in the process method we first check if we know the join key for the data and append the new data to that cached table. If we do not know the join key we place it in some kind of limbo cache, (sort of like your stream cache already). Then we check all the data in the limbo cache if we can determine any group keys (i.e. tables) concretely. If so we move those rows into that table cache. But we do not perform the join until the cached table is ready to be materialized, i.e. during the `Table()` method call on the mergeJoinCache.

In summary there are two caches, the limbo cache that caches the left and right rows. The table cache which has its own list of left and right rows. If a new table arrives we can append directly to this cache if we know the group key. When performing the `For*` methods on the cache we iterate over these cached tables. Then as new tables arrive after each we check if we can promote any data from the limbo cache into the table cache.

jlapacik:
Joins are a very interesting beast in stream processing engines. By default you can't get rid of data that has been joined because that same data might join with upstream data that is yet to flow through the operator. That makes the join operator a massive consumer of memory. However there are a few times when you can get rid of previously joined data because you know it will never be joined with any upstream data. I list all such cases below:

* Problem Setup

    * There are two streams of tables being joined - `LeftStream` and `RightStream`
    * Each table in `LeftStream` has a group key `leftKey`
    * Each table in `RightStream` has a group key `rightKey`
    * When a record from `LeftStream` is joined with a record from `RightStream`, a new record is
       produced with a new group key `outKey`
    * The join operation contains two buffers for storing the tables of `LeftStream` and `RightStream` respectively
    * Each buffer is a map that associates a group key with the table corresponding to that group key
    * For any `outKey` there is a single `leftKey` and a single `rightKey` that produced that `outKey`

* The Big Question

    Given an `outKey`, we can decompose it into a `leftKey` and a `rightKey`. Without loss of generality, under what circumstances can we remove the table associated with `leftKey` from the left buffer mentioned above?

* The Big Answer

  **case 1 (Happy Path):** The columns of the join key are exactly the columns of `leftKey` and exactly the columns of `rightKey`. I.e `rightKey = leftKey = joinKey`

    We **can** remove the table (call it "table") associated with `leftKey` from the left buffer because we know there won't be an upstream table that arrives from `RightStream` that can be joined with any of the rows of "table".

  **NOTE** this also means we can remove the table associated with `rightKey` from the right buffer.

  **case 2:** The columns of the join key form a superset of the columns of `rightKey`.  
`rightKey < joinKey`

    We **can** remove all rows of the table associated with `leftKey` from the left buffer that join with the table associated with `rightKey`. This is guaranteed to be the entire table only if the columns of `rightKey` form a subset of the columns of `leftKey`.

    **NOTE** case 1 is a special case of this.

  **case 3:** The columns of the join key form a **strict** subset of the columns of `rightKey`. 
`joinKey < rightKey`

    We **cannot** remove the table (call it "table") associated with `leftKey` from the left buffer. This is because upstream data may arrive from `RightStream` that has the same value on the join columns as some of the rows in "table". Hence the data in "table" will eventually be joined with this new data from `RightStream`.

  **case 4:** The columns of the join key don't overlap with the columns of `rightKey` at all

    Same as case3.


So what we can see is that if `rightKey = leftKey = joinKey` we can very quickly (in constant time) remove data from our buffers by just removing a key mapping. If we are in case2 but not case1, we can still remove data but at a linear rate. If we are in case3 or 4 unfortunately we cannot remove any data from our buffers.

nathanielc:
I have one follow up question:

Given a set of `outKey`s. (i.e. we have had multiple ExpireTable calls). For case 3 is there a set of outKeys that span the entire range of `rightKey` that we can then declare that subset of right tables can be removed? If so what is that mapping? Is it computationally reasonable to compute?

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
-* `tables` map of tables
-    Map of tables to join. Currently only two tables are allowed.
-* `on` array of strings
-    List of columns on which to join the tables.
-* `fn`
-    Defines the function that merges the values of the tables.
-    The function must defined to accept a single parameter.
-    The parameter is an object where the value of each key is a corresponding record from the input streams.
-    The return value must be an object which defines the output record structure.
+    | _time | temperature |
```
jsternberg: Serious question: would this schema ever happen? To my recollection, `from()` always returns a field value as `_value` so the only way this could happen is if a column was renamed before the join. Can the example in the spec show `_value` instead as that is more likely to be seen?

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+
+##### options
+
+The join operator accepts the following named parameters:
+
+| Name | Type | Required | Default Value |
+| ---- | ---- | -------- | ------- |
+| tables    | map           | yes   | no default value - must be specified with every call |
+| on        | string array  | no    | list of common columns among the input streams |
+| method    | string        | no    | "inner" |
```
jsternberg: Can we get a column for the acceptable values for enumerated values like this one? I'm not sure if we need another column for it since the other two wouldn't, but it would help if the accepted values are in this table. I understand that they are explained below.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+* method
+
+    **method** is an optional parameter that specifies the type of join to be performed. By default it
+    is set to `"inner"`. The **method** parameter may take on any one of the following values:
+
+    * inner - The result of the join is the set of all combinations of tuples in the input streams that are either:
+        1. Equal on their common or shared columns (when **on** is not present)
+        2. Equal on the columns specified by the **on** parameter
+
+    * cross - The result of the join is the set of all combinations of tuples in the input streams, i.e the cross product
```
jsternberg: Can we get `outer`, `left`, and `right` in here? I'm not sure what `left` and `right` would mean exactly (maybe the argument should be different?), but we're going to need some way to implement the equivalent of those types of joins. I believe outer and left are specced out to be used by the influxql transpiler if I remember correctly.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+| method    | string        | no    | "inner" |
+
+* tables
+
+    Map of tables (or streams) to join together. It is the one required parameter of the join.
+
+* on
+
+    An optional parameter for specifying a list of columns common among the input streams.
+    When present, the join operator will only combine those tuples in the intput streams that are equal
```
jsternberg: intput -> input

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+    on the columns specified by the **on** parameter. Only columns that are shared among **all** of the input
+    streams are allowed in the **on** parameter. If the **on** parameter is present, the **method**
```
jsternberg: > The **on** parameter and the `cross` **method** are mutually exclusive.

The current wording is too verbose and repeats itself.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+
+The output will be:
+
+| _time | temperature | humidity |
+| ----- | ----------- | -------- |
+| 0001  | 100.3 | 55.4 |
+| 0002  | 100.1 | 56.1 |
+| 0003  | 98.8  | 55.9 |
+
+As outlined by the previous example, the default behavior of the join operator is that of the natural join.
```
jsternberg: This seems like a mistake. The most common query I expected is something like this:

```
user = from(db: "telegraf") |> range(start: -5m) |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_user")
system = from(db: "telegraf") |> range(start: -5m) |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_user")
join(tables: {user_cpu: user, system_cpu: system})
```

If I do a natural join on this, it will not give me anything that I want because both of these tables have `_time` and `_value` columns and the `_value` column doesn't match anything and isn't expected to. It also isn't clear to me how this deals with multiple tables. Does it join across tables or do I have to perform a grouping beforehand?

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+
+    * inner - The result of the join is the set of all combinations of tuples in the input streams that are either:
+        1. Equal on their common or shared columns (when **on** is not present)
+        2. Equal on the columns specified by the **on** parameter
+
+    * cross - The result of the join is the set of all combinations of tuples in the input streams, i.e the cross product
+
+##### output schema
+
+If the **on** parameter is not present, then every column that is shared among all of the input streams will be present in
```
jsternberg: This paragraph is very confusing to read because there are too many conditionals.

I think I finally figured it out, but there's a better way you can do this. The above already states that if `on` is missing, a natural join is performed. I didn't know what that meant until I looked it up, but it appears that it just means we set `on` to be whatever columns are in common. I already pointed out the problem with that above, but I'm going to comment as if we're not changing that for the sake of clarity here.

It seems to me like you're trying to describe how the column renaming works when there are multiple columns that have the same name, but aren't being joined. But, you're trying to describe that without defining what `on` does. This sentence:

> then every column that is shared among all of the input streams

This is just describing what a natural join is without saying it's a natural join or referencing the above documentation for `on`. It would be easier to just say, "Columns that are joined on will keep their name." It says the same thing and applies for both if `on` is present and if it is not present since it references the operation instead of the parameter.

> A column whose name is unique to a specific stream will also be present in the output schema

I think this should be changed as it will create more confusion and I suspect it is possible it would cause a change in query behavior if certain data was written. All columns that aren't joined should have the same behavior in my opinion. If we don't change that, then this sentence is fine.

> any column that is common to at least two of the input streams, but not every input stream

Any conflicts that aren't joined on is what this is describing and it applies to when `on` is present too. So you can join this and the next paragraph without changing any of the meaning.

I think the behavior it is trying to say is if a column , they will all be joined and not separate. If a column is not a tag and it is unique, it will also not be changed. But, if

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+the output schema. A column whose name is unique to a specific stream will also be present in the output schema. However any
+column that is common to at least two of the input streams, but not every input stream, will be renamed in the output schema
+according to the template `<table>_<column>` for each table in which it is present.
+
+If the **on** parameter is present, then every column specified by the **on** parameter will be present in the output schema.
+A column whose name is unique to a specific stream will also be present in the output schema. However any column that is not
+specified by the **on** parameter and is common to at least two of the input streams will be renamed in the output schema
+according to the same template `<table>_<column>` for each table in which it is present.
+
+If `method = "cross"` then every column that is common to at least two of the input streams will be renamed in the output
```
jsternberg: > If the method is `cross`

The rest of the sentence is difficult to understand. It would probably be easier to simplify the renaming.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+
+    | _time | _value |
+    | ----- | ------ |
+    | 0001  | 55.4 |
+    | 0002  | 56.1 |
+    | 0003  | 55.9 |
+
+`join(tables: {a: Temperature, b: Humidity}, on: ["_time"])` produces:
+
+| _time | a__value | b__value |
```
jsternberg: This is personal preference, but I _really_ do not like this double underscore thing. @nathanielc is there anyway we could not do this? It is going to be terrible for usability and debugging. I understand we want to add an underscore to everything that is not user added, but it is really coming at a cost to the language itself by forcing these implementation details and ugliness onto users.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,128 @@ Window has the following properties:
...
+| ----- | -------- | -------- |
+| 0001  | 100.3 | 55.4 |
+| 0002  | 100.1 | 56.1 |
+| 0003  | 98.8  | 55.9 |
+
+`join(tables: {a: Temperature, b: Humidity}, method: "cross")` produces:
+
+| a__time | a__value | b_time | b__value |
```
jsternberg: `b__time`?

query/docs/SPEC.md
jlapacik: outer, left, and right are each equivalent to a cross join followed by the appropriate filter. I'd prefer to not add those here as they aren't currently implemented as part of the join operator itself.

query/docs/SPEC.md
jsternberg: I would say join on `_time` as a default is probably easiest. But natural joins likely won't work.

query/docs/SPEC.md
jlapacik: fixed.

query/docs/SPEC.md
jlapacik: Yes

query/docs/SPEC.md
jlapacik: I'm sorry, I misspoke. With cross product we "should" have everything we need to perform a left or right join. It would just involve a union and the distinct operator I think, so of course not very user friendly. I think we should definitely add them as join types, but in order to do that Flux has to have a notion of null values which currently is not implemented (although it is spec'd).  So I think we should hold off on adding those join types to the spec until we actually implement them.

jsternberg: I really would like to see left, right, and outer in here as a reference even if it isn't implemented. If it turns out cross product gives us those with a filter, then we can put that in the implementation. But the user should be able to specify them.

query/docs/SPEC.md
```diff
@@ -1231,24 +1231,106 @@ Window has the following properties:
...
+    | ----- | ------ | ------ |
+    | 0001  | "temp" | 55 |
+    | 0002  | "temp" | 56 |
+    | 0003  | "temp" | 55 |
+
+And the following join query: `join(tables: {sf: SF_Temperature, ny: NY_Temperature}, on: ["_time", "_field"])`
+
+The output will be:
+
+| _time | _field | ny__value | sf__value |
```
jsternberg: I would really like to see this be `ny_value` or `ny` by itself through a special rule for the join, but I'm not going to block this over that since I think it is reliant on a future discussion. I'll create an issue with the proposal after this is merged.

jlapacik:
Closes #511 #454

PR #476, opened 2018-07-26
Update spec and readme for the new Collate function

### LINKED ISSUE #455: [Design] Specify a collate operation for Flux
The pivot operation needs to be able to add all the 1.x fields as columns on a table.

Hopefully that operation can be described as   "pure" pivot, and then we can wrap it in helper function for the 1.x specific data model.

aanthony1243:
spec update only, no code change.  

Not in this proposal, because I'm not sure it's a good idea:  should we also implement a tie-breaker function?  That is, if duplicate entries are found for the same row+column entry,  we could apply a function  (current, new) => {...}   that determines what value to place in the final result.  This function would need to be commutative+associative but is otherwise unrestricted.  

Closes #455


query/README.md
```diff
@@ -119,6 +119,36 @@ Example: `from(db:"telegraf")`
...
+##### options
+
+* `rowKey` array of strings
+    List of columns used to uniquely identify a row. Defaults to ["_time"].
+* `colKey' array of strings
+    List of columns used to identify the set of columns to create.  Defaults to ["_field"].
+* `valueCol` column that contains the value to be collated
+    The value from each vertical entry that will be copied into a row in the collated result. Defaults to "_value".
+* `fillMap` Map of primitive types
+    Map to specify default values for each expected type in the result.  Defaults to zero-values for each type.
```
jlapacik: What about defaulting to `null` values? This way it would be clear that there was no instance of a specific `rowKey` for a specific `_field` value.

aanthony1243: we'd need support for null first. I'll check the spec and whether it defines null.  I may be mixing up proposed + actual functionality here.

query/README.md
```diff
@@ -119,6 +119,36 @@ Example: `from(db:"telegraf")`
...
+#### collate
+
+Collects values stored vertically in a stream and aligns them horizontally into logical rows.
+
+##### options
+
+* `rowKey` array of strings
+    List of columns used to uniquely identify a row. Defaults to ["_time"].
+* `colKey' array of strings
+    List of columns used to identify the set of columns to create.  Defaults to ["_field"].
```
nathanielc: How often will these defaults work without modification?

The `_field` default is 1.x specific. Should we create a 1.x wrapper around this function that applies the `_field` default and leave the `collate` function with no default?

query/README.md
```diff
@@ -119,6 +119,36 @@ Example: `from(db:"telegraf")`
...
+* `valueCol` column that contains the value to be collated
+    The value from each vertical entry that will be copied into a row in the collated result. Defaults to "_value".
+* `fillMap` Map of primitive types
```
nathanielc: Should we leave the action of filling missing data to a separate step in the query?

query/docs/SPEC.md
```diff
@@ -1169,6 +1169,36 @@ Window has the following properties:
 
 #### Collate
 
+Collate
```
nathanielc: Remove this line.

query/docs/SPEC.md
```diff
@@ -1169,6 +1169,36 @@ Window has the following properties:
...
+
+Collate collects values stored vertically in a stream and aligns them horizontally into logical rows.  The primary use-case
```
nathanielc: I like the reference to vertical and horizontal here and it helps visualize the process. So far in the spec I don't think we have referred to those terms. Maybe we can add a bit of clarity by stating cols and rows with the words vertical and horizontal to make it clear?

query/docs/SPEC.md
```diff
@@ -1169,6 +1169,36 @@ Window has the following properties:
...
+for collate is to collect common fields on a series into a single row for processing, but it may have other applications
+such as combining aggregate computations into a single row.
+
+The group key of the resulting table will be the same as the input tables, meaning that this function may be applied
+either before or after a grouping and the result will be the same.
```
nathanielc: How does this work if the `valueCol` is part of the group key? Should we explicitly deny that case?

query/README.md
aanthony1243: that makes sense, since the spec has `null` as a default.

aanthony1243:
@pauldix  There may be other use-cases.  One that came up for the transpiler today is to collate several aggregates into a single row.  But the primary use-case will be to form tuples out of fields that have the same series key+timestamp.   A common flux script would look like this: 

```
from(db:"telegraf") 
  |> range(start:-2h)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> collate(rowKey: ["_time"], colKey: ["_field"], valueCol: "_value")
```
So collate takes a table like: 

_time, _measurement, _field, _value
1, cpu, usage_user, 0.2
1, cpu, usage_system, 0.4
1, cpu, usage_idle, 0.1
2, cpu, usage_user, 0.2
2, cpu, usage_system, 0.5

and outputs: 
_time, _measurement, usage_user, usage_system, usage_idle
1, cpu, 0.1, 0.4, 0.1
2, cpu, 0.2, 0.5, null

aanthony1243:
@pauldix @nathanielc  agree about eliminating defaults.  I dropped them from the SPEC last week, but forgot to get them out of the README.md  file.

query/docs/SPEC.md
aanthony1243: I think this unnecessarily restrictive.  It is also the case, after further analysis that the colKey columns need to be dropped from the output.  I'm updating the spec now.

query/docs/SPEC.md
```diff
@@ -1225,8 +1225,42 @@ Window has the following properties:
 
 [IMPL#319](https://github.com/influxdata/platform/query/issues/319) Remove concept of Bounds from tables
 
-#### Collate
-
+#### Pivot
+
+Pivot collects values stored vertically (column-wise) in a table and aligns them horizontally (row-wise) into logical sets.  
+A common use-case for the pivot operation is to tuple-construction, where related values on a column are rearranged into a single row.
```
nathanielc: `operation is to tuple-construction,` -> `operation is to do tuple-construction,`

query/docs/SPEC.md
```diff
@@ -1225,8 +1225,42 @@ Window has the following properties:
...
+However, pivot may have other applications such as combining aggregate computations into a single row, performing joins
```
nathanielc: This part feels odd in a SPEC. I think the spec should just address what pivot does and leave use cases for other docs.

query/docs/SPEC.md
```diff
@@ -1225,8 +1225,42 @@ Window has the following properties:
...
+The row key should have a 1:1 mapping for any input row to an output row.  In the case where more than one value is
+identified for the same row+column pair in the output, the last value encountered in the stream is taken for the result.
+
+The output table will have columns based on the row key, plus the group key (minus any group key colums in the column key) 
+plus new columns for each unique tuple of values identified by the column key.  
+Any columns in the original table that are not referenced in the rowKey or the original table's group key will be dropped.  
+
+The output is constructed as follows:
+1. A new row is created for each unique value identified in the input by the rowKey parameter.
+2. The initial set of columns for the new row is the row key unioned with the group key excluding the keyCol columns, and the valueCol.
```
nathanielc: What are the `keyCol` columns? I do not see that defined anywhere.

nathanielc:
Also a formatting issue. Can you place each sentence on its own line? Then when reviewing in a diff its clear which sentence it being referenced.

query/docs/SPEC.md
```diff
@@ -1225,8 +1225,40 @@ Window has the following properties:
...
+* `colKey` array of strings
+    List of columns used to pivot values onto each row identified by the rowKey. 
+* `valueCol` string
+    Identifies the single column that contains the value to be moved around the pivot
+    
+
+The group key of the resulting table will be the same as the input tables, excluding the columns found in the colKey and valueCol. 
+This is because these columns are not part of the resulting output table.  
+
+The row key should have a 1:1 mapping for any input row to an output row.  In the case where more than one value is
```
jlapacik: Should this be `There is 1:1 mapping between each unique value of the row key and a row in the output table`?

aanthony1243: I acknowledge the wording is confusing here.  We'll encounter the same row key multiple times, but the column key will (should) be different.  If we get the same rowKey + columnKey twice, then the last one wins.  I'll fix to explain this more clearly.

query/docs/SPEC.md
```diff
@@ -1225,9 +1225,43 @@ Window has the following properties:
...
+    List of columns used to pivot values onto each row identified by the rowKey. 
+* `valueCol` string
+    Identifies the single column that contains the value to be moved around the pivot
+    
+
+The group key of the resulting table will be the same as the input tables, excluding the columns found in the colKey and valueCol. 
+This is because these columns are not part of the resulting output table.  
+
+Every input row should have a 1:1 mapping to a particular row + column in the output table, determined by its values for 
+the rowKey and colKey.
```
nathanielc: nit, this sentence doe not need to be split across lines.

query/docs/SPEC.md
```diff
@@ -1225,9 +1225,43 @@ Window has the following properties:
...
+
+The output table will have columns based on the row key, plus the group key (minus any group key colums in the column key) 
+plus new columns for each unique tuple of values identified by the column key.  
+Any columns in the original table that are not referenced in the rowKey or the original table's group key will be dropped.  
+
+The output is constructed as follows:
+1. A new row is created for each unique value identified in the input by the rowKey parameter.
+2. The initial set of columns for the new row is the row key unioned with the group key, but excluding the columns indicated by the keyCol and the valueCol.
+3. A set of value columns are added to the row for each unique value identified in the input by the columnKey parameter. 
+The label is a concatenation of the valueCol string and the keyCol values.
```
nathanielc: Is there a way to not use the `valueCol` string? Is it needed? It does not add uniqueness to the col label since all columns will have the same prefix.

query/docs/SPEC.md
```diff
@@ -1225,9 +1225,43 @@ Window has the following properties:
...
+
+
+Pivot has the following properties:
+
+* `rowKey` array of strings
+    List of columns used to uniquely identify a row for the output.
+* `colKey` array of strings
+    List of columns used to pivot values onto each row identified by the rowKey. 
+* `valueCol` string
+    Identifies the single column that contains the value to be moved around the pivot
```
nathanielc: Do these have defaults that work for the `_field` and `_value` case?

aanthony1243: no, in a previous review we decided to drop those defaults.

nikolay-t:
May I ask a question here? In addition to a separate function like Pivot(ex Collate) to put all fields as separate columns, would it be possible to change function From so that it can actually do the same thing more efficiently by special option?

Previously with SQL language, we could do something like this "SELECT a, b, a/b as "division" FROM...", and it looks simple. Now, to do math between fields (I'm not talking about math between measurement), we will need to do multiple things, like calling additional Group (to remove groping by _field column), then Pivot to complete. 

I suggest that adding explicit option to From to enable creation of separate columns for values instead of single _field/_value pair of columns can simplify query and make it more efficient. What do you think about it?

aanthony1243:
hi @nikolay-t  much of our work is currently in developing core, fundamental functions.  The operation you describe does sound convenient and it could be implemented as a compound function of  From() and Pivot().  This would certainly help the developer in being more efficient with their time and keystrokes.  

Regarding computational efficiency, the fields are stored in separate blocks on disk and in memory, so that no matter how the language specifies it, there must be a step where the columns are aligned/joined.  This is a common situation in any column oriented database.

nikolay-t:
Thanks for your reply, @aanthony1243. If you say that we receive fields separately from grpc, it makes less sense to add this handling in From() right away. However it still will be good to do it there in From() eventually. So my suggestion is to organize the code in a way that will simplify further calling it from From() - note that this will at least make it not required to do Group to remove _field from grouping.

aanthony1243:
@nikolay-t  no need to remove _field from the grouping. Any columns referenced in colKey will be automatically removed by the pivot function.  it's simple to get all fields aligned. You could write a helper: 

`fromRows = (db) => from(db:db) |> pivot(rowKey: ["_time"], colKey: ["_field"], valueCol:"_value")`

Even simpler, we identified the above pivot pattern and built it in as a helper alignFields: 
`fromRows = (db) => from(db:db) |> alignFields()`

With the new helper, then, you can do: 
`fromRows(db:"telegraf") |> range(start:-1h) |> ...`

nathanielc:
@aanthony1243  Does it make sense to add `fromRows = (db) => from(db:db) |> alignFields()` as a helper function?

aanthony1243:
That was a bit off the cuff design for me there, but if there's consensus on the name + parameters then I think it could make sense to make it a built-in.

query/docs/SPEC.md
```diff
@@ -10,7 +10,7 @@ Any section that is not currently implemented is commented with a IMPL#XXX where
 The Flux language is centered on querying and manipulating time series data.
 
 ### Notation
-
+by its values for
```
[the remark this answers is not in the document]
aanthony1243: strange...fixing now

aanthony1243:
@nikolay-t we decided to work with your feedback and add the `fromRows`  function as a built-in.  Thanks for reaching out!

PR #482, opened 2018-07-26
WIP Flux drop and rename functions

adamperlin:
Adds a base implementation for `drop` and `rename` in flux. See #461 and #462. Currently, only `drop(columns: [...])` and `rename(columns[...])` type calls are supported. Predicate function parameters may soon be added as well (i.e., `rename(fn: (name) => "new_{name}")`)

  - [x] Rebased/mergeable
  - [x] Tests pass


adamperlin:
@aanthony1243 added `fn` parameter support and your refactoring changes. Ready for another review.

query/functions/rename_drop.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+	}
+
+	return &RenameDropProcedureSpec{
+		DropCols:      dropCols,
+		DropPredicate: s.DropPredicate,
+	}, nil
+}
+
+func createRenameDropTransformation(id execute.DatasetID, mode execute.AccumulationMode, spec plan.ProcedureSpec, a execute.Administration) (execute.Transformation, execute.Dataset, error) {
```
nathanielc: Procedures should map 1:1 to transformations. So there should be a RenameDropProcedure. Then the operations can both map to the same procedure.

Then adding a rewrite rule just need to know how to combine multiple RenameDropProcudures into a single procedure.

query/functions/rename_drop.go
```diff
@@ -240,22 +299,44 @@ func (t *renameDropTransformation) Process(id execute.DatasetID, tbl query.Table
...
+		// Cannot have both column list and renameFn; one must be nil
 		if t.renameCols != nil {
 			if newName, ok := t.renameCols[name]; ok {
-				col = query.ColMeta{
-					Label: newName,
-					Type:  c.Type,
-				}
+				col.Label = newName
+			}
+		} else if t.renameFn != nil {
```
aanthony1243: @nathanielc currently renamFn is applied to all columns.  Do you think we should also have a renamePredicate parameter to identify which columns to fetch for renaming?

aanthony1243: Related:  @nathanielc  is it a problem if the user gives two columns the same name?  I mean from more of a functional perspective, will it break the data flow?  will that error surface on its own or should we catch that here?

query/functions/rename_drop_test.go
```diff
@@ -179,3 +379,40 @@ func TestRenameDrop_Process(t *testing.T) {
 		})
 	}
 }
+
+/*
```
aanthony1243: fix or drop this test.

query/functions/testdata/rename_fn.flux
```diff
@@ -0,0 +1,4 @@
+from(db: "test")
+	|> range(start:2018-05-22T19:53:26Z)
+	|> rename(fn: (col) => col)
```
aanthony1243: make an issue and link it to the string interpolation issue so that we update this test when that feature is out.

query/execute/column_fn.go
```diff
@@ -0,0 +1,118 @@
+package execute
```
aanthony1243: I feel like these new functions are not necessary.  They take strings as input and we should already have a way to execute a function on strings.  Maybe look at what Map()   does,  with some extra validation on the function settings during parameter validation.  

If you have a compelling reason why we need these, please do speak up and let us know.

query/functions/testdata/rename.flux
```diff
@@ -0,0 +1,5 @@
+from(db: "test")
```
aanthony1243: These tests look good,  but let's put in some of those edge cases we talked about as tests: 

A)  Given a `rename(columns: {old:new})` what happens if we: 
    -- drop old before the rename.  
   -- drop old after
   -- drop new before/after

b) chain multiple renamings 

c)  drop a column that's needed for some operation (e.g. filter), but not afterward.  

d)  drop a column that's never used.

adamperlin:
@aanthony1243 `keep` is implemented, should be ready for another look

query/functions/rename_drop.go
```diff
@@ -299,45 +424,89 @@ func (t *renameDropTransformation) Process(id execute.DatasetID, tbl query.Table
 	// builder and the table - we need to keep track
 	colMap := make([]int, builder.NCols())
 	// TODO: error if overlap between dropCols and renameCols?
-	if t.dropPredicate != nil {
-		if err := t.dropPredicate.Prepare(); err != nil {
-			return err
+
+	if t.dropKeepCols != nil && t.renameCols != nil {
+		for k := range t.renameCols {
+			if _, ok := t.dropKeepCols[k]; ok {
```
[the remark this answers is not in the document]
aanthony1243: I think to clean up the design, this function should have 2 helper functions: 

1.  keepToDropCols()   that converts a keep set to a drop set.  
2.  shouldDropCol(fn, bool)  that returns true if the column should be dropped.  

then, at the very top of `process` you can have 
```if t.keepSpecified {
     t.keepToDropCols()
}```

then later, instead of calling the predicate directly, you can use t.shouldDropCol()  which will call the predicate and then return  the correct truth value depending on t.keepSpecified.  

doing this eliminates a lot of the if/else you have going on here, and makes `process` easier to follow and maintain.

File: query/functions/rename_drop.go [hunk 8]
```diff
@@ -299,45 +424,89 @@ func (t *renameDropTransformation) Process(id execute.DatasetID, tbl query.Table
...
+				}
+				return fmt.Errorf(`drop error: column "%s" doesn't exist`, c)
+			}
+		}
+	}
+
+	// If `keepSpecified` is true, i.e., we want to exclusively keep the columns listed in dropCols
+	// as opposed to exclusively dropping them, we update dropCols to be the list of all columns to drop
+	// to simplify further logic.
+	if t.keepSpecified && t.dropKeepCols != nil {
```
aanthony1243: this looks like what should go in a helper function.  can call it here or at the start of `process`

query/functions/query_test.go
```diff
@@ -50,6 +50,10 @@ var skipTests = map[string]string{
 	"null_as_value":             "null not supported as value in influxql (https://github.com/influxdata/platform/issues/353)",
 	"difference_panic":          "difference() panics when no table is supplied",
 	"string_interp":             "string interpolation not working as expected in flux (https://github.com/influxdata/platform/issues/404)",
+	"drop_before_rename":        `error: rename error: column "old"  doesn't exist`,
```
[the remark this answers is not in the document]
adamperlin: Currently we don't have any "expected error" functionality. We could add it though.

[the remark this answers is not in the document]
aanthony1243: whatever text is returned as a response, put that in the out.csv file.  if it represents correct behavior, even if that behavior is to print an error message, then it's good to go.

query/functions/rename_drop.go
```diff
@@ -0,0 +1,578 @@
...
+	}
+
+	return nil
+}
+
+func (t *renameDropTransformation) shouldDrop(col string) (bool, error) {
+	t.dropKeepScope[t.dropKeepColParam] = values.NewStringValue(col)
+	if shouldDrop, err := t.dropKeepPredicate.EvalBool(t.dropKeepScope); err != nil {
+		return false, err
+	} else if shouldDrop != t.keepSpecified {
```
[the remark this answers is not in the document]
adamperlin: sure.

aanthony1243:
The new tests are causing failures as we discussed on a call yesterday.  Disable them and create an issue to fix the testing code.  We'll do that after @nathanielc pushes the big update to the query service.

adamperlin:
Ok. There is still a bit more work to be done here: As @nathanielc pointed out in #462, `drop`, `rename`, and `keep` must modify the group key in order to work as expected. I've implemented that functionality, now just writing a few more tests.

adamperlin:
@aanthony1243 added group key support. Could you give another look over?

query/functions/rename_drop.go
```diff
@@ -0,0 +1,597 @@
...
+
+type RenameDropProcedureSpec struct {
+	RenameCols map[string]string
+	RenameFn   *semantic.FunctionExpression
+	// The same field is used for both columns to drop and columns to keep
+	DropKeepCols map[string]bool
+	// the same field is used for the drop predicate and the keep predicate
+	DropKeepPredicate *semantic.FunctionExpression
+	// Denotes whether we're going to do a drop or a keep
+	KeepSpecified bool
```
nathanielc: I think this kind of structure makes following the logic complicated. It also means that it is impossible to merge multiple drop/rename/keep operations into a single operation.

What if we instead called these kinds of operations `schema mutations` and a mutation is a rename, drop or keep operation on the table schema. Then this RenameDropProcedureSpec could be called SchemaMutationProcedureSpec and it would have a list of Mutations. 

Then to combine multiple SchemaMutationProcedureSpec all that needs to be done is append the new list of mutations.

The mutations would be an interface that can mutate a table's column and group key. Then the transformation applies each mutation in order.

Thoughts? Its quite a different design than what exists here, but it sets us up for good planning of this kind of procedure. Even if we do not do the planning bits right now.  It also keeps the mutations separate so there code is easy to follow without having to create unique procedures and transformations for each.

 This design is easy to add to as well. For example the need for a duplicate column operation came up today. It would be added as a new Mutation without having to change any other code.

adamperlin: I agree that this would be much more maintainable, efficient, and extensible than what's here currently. 

I had initially thought that we would want rewrite rules to combine multiple rename, drop, and keep operations into one type of procedure spec, hence the `RenameDropProcedureSpec`. However, the order of application of drop, rename, and keep procedures matters (a fact which did not go into the initial design), preventing the type of re-write rules that I thought we could use.  So this current design isn't even all that useful, in addition to being hard to follow. 

It makes a lot more sense to apply these types of schema-mutating transformations in sequence and to have a higher level procedure spec to do it.  I like that this then allows for relatively painless procedure spec rewrites, since adding schema mutations together becomes so simple. 

The mutation interface would not have to be very complicated. A mutation method could have a signature similar or identical to `Process`.  An idea would be: 
```go
type SchemaMutation interface {
     // Changes schema of table, updating group key if necessary 
     MutateSchema(id execute.DatasetID, tbl query.Table) error
}
```
Leading to a structure like the following in implementation:
```go
type SchemaMutationTransformation struct {
    ... 
    Mutations []SchemaMutation
}
```
Much of the core logic in the current code could be wrapped in `SchemaMutation` implementations relatively easily, so hopefully that wouldn't have to be scrapped.

nathanielc: > Much of the core logic in the current code could be wrapped in SchemaMutation implementations relatively easily, so hopefully that wouldn't have to be scrapped.

Yeah, this was my thought as well. The core logic remains the same, we just organize it a bit differently.

Also just some Go idioms to follow. Typically single method interfaces are named with an `er` or `or` like `io.Writer` and `io.Closer` that have `Write` and `Close` methods respectively.

In this case we could use `SchemaMutator` with a method named `Mutate`. Does the SchemaMutator need the `id`? If not lets leave it off. 

This looks like a good design.

adamperlin: Ok, I'll start implementing these interfaces.

aanthony1243: this all sounds good to me.  We did discuss this previously, that we should be collecting tuples of rename/drop lists so that ordering could be preserved.  Not sure why we didn't make the change then, but it's good that we're getting to it now.

adamperlin:
@aanthony1243 @nathanielc I've implemented a first version of the refactor - let me know what you think.

query/functions/rename_drop.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+	ret := make(map[string]bool, len(arr))
+	for _, s := range arr {
+		ret[s] = true
+	}
+	return ret
+}
+
+func NewDropMutator(qs query.OperationSpec) (*DropKeepMutator, error) {
+	m := &DropKeepMutator{}
```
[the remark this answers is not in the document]
adamperlin: funny you mention this - I actually had it as `NewDropKeepMutator` but wondered if this way would be clearer. I guess if the implementations are combined, the constructors probably should be.

adamperlin:
@aanthony1243 @nathanielc This could probably use another look over. I've reorganized the file structure here. `schema_mutators.go` contains all of the `SchemaMutator` specific definitions and implementations. All new implementations of `SchemaMutator` should go in this file. `schema_functions.go` contains all the `OpSpec`s for schema mutation functions, in addition to the `SchemaMutationProcedureSpec`, `schemaMutationTransformation` and `Process` code for performing mutations.  I was thinking all schema mutation related code would live in just those two files. Giving each mutation its own file doesn't seem idiomatic since mutations don't have their own procedures and transformations like other flux functions.

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+	newMutations := make([]SchemaMutator, len(s.Mutations))
+	for _, m := range newMutations {
+		newMutations = append(newMutations, m.Copy())
+	}
+	return &SchemaMutationProcedureSpec{
+		Mutations: newMutations,
+	}
+}
+
+func newFunc(fn *semantic.FunctionExpression, types [2]semantic.Type) (compiler.Func, string, error) {
```
nathanielc: I am having a hard time understanding what this function does based on its signature and name. Maybe it should be called compilePredicateFn? What are the `types` for?

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+
+	return nil
+}
+
+func (m *RenameMutator) Copy() SchemaMutator {
+	nm := new(RenameMutator)
+	nm.RenameCols = m.RenameCols
+	nm.RenameFn = m.RenameFn
+	return nm
+}
```
nathanielc: These Copy methods are tricky, why do Scope and ParamName not need to be copied? Also the RenameCols and RenameFn need to be deep copied. The RenameFn should have a Copy method, and RenameCols is a map that needs to be copied.

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+
+type DropKeepMutator struct {
+	KeepCols          map[string]bool
+	DropCols          map[string]bool
```
nathanielc: I think how you had this before with a single column map makes sense here since they behave the same and the FlipPredicate flag takes care of the rest. That should help to reduce duplicated code in the methods here.

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+
+func (b *BuilderContext) UpdateKey(newKey query.GroupKey) {
+	b.TableKey = newKey
+}
+
+func (b *BuilderContext) UpdateColMap(newMap []int) {
+	b.ColIdxMap = newMap
+}
+
+func (b *BuilderContext) Copy() *BuilderContext {
```
nathanielc: Why does the BuilderContext need to be copied?

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+	return nil
+}
+
+func (m *RenameMutator) Mutate(ctx *BuilderContext) error {
+
+	if err := m.checkColumnReferences(ctx.Cols()); err != nil {
+		return err
+	}
+
+	keyCols := make([]query.ColMeta, 0, len(ctx.Cols()))
```
nathanielc: Why do we need to make a copy of everything in the context in each of these Mutate methods?

For rename it should be able to just apply the rename inplace? 

Something useful here is that in Go the `for _,c := range ctx.Cols()` will use the same memory space for `c` in each iteration of the loop, which means that `c` is a copy. If you want to mutate the value of the slice in place do:

```
for i := range ctx.TableCols {
 ...
  m.renameCol(&ctx.TableCols[i]) // pass a pointer to the value within the slice.
 ...
}
```

query/functions/schema_mutators.go
adamperlin: `types` are the types that should be enforced for parameter and return. It's used for compiling functions with different return types, so the type checking couldn't be hardcoded.

query/functions/schema_mutators.go
adamperlin: We could just modify the slice in place by reference for rename, that's true. Group key columns need to be copied because to my knowledge you can't modify a group key in place without breaking interface barriers, so we must create a new one each time. If we're creating a new group key each time, group key values must be copied as well. The column list and map must be copied in drop/keep because elements might be removed, which is easier to do using a new slice. I would say most of these copies are necessary, this being an exception.

query/functions/schema_mutators.go
adamperlin: We actually need two slices here. I preferred what I had initially, but the reason I changed this is that it fixes a serious bug. If `Mutate` is called more than once on a `DropKeepMutator`, and it's a keep operation, we need to be able to maintain both an original copy of the columns we want to keep and a temporary list of all other columns to drop. We can only re-use the same list for this if we can guarantee `Mutate` will be called only once.

query/functions/schema_mutators.go
adamperlin: I'll remove this

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+func (s *SchemaMutationProcedureSpec) Kind() plan.ProcedureKind {
+	return SchemaMutationKind
+}
+
+func (s *SchemaMutationProcedureSpec) Copy() plan.ProcedureSpec {
+	newMutations := make([]SchemaMutator, len(s.Mutations))
+	for _, m := range newMutations {
+		newMutations = append(newMutations, m.Copy())
```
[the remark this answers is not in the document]
adamperlin: Thanks. I suppose the above would also work if the slice was declared with 0 initial size but a defined capacity: `newMutuations := make([]SchemaMutator, 0, len(s.Mutations))`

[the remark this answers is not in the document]
adamperlin: but doesn't matter because I'm going to remove this function

query/functions/drop_rename_keep.go
```diff
@@ -0,0 +1,452 @@
...
+
+	return &SchemaMutationProcedureSpec{
+		Mutations: newMutations,
+	}
+}
+
+func newSchemaMutationProcedure(qs query.OperationSpec, pa plan.Administration) (plan.ProcedureSpec, error) {
+	s, ok := qs.(SchemaMutation)
+	if !ok {
+		return nil, fmt.Errorf("invalid spec type %T doesn't implement SchemaMutationProcedureSpec", qs)
```
nathanielc: This should read `doesn't implement SchemaMutation`.

query/functions/drop_rename_keep.go
```diff
@@ -0,0 +1,452 @@
...
+
+	query.RegisterFunction(string(m.Kind), m.Create, signature)
+	query.RegisterOpSpec(m.Kind, m.New)
+
+	// Add to list of SchemaMutations which should map to a
+	// SchemaMutationProcedureSpec
+	SchemaMutationOps = append(SchemaMutationOps, m.Kind)
+}
+
+// A list of all MutationOps to register.
```
nathanielc: Can you remove this comment to use the MutationRegistrar name? Also remove the empty line so it gets associated with the list?

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,335 @@
...
+
+type SchemaMutation interface {
+	Mutator() (SchemaMutator, error)
+	Copy() SchemaMutation
+}
+
+// Utility function for compiling an `fn` parameter for rename or drop/keep. In addition
+// to the function expression, it takes an array of two types to verify the result against:
+// a single argument type, and a single return type.
+func compileFnParam(fn *semantic.FunctionExpression, types [2]semantic.Type) (compiler.Func, string, error) {
```
nathanielc: I think this function would read better if it named the two type variables, maybe `inType` and `outType`, instead of using an array of two types. Also since the implementation doesn't need to iterate over the array this should be an easy change.

adamperlin:
Done.

query/functions/schema_mutators.go
```diff
@@ -0,0 +1,335 @@
...
+
+type SchemaMutation interface {
+	Mutator() (SchemaMutator, error)
+	Copy() SchemaMutation
+}
+
+// Utility function for compiling an `fn` parameter for rename or drop/keep. In addition
+// to the function expression, it takes an array of two types to verify the result against:
+// a single argument type, and a single return type.
+func compileFnParam(fn *semantic.FunctionExpression, paramType semantic.Type, returnType semantic.Type) (compiler.Func, string, error) {
```
nathanielc: This looks great! You do not need to change it but in the future know that Go can list multiple params with the same type without repeating the type. Like this:
```
func compileFnParam(fn *semantic.FunctionExpression, paramType , returnType semantic.Type) (compiler.Func, string, error) 
```

PR #483, opened 2018-07-26
Implement CRUD for sources

desa:


desa:
will need more changes. closing until it is ready

desa:
will need more changes

PR #502, opened 2018-07-27
feat: do not require the makefile for go generate

jsternberg:
There is now a gorunpkg.go script that will perform the simple tasks of
compiling a module and running it similar to how `go run` works for a
file. This script can be run and used for generating files so `go
generate ./...` works properly without installing any dependencies
manually.

### DIFF
--- a/task/backend/pb/pb.go
+++ b/task/backend/pb/pb.go
@@ -4,4 +4,4 @@ package pb
 
 // The tooling needed to correctly run go generate is managed by the Makefile.
 // Run `make` from the project root to ensure these generate commands execute correctly.
-//go:generate protoc -I ../../../vendor -I . --plugin ../../../bin/${GOOS}/protoc-gen-gogofaster --gogofaster_out=plugins=grpc:. ./tasks.proto
+//go:generate protoc -I ../../../vendor -I . --plugin ../../../scripts/protoc-gen-gogofaster --gogofaster_out=plugins=grpc:. ./tasks.proto
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -208,14 +208,37 @@
   version = "v1.38.1"
 
 [[projects]]
-  digest = "1:32e6bf93f32b6562d44dcdae34fd11a819224db80627bbc4ee108bfc01f81922"
+  digest = "1:f211070279a54483ab7923376cebb4c024e84aeee17a989537e054413a0f88f4"
   name = "github.com/gogo/protobuf"
   packages = [
     "gogoproto",
+    "plugin/compare",
+    "plugin/defaultcheck",
+    "plugin/description",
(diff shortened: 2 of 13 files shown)


task/backend/pb/pb.go
```diff
@@ -4,4 +4,4 @@ package pb
 
 // The tooling needed to correctly run go generate is managed by the Makefile.
 // Run `make` from the project root to ensure these generate commands execute correctly.
-//go:generate protoc -I ../../../vendor -I . --plugin ../../../bin/${GOOS}/protoc-gen-gogofaster --gogofaster_out=plugins=grpc:. ./tasks.proto
+//go:generate protoc -I ../../../vendor -I . --plugin ../../../scripts/protoc-gen-gogofaster --gogofaster_out=plugins=grpc:. ./tasks.proto
```
nathanielc: This is the tricky case since we cannot rely on the PATH to find the plugin. Maybe protoc has an ENV for where it searches for plugins we can leverage?

Also notice that `protoc` is required by the go generate command and is not installed by the Go toolchain.

I do not see value in enforcing that all build tooling be part of the Go ecosystem. In fact go generate exists exactly for this purpose, allow for the generation step to be encapsulated in the Go build process, while allowing it to call out to other tooling. 

All the makefile does is encode the dependencies of generated code and tooling.

jsternberg: I don't think I said all tooling needs to be part of the Go ecosystem, but we don't vendor `protoc` or its sources and build them from our vendor directory so I think that these are different situations. `protoc` is also a much more common build tool and I think it's perfectly acceptable to have that as an external dependency.

But the trick that is used her lets us use the vendored generator which is go code and which is something we compile. At the moment, the generate uses a binary that is produced by the makefile which means, effectively, you have to use the makefile to generate the file. The `go generate` command doesn't really work without intervention from another tool.

So I disagree that the makefile is just encoding dependencies because the makefile is also performing building tasks. If we insist on using the makefile, we should probably encode the generate functionality in the makefile instead so we don't have a mixture of the two different methods. That way, doing something like:

```bash
$ git clean -xdf
$ go generate -x ./...
```

At the moment, the above will fail because we have encoded the creation of the binaries into the makefile.

nathanielc: > At the moment, the above will fail because we have encoded the creation of the binaries into the makefile.

This is true, but teaching the makefile how to build them does not mean you cannot build them yourself externally.

With the exception of the protoc plugins if you run those commands, it will work, if you first install the build tools manually. 

So, the makefile is a helper to install the build tools for you. That is all.

jsternberg:
Not relevant anymore. We're now also using `go run` which solves the conflict between Makefile vs compiling within `go generate` by having `go run` just do all the work for us.

PR #505, opened 2018-07-27
Update the vairous QueryService to be consistent

nathanielc:

Update each of the QueryService to consume a query.Request.

_What was the problem?_

There were three different query services. Each had a unique and different API that did not clearly communicate why you would use one over the other.

_What was the solution?_

The solution was three fold:

* Move all query services into platform
* Update all service to use the query.Request and query.ProxyRequest types
* Make use of the Compiler interface in the implementation of all services.

As a result of these changes the differences between queryd and transpilerd vanished. As such transpilerd has been removed from platform.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] swagger.json updated 
  - [x] Update README for changes


nathanielc:
@pauldix @nhaugo The last commit updates the swagger file for the 1.x and 2.x query endpoints.

http/swagger.yml
```diff
@@ -972,80 +963,114 @@ paths:
...
   schemas:
-    Query:
+    QueryRequest:
       description: query influx with specified return formatting.
       type: object
       properties:
-        query:
-          description: query script to execute
+        authorization:
+          $ref: "#/components/schemas/Authorization"
```
goller: This is incorrect.  Authorization is a header and not a property of the body schema.

http/swagger.yml
```diff
@@ -972,80 +963,114 @@ paths:
...
+        organization_id:
```
goller: We won't have organization_id as a property of the schema.  The paths are changing to /v2/org/{id}/query

http/swagger.yml
```diff
@@ -972,80 +963,114 @@ paths:
...
+          description: ID of organization for the query
           type: string
-        type:
-          description: type of query
+        complier_type:
```
goller: I would rather type be in the object (`FluxCompiler`, `SpecCompiler`, `InfluxQLCompiler`) and not a different key.  an object should be self-descriptive rather than rely on a key on the outer level.

http/swagger.yml
```diff
@@ -972,80 +963,114 @@ paths:
...
+          description: type of the compiler to use for the query
           type: string
           default: flux
           enum:
             - flux
+            - spec
+            - influxql
+        compiler:
```
goller: Should have a default

http/swagger.yml
```diff
@@ -972,80 +963,114 @@ paths:
...
+          oneOf:
+            - $ref: "#/components/schemas/FluxCompiler"
+            - $ref: "#/components/schemas/SpecCompiler"
+            - $ref: "#/components/schemas/InfluxQLCompiler"
+        dialect_type:
```
goller: Same comment here... `type` should be in the specific `CSVDialect` and `InfluxQLDialect` themselves.

http/swagger.yml
```diff
@@ -252,6 +248,8 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/InfluxqlQueryError"
+  # 2.x query endpoint
+  /query:
```
pauldix: Is this the path? The v2 query path should be `/organizations/:<id|name>/buckets/:<id|name>/query`

pauldix: err, put `/api/v1` before orgs..

goller: @pauldix query will be under org rather than bucket because queries can be multi bucket

pauldix: ah yes, you're right

nathanielc:
> I'd like to see a different flatter schema.

I agree. Internally,  between gateway and transpiler we are using this deep schema. Would you like to flatten the schema for both the external API and internal APIs or would you like to see the external API be a wrapper around the internal one?

Also this change is not primarily about changing the external API, its just that the current internal API is changing and it was being exposed directly. If we want to decouple the internal vs external APIs then I can make this change in isolation and ignore the changes to the external API. We will need to write the wrapper bits at some point though.

goller:
I think this should be restrained to the internal api for now.  Internal api should be free to change rapidly underneath the external api

http/swagger.yml
desa: This might not be the place to talk about this, but how should this work with sources? Specifically what values do we use for organization ID or name when talking to `v1` sources?

`/api/sources/:id/organizations/:<id|name>/query`

Maybe just use an organization called `default` or `autogen` for any `v1` sources?

http/query_service.go
```diff
@@ -169,112 +125,73 @@ type QueryService struct {
...
-
 	var decoder query.MultiResultDecoder
 	switch resp.Header.Get("Content-Type") {
 	case "text/csv":
 		fallthrough
 	default:
 		decoder = csv.NewMultiResultDecoder(csv.ResultDecoderConfig{})
 	}
-	result, err := decoder.Decode(resp.Body)
+	results, err := decoder.Decode(resp.Body)
```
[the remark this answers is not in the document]
nathanielc: No, the caller needs to defer because this method does not read the results.

http/query_service.go
```diff
@@ -169,112 +125,73 @@ type QueryService struct {
 	InsecureSkipVerify bool
 }
 
-func (s *QueryService) Query(ctx context.Context, orgID platform.ID, query *query.Spec) (query.ResultIterator, error) {
+func (s *QueryService) Query(ctx context.Context, req *query.Request) (query.ResultIterator, error) {
```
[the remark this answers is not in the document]
nathanielc: This code is essentially the client code for the QueryService over HTTP, I believe the task code consumes this implementation.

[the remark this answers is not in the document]
aanthony1243: ok, that's new to me, but it's been a while since I checked.  just thought it worth checking.

http/query_service.go
```diff
@@ -169,112 +125,73 @@ type QueryService struct {
...
 		return nil, err
 	}
-	values := url.Values{}
-	values.Set("orgID", orgID.String())
-	u.RawQuery = values.Encode()
-
-	var buf bytes.Buffer
-	if err := json.NewEncoder(&buf).Encode(query); err != nil {
+	var body bytes.Buffer
+	if err := json.NewEncoder(&body).Encode(req); err != nil {
```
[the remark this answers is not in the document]
aanthony1243: @nathanielc  is this an issue?  other than this I'm 👍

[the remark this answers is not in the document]
nathanielc: Thanks, fixed

PR #512, opened 2018-07-30
owner mapping -> user resource mapping

jademcgough:
Change owner mapping to user resource mapping.

This will allow us to use the same service for mapping users to their resources, whether they are owners or members.


desa:
Looks good, I like the name a lot better than Owner as well. Looks like there are a few failing tests in the tasks package related to owners
```
# github.com/influxdata/platform/task
task/platform_adapter.go:68:50: t.Owner undefined (type *platform.Task has no field or method Owner)
task/platform_adapter.go:97:8: unknown field 'Owner' in struct literal of type platform.Task
task/platform_adapter.go:146:8: unknown field 'Owner' in struct literal of type platform.Task
```

PR #514, opened 2018-07-31
feat(tasks): add get meta to the store interface.

lyondhill:
Grant access to backend task meta which should allow us to pick up where we left off in the event of a crash.

  - [ ] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/task/backend/bolt/bolt.go
+++ b/task/backend/bolt/bolt.go
@@ -357,7 +357,7 @@ func (s *Store) FindTaskMetaByID(ctx context.Context, id platform.ID) (*pb.Store
 		b := tx.Bucket(s.bucket)
 		stmBytes = b.Bucket(taskMetaPath).Get(paddedID)
 		if stmBytes == nil {
-			return ErrNotFound
+			return errors.New("task meta not found")
 		}
 		return nil
 	})
@@ -371,7 +371,7 @@ func (s *Store) FindTaskMetaByID(ctx context.Context, id platform.ID) (*pb.Store
 		return nil, err
 	}
 
-	return &stm, err
+	return &stm, nil
 }
 
 // DeleteTask deletes the task
--- a/task/backend/bolt/bolt.go
+++ b/task/backend/bolt/bolt.go
@@ -158,7 +158,7 @@ func (s *Store) CreateTask(ctx context.Context, org, user platform.ID, script st
 
 		// metadata
 		stm := pb.StoredTaskInternalMeta{
-			MaxConcurrency: 1,
+			MaxConcurrency: int32(o.Concurrency),
 		}
 
 		stmBytes, err := stm.Marshal()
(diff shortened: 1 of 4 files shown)


task/backend/bolt/bolt.go
```diff
@@ -350,6 +350,30 @@ func (s *Store) FindTaskByID(ctx context.Context, id platform.ID) (*backend.Stor
...
+		return nil, err
+	}
+
+	stm := pb.StoredTaskInternalMeta{}
+	err = stm.Unmarshal(stmBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &stm, err
```
mark-rushakoff: We know err is nil here, so may as well be `return &stm, nil`.

task/backend/inmem_store.go
```diff
@@ -57,7 +57,7 @@ func (s *inmem) CreateTask(_ context.Context, org, user platform.ID, script stri
 
 	s.mu.Lock()
 	s.tasks = append(s.tasks, task)
-	s.runners[id.String()] = pb.StoredTaskInternalMeta{MaxConcurrency: 1} // TODO:  make settable in the create
+	s.runners[id.String()] = pb.StoredTaskInternalMeta{MaxConcurrency: int32(o.Concurrency)} // TODO:  make settable in the create
```
mark-rushakoff: TODO can be removed now?

task/backend/inmem_store.go
```diff
@@ -147,6 +147,14 @@ func (s *inmem) FindTaskByID(_ context.Context, id platform.ID) (*StoreTask, err
 
 	return nil, nil
 }
+func (s *inmem) FindTaskMetaByID(ctx context.Context, id platform.ID) (*pb.StoredTaskInternalMeta, error) {
+	meta, ok := s.runners[id.String()]
+	if !ok {
+		return nil, errors.New("task meta not found")
```
mark-rushakoff: This error isn't the same as the one the bolt store returns, so I'm wondering if they should be identical.

task/backend/storetest/storetest.go
```diff
@@ -7,39 +7,51 @@ import (
...
 // NewStoreTest creates a test function for a given store.
-func NewStoreTest(name string, cf CreateStoreFunc, df DestroyStoreFunc) func(t *testing.T) {
+func NewStoreTest(name string, cf CreateStoreFunc, df DestroyStoreFunc, funcNames ...string) func(t *testing.T) {
+	if len(funcNames) == 0 {
+		funcNames = []string{
+			"CreateTask",
+			"ModifyTask",
+			"ListTasks",
+			"FindTask",
+			"findMeta",
```
mark-rushakoff: Uppercase `FindMeta` for consistency here?

task/backend/storetest/storetest.go
```diff
@@ -361,6 +373,65 @@ from(db:"test") |> range(start:-1h)`
...
+	err = s.FinishRun(context.Background(), id, qr.RunID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	meta, err = s.FindTaskMetaByID(context.Background(), id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
mark-rushakoff: Remove blank line.

task/backend/storetest/storetest.go
```diff
@@ -361,6 +373,65 @@ from(db:"test") |> range(start:-1h)`
...
+	if meta != nil {
+		t.Fatalf("expected nil meta when finding nonexistent ID, got %#v", meta)
+	}
+
+	qr, err := s.CreateRun(context.Background(), id, time.Unix(1, 0).Unix())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = s.CreateRun(context.Background(), id, time.Unix(2, 0).Unix())
```
mark-rushakoff: I think this and the one above can be just `time.Unix`? Maybe would need to be `time.Unix(...).UTC()` if the time zone is an issue.

PR #515, opened 2018-07-31
Delete users and orgs

docmerlin:
_What was the problem?_
 Tasks for entire users and orgs were not deletable in the tasks backend
_What was the solution?_
Make a method for deleting users and one for deleting orgs in the backend.Store interface

### DIFF
--- a/task/backend/bolt/bolt.go
+++ b/task/backend/bolt/bolt.go
@@ -523,3 +523,119 @@ func padID(id platform.ID) platform.ID {
 	copy(buf[len(buf)-len(id):], id)
 	return buf[:]
 }
+
+// DeleteUser syncronously deletes a user and all their tasks from a bolt store.
+func (s *Store) DeleteUser(ctx context.Context, id platform.ID) error {
+	userID := padID(id)
+	err := s.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket(s.bucket)
+		ub := b.Bucket(usersPath).Bucket(userID)
+		if ub == nil {
+			return backend.ErrUserNotFound
+		}
+		c := ub.Cursor()
+		i := 0
+		for k, _ := c.First(); k != nil; k, _ = c.Next() {
+			i++
+			// check for cancelation every 256 tasks deleted
+			if i&0xFF == 0 {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+				}
+			}
+			if err := b.Bucket(tasksPath).Delete(k); err != nil {
+				return err
+			}
+			if err := b.Bucket(taskMetaPath).Delete(k); err != nil {
+				return err
+			}
+			if err := b.Bucket(orgByTaskID).Delete(k); err != nil {
+				return err
+			}
+			if err := b.Bucket(userByTaskID).Delete(k); err != nil {
+				return err
+			}
+			if err := b.Bucket(nameByTaskID).Delete(k); err != nil {
+				return err
+			}
+
+			org := b.Bucket(orgByTaskID).Get(k)
+			if len(org) > 0 {
+				ob := b.Bucket(orgsPath).Bucket(org)
+				if ob != nil {
+					if err := ob.Delete(k); err != nil {
+						return err
+					}
+				}
+			}
+		}
+
+		// check for cancelation one last time before we return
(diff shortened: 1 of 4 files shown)


task/backend/inmem_store.go
```diff
@@ -236,3 +236,54 @@ func (s *inmem) FinishRun(ctx context.Context, taskID, runID platform.ID) error
...
+func getOrg(st StoreTask) platform.ID {
+	return st.Org
+}
+
+// DeleteOrg syncronously deletes an org and all their tasks from a bolt store.
+func (s *inmem) DeleteOrg(ctx context.Context, id platform.ID) error {
+	return s.delete(ctx, id, getOrg)
+}
+
+// DeleteUser syncronously deletes a user and all their tasks from a bolt store.
```
mark-rushakoff: Should be "from an in-mem store." And synchronously has an `h` in it.

task/backend/store.go
```diff
@@ -38,6 +45,12 @@ type Store interface {
 	// FinishRun removes runID from the list of running tasks and if its `now` is later then last completed update it.
 	FinishRun(ctx context.Context, taskID, runID platform.ID) error
 
+	// DeleteOrg if the org does not exist, nil is returned
```
mark-rushakoff: Godoc should be complete sentences, with a period. https://blog.golang.org/godoc-documenting-go-code

Please fix this sentence and add punctuation to the other godoc introduced in this file.

task/backend/storetest/storetest.go
```diff
@@ -469,3 +474,86 @@ from(db:"test") |> range(start:-1h)`
...
+
+func testStoreDeleteUser(t *testing.T, create CreateStoreFunc, destroy DestroyStoreFunc) {
+	s := create(t)
+	defer destroy(t, s)
+	ids := createABunchOFTasks(t, s,
+		func(u, _ uint64) bool {
+			if u == 0 {
+				return true
+			}
+			return false
```
mark-rushakoff: This can be simplified to `return u == 0`.

task/backend/storetest/storetest.go
```diff
@@ -469,3 +474,86 @@ from(db:"test") |> range(start:-1h)`
...
+
+func testStoreDeleteOrg(t *testing.T, create CreateStoreFunc, destroy DestroyStoreFunc) {
+	s := create(t)
+	defer destroy(t, s)
+	ids := createABunchOFTasks(t, s,
+		func(_, o uint64) bool {
+			if o == 0 {
+				return true
+			}
+			return false
```
mark-rushakoff: This can be simplified to `return o == 0`.

task/backend/storetest/storetest.go
```diff
@@ -469,3 +474,86 @@ from(db:"test") |> range(start:-1h)`
...
+		},
+	)
+	org := make(platform.ID, 8)
+	err := s.DeleteOrg(context.TODO(), org)
+	if err != nil {
+		t.Fatal(err)
+	}
+	for i := range ids {
+		task, err := s.FindTaskByID(context.TODO(), ids[i])
```
mark-rushakoff: The godoc for `context` says to use context.Background for tests: https://golang.org/pkg/context/#Background

task/backend/bolt/bolt.go
```diff
@@ -499,3 +499,119 @@ func padID(id platform.ID) platform.ID {
...
+		ub := b.Bucket(usersPath).Bucket(userID)
+		if ub == nil {
+			return backend.ErrUserNotFound
+		}
+		c := ub.Cursor()
+		i := 0
+		for k, _ := c.First(); k != nil; k, _ = c.Next() {
+			i++
+			// check for cancelation every 128 tasks deleted
+			if i&(256-1) == 0 {
```
mark-rushakoff: The comment says 128, but the value differs. IMO `0xff` is slightly clearer than `256-1` for a bitwise operation, but I'm fine with it as-is.

task/backend/storetest/storetest.go
docmerlin: doh!

task/backend/storetest/storetest.go
docmerlin: Again, doh

PR #516, opened 2018-07-31
fix: Properly check aggregate type

nathanielc:


query/execute/aggregate.go
```diff
@@ -133,7 +133,8 @@ func (t *aggregateTransformation) Process(id DatasetID, tbl query.Table) error {
 			vf = t.agg.NewFloatAgg()
 		case query.TString:
 			vf = t.agg.NewStringAgg()
-		default:
+		}
+		if vf == nil {
```
jsternberg: It seems to me that the error should be returned in the default for the switch statement. It seems more direct to me.

PR #518, opened 2018-07-31
fix(query): fix passing down the bucket id in the from call

jsternberg:
The bucket id was not being copied when the from procedure was being
created so it never showed up and the bucket id was not set.

In addition, the range spec generated by the transpiler did not include
the default values so this adds those default values for those calls.


jsternberg:
Removed and put into a separate PR.

PR #519, opened 2018-07-31
fix: Explicilty handle overflow of currentTime is storage source

nathanielc:


query/functions/storage/storage.go
```diff
@@ -136,10 +137,15 @@ func (s *source) next(ctx context.Context, trace map[string]string) (query.Table
...
 
-	s.currentTime = s.currentTime + execute.Time(s.window.Every)
-	if stop > s.bounds.Stop {
+	if stop > s.bounds.Stop || s.overflow {
 		return nil, 0, false
 	}
+
+	currentTime := s.currentTime + execute.Time(s.window.Every)
+	// Check if we had an overflow, we are done after this pass
+	s.overflow = s.currentTime < currentTime
```
jsternberg: To my knowledge, the better way to handle is to check if you would overflow before you perform the addition. I think we need to come back to this and look at how influxql handles overflow in its windowing and how it mitigates the overflow problem.

I think it's fine for now just to get the error handled, but can you put a TODO that we need to revisit this?

PR #524, opened 2018-07-31
Specify how group key is handled in rename/drop spec

adamperlin:

Update to #462 -  added explanations around how `rename`, `drop`, and `keep` handle group keys. 

_What was the problem?_
The previous spec did not specify how group keys were handled.
_What was the solution?_
Added additional explanation about `rename`, `drop` and `keep` group key manipulation.

  - [x] Rebased/mergeable
  - [x] Tests pass


query/docs/SPEC.md
```diff
@@ -1105,8 +1106,9 @@ Rename all columns with `fn` parameter: `rename(fn: (col) => "{col}_new")`
 
 #### Drop 
 
-Drop will exclude specified columns from a returned data table. Columns to exclude can be specified either through a 
-list, or a predicate function. 
+Drop will exclude specified columns from a table. Columns to exclude can be specified either through a 
+list, or a predicate function. In the case that columns are both specified in a call to `drop` and part of the 
+group key, the group key will be updated to exclude those columns.
```
nathanielc: I think it is simpler to say that: `when a dropped column is part of the group key it will also be dropped from the key.`

This means that these functions can merge groups depending on the data contained in the dropped column.

query/docs/SPEC.md
```diff
@@ -1086,9 +1086,10 @@ Range has the following properties:
 
 #### Rename 
 
-Rename will rename specified columns in a data table. 
+Rename will rename specified columns in a table. 
 There are two variants: one which takes a map of old column names to new column names,
- and one which takes a mapping function.
+and one which takes a mapping function. If a column is renamed and is part of the group key,
```
nathanielc: The md format places lines in the same paragraph so long as an empty line does not separate them. In other words you can place this sentence on its own line and if will be part of the same paragraph, but it will have its own source line, which makes it easier to diff.

query/docs/SPEC.md
```diff
@@ -1123,8 +1125,9 @@ Drop all columns matching a predicate: `drop(fn: (col) => col =~ /usage*/)`
 
 #### Keep 
 
-Keep can be thought of as the inverse of drop. It will return a table containing only columns that are specified,
-ignoring all others.
+Keep is the inverse of drop. It will return a table containing only columns that are specified,
+ignoring all others. Only columns in the group key that are also specified in `keep` will be kept
```
nathanielc: I like this phrasing, can you move this sentence to its own line as well?

adamperlin:
@nathanielc changes made

PR #527, opened 2018-08-01
fix(http): only log an error and reference code if it exists

jsternberg:
We should not log fields that don't exist since it clutters up the logs.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -95,16 +95,23 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			"status":  statusClass,
 		}).Observe(duration.Seconds())
 		if h.Logger != nil {
-			err := errors.New(w.Header().Get(ErrorHeader))
-			errReference := w.Header().Get(ReferenceHeader)
-			h.Logger.Info("served http request",
+			errField := zap.Skip()
+			if errStr := w.Header().Get(ErrorHeader); errStr != "" {
+				errField = zap.Error(errors.New(errStr))
+			}
+			errReferenceField := zap.Skip()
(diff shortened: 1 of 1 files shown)


nathanielc:
I disagree slightly. Since we have structured logging I see these logs being placed in something searchable. I want to be able to search for logs where `error == ""` for example. I guess this will depend on the implementation of said log searching solution, but in the mean time its still nice to be able to grep for `error= ` etc.

I am not super opinionated here, those are just my initial thoughts.

jsternberg:
After having given this a try, I've been finding the current behavior inconvenient and difficult to read. When it comes to grepping, it makes it harder to grep for errors because the `error` string exists in lines without errors and if you want to search for logs that didn't have errors, you can still use the same query because a non-existent key is considered an empty tag.

I think it is more correct to only record an error when there is an error.

nathanielc:
@jsternberg To me it just comes down to the system consuming these logs.

I had a chat @goller today and he mentioned that there is some in-progress work to take the key value pairs in these logs and store them as fields or tags in the db. 

@goller   Will that work be tolerant of log lines that have an inconsistent set of keys? Meaning if we omit the `error` key when the there is no error will the system still be able to handle that? And query data with predicates like `error == ""`?

goller:
@danielnelson regarding logfmt secondary parsing, would you be able to chime in for @nathanielc ?

danielnelson:
Yes, it will work with an inconsistent keyset.   After parsing a field with any of the parsers it can either merge the output onto the original metric, overriding in the case of conflicts, or it can produce a new metric.

https://github.com/influxdata/telegraf/tree/master/plugins/processors/parser#example

PR #529, opened 2018-08-01
feat(task): update the scheduler and logwriter interface

lyondhill:
Preperatory change that should enable us to build more complex log writers.

change the log writer interface to include more then just the task id
_What was the problem?_
The log writer interface didn't provide enough information to allow us to write task data back into the system bucket

  - [x] Rebased/mergeable
  - [x] Tests pass


lyondhill:
Yes Im preparing it now.

PR #532, opened 2018-08-01
Revert/ids

desa:
Reverts commits from  https://github.com/influxdata/platform/pull/442 as the corresponding PR in idpe https://github.com/influxdata/idpe/pull/1271 could not be merged as it was and the required changes would be substantial.


mark-rushakoff:
Could this have been a single revert commit? Probably `git revert -n d2a11aa^..bdaee706; git commit -m "revert #442"`?

desa:
@mark-rushakoff missed `-n`. Made into a single commit

PR #536, opened 2018-08-01
Feature/query v1 sources

121watts:


http/flux_lang.go
```diff
@@ -0,0 +1,148 @@
+package http
```
goller: please update the swagger docs.
Also, add tests for this

http/flux_lang.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if err := encodeResponse(ctx, w, http.StatusOK, ast); err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+}
+
+// getFluxSuggestions returns a list of available Flux functions for the Flux Builder
+func (h *FluxLangHandler) getFluxSuggestions(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
```
goller: make sure to add an issue about token checking

http/platform_handler.go
```diff
@@ -30,6 +33,26 @@ func setCORSResponseHeaders(w nethttp.ResponseWriter, r *nethttp.Request) {
 	}
 }
 
+var platformLinks = map[string]interface{}{
```
goller: more swagger plz

http/platform_handler.go
```diff
@@ -55,8 +89,8 @@ func (h *PlatformHandler) ServeHTTP(w nethttp.ResponseWriter, r *nethttp.Request
 	ctx := r.Context()
 	var err error
 	if ctx, err = extractAuthorization(ctx, r); err != nil {
-		nethttp.Error(w, err.Error(), nethttp.StatusBadRequest)
-		return
+		// TODO(desa): add back eventually when things have settled
```
goller: ... and link to this code in the issue.

http/platform_handler.go
```diff
@@ -110,3 +144,12 @@ func extractAuthorization(ctx context.Context, r *nethttp.Request) (context.Cont
 	}
 	return idpctx.SetToken(ctx, t), nil
 }
+
+func mustMarshalJSON(i interface{}) []byte {
+	b, err := json.Marshal(i)
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal json: %v", err))
```
goller: why panic?  I'd much prefer not to see this

query.go
```diff
@@ -0,0 +1,25 @@
+package platform
+
+import (
+	"context"
+	"io"
+)
+
+// TODO(desa): These files are possibly a temporary. This is needed
```
goller: make sure to add this issue so you come back to it

http/platform_handler.go
desa: Will remove. This was old. Was previously used to turn links into a byte slice at compile time.

http/platform_handler.go
desa: Opened issue. https://github.com/influxdata/platform/issues/593 will like to issue in the code here.

http/flux_lang.go
desa: Which token checking issue?

query.go
desa: Opened https://github.com/influxdata/platform/issues/594

http/flux_lang.go
desa: @goller Just started doing this now. While doing it, I came to realize that the `/v2/flux/ast` endpoint will take quite a bit of time to implement. Should I do that now or create a separate task to do it later and add a note in the route linking to the issue?

http/flux_lang.go
goller: @desa well, I guess I was thinking that one should have a token that allows querying to some bucket to reach this endpoint.  do you think it should be unauthenticated?

http/flux_lang.go
goller: @desa separate issue would be better.  Easier then to distribute the work

http/flux_lang.go
desa: good question. I'll add a note here mentioning to think about how this route should be authorized. My current thinking was that any authenticated user should be able to use these routes.

PR #537, opened 2018-08-01
fix(http): increase errorHeaderMaxLength

### LINKED ISSUE #530: Errors are being truncated
Running a following request, for example, will return an error that is truncated:
> curl -i -X POST  "http://localhost:9999/v1/tasks" -d '{"organization":"0268be95082dc000", "name":"second name", "flux":"option task = {name:\"first name\", cron:\"* * * * *\"}"}'

Returns:
> **X-Influx-Error: rpc error: code = Unknown desc = missing required fields to crea**

mark-rushakoff:
The limit was introduced in 510325ea5cfe1e77c45846a08f87d96dbbbfaba8 but
it didn't specify why 64 was chosen.

According to https://stackoverflow.com/a/6160643, we should be able to
assume at least 8kb of space for HTTP headers, so I'm assuming 256 bytes
should be sufficient to avoid truncating most error messages, without
being likely to hit the total HTTP header limit.

Fixes #530.


jsternberg:
As an aside, maybe we should be using trailers for stacktraces? I know we use a trailer for returning query statistics and it seems to me like that would be more appropriate for stacktraces rather than putting it into a header.

mark-rushakoff:
The godoc for the Trailer field on the [Request type](https://golang.org/pkg/net/http/#Request) says

>  Few HTTP clients, servers, or proxies support HTTP trailers.

So I don't know if using trailers means we would have to sacrifice support for other HTTP clients.

I'm also not sure why we're packing errors into headers in the first place, rather than encoding them in the response body.

jsternberg:
I do agree that we should define some standards for interprocess behavior. I think it would probably be good to define a common error type that can be decoded such as using protobuf or flatbuffers and stick to those for interprocess communication.

jsternberg:
This still truncates panics.

fntlnz:
Maybe since the problem we tried to solve by truncating was only happening
when there’s a proxy in front of it we could truncate only if there’s a
proxy pass header in the request is coming from it.

In this way we can be sure that we respect standards when exposing via
proxies and can still debug the server when calling it directly.

I don’t think we want to expose panics as errors to the public anyway.

PR #539, opened 2018-08-02
Uint64 platform.IDs - re-opening #442

desa:
https://github.com/influxdata/platform/pull/442

cc @leodido


leodido:
https://github.com/influxdata/platform/blob/14de10ad981c68aa68a88470439cfaf65693bf10/bolt/source_test.go#L19-L21

This errors because `b` (a `platform.Source`) contains an `OrganizationID` equal to zero; which is an invalid `platform.ID`.

cc @nathanielc, @desa

leodido:
PR is WIP.

Help in completing it would be appreciated since it touches the whole platform and it is very invasive.

bolt/source.go
```diff
@@ -21,15 +20,26 @@ var DefaultSource = platform.Source{
...
 
+const (
+	// DefaultSourceID it the default source identifier
+	DefaultSourceID = "020f755c3c082000"
+	// DefaultSourceOrganizationID is the default source's organization identifier
+	DefaultSourceOrganizationID = "50616e67652c206c"
+)
+
 func init() {
 	// TODO(desa): This ID is temporary. It should be updated to be 0 when we switch to integer ids.
```
leodido: 0 is not a valid ID in the ongoing design of IDs within this PR.

bolt/source.go
```diff
@@ -148,6 +163,10 @@ func (c *Client) findSources(ctx context.Context, tx *bolt.Tx, opt platform.Find
 func (c *Client) CreateSource(ctx context.Context, s *platform.Source) error {
 	return c.db.Update(func(tx *bolt.Tx) error {
 		s.ID = c.IDGenerator.ID()
+		// fixme > what if s does not contain a valid OrganizationID ? or contains an empty, thus invaid, OrganizationID ?
+		if !s.OrganizationID.Valid() {
+			s.OrganizationID = c.IDGenerator.ID()
+		}
```
leodido: @desa I choose to generate a valid `OrganizationID` at creation, since a `0` one would not be valid and it will cause an error (ie., `invalid ID`) during unmarshalling.

leodido: Do you think it is fine?

leodido: Or we should throw an error?

http/dashboard_service.go
```diff
@@ -786,6 +785,7 @@ func (s *DashboardService) AddDashboardCell(ctx context.Context, id platform.ID,
 		return err
 	}
 
+	// fixme > in case c does not contain a valid ID this errors out
```
leodido: What to do if `c`, a pointer to `platform.Cell` does not contain an ID neither a ViewID ? @goller

In such case the marshal will error.

http/dashboard_service.go
```diff
@@ -841,6 +841,7 @@ func (s *DashboardService) UpdateDashboardCell(ctx context.Context, dashboardID,
 		return nil, err
 	}
 
+	// fixme > in case upd does not containa a valid ViewID this errors out
```
leodido: What to do il `upd` does not contain a valid (or not empty) `ViewID` ? @goller 

The marshal will error out in such case.

http/dashboard_service.go
```diff
@@ -786,6 +785,7 @@ func (s *DashboardService) AddDashboardCell(ctx context.Context, id platform.ID,
 		return err
 	}
 
+	// fixme > in case c does not contain a valid ID this errors out
 	b, err := json.Marshal(c)
```
leodido: What to do if c, a pointer to platform.Cell does not contain an ID neither a ViewID ? @goller

In such case the marshal will error.

http/dashboard_service.go
```diff
@@ -841,6 +841,7 @@ func (s *DashboardService) UpdateDashboardCell(ctx context.Context, dashboardID,
 		return nil, err
 	}
 
+	// fixme > in case upd does not containa a valid ViewID this errors out
 	b, err := json.Marshal(upd)
```
leodido: What to do il `upd` does not contain a valid (or not empty) `ViewID` ? @goller

The marshal will error out in such case.

bolt/authorization.go
```diff
@@ -187,7 +193,7 @@ func (c *Client) findAuthorizations(ctx context.Context, tx *bolt.Tx, f platform
 // CreateAuthorization creates a platform authorization and sets b.ID, and b.UserID if not provided.
 func (c *Client) CreateAuthorization(ctx context.Context, a *platform.Authorization) error {
 	return c.db.Update(func(tx *bolt.Tx) error {
-		if len(a.UserID) == 0 {
+		if !a.UserID.Valid() {
```
desa: ❤️

bolt/dashboard.go
```diff
@@ -254,7 +257,7 @@ func (c *Client) RemoveDashboardCell(ctx context.Context, dashboardID, cellID pl
 
 		idx := -1
 		for i, cell := range d.Cells {
-			if bytes.Equal(cell.ID, cellID) {
+			if cell.ID == cellID {
```
desa: ❤️

bolt/source.go
desa: I think throw an error if it's not valid

http/dashboard_service.go
desa: Hmm, could we add methods `DecodeWithZeroValue` and `EncodeWithZeroValue` methods for ID and use those in `MarshalJSON` and `UnmarshalJSON`?

bolt/source.go
leodido: Throwing an error would make the overall behavior different.

Look at this test case, for example.

https://github.com/influxdata/platform/blob/5e20c766a5fb9038d9501bae9c2c1adfdb19be94/testing/source.go#L60-L64

In the case we choose to throw an error we MUST have and `OrganizationID` for the `platform.Source`.

bolt/source.go
leodido: :+1:

task/backend/coordinator/coordinator_test.go
```diff
@@ -30,8 +30,8 @@ func TestCoordinator(t *testing.T) {
 	createChan := sched.TaskCreateChan()
 	releaseChan := sched.TaskReleaseChan()
 
-	orgID := platform.ID("org")
-	usrID := platform.ID("usr")
+	orgID := platformtesting.MustIDFromString("69746f7175650d0a")
```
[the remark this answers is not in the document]
leodido: Here I am just using a valid ID string substituting the old ones.

Ideally in tests we should use a mock of the ID generator (in other test it has been used) but it's not the scope of this PR.

I just want to do a minimal set of changes to adapt services and tests to new uint64 IDs, feel free to open an issue about that when this gets merged.

snowflake/id.go
```diff
@@ -87,8 +86,12 @@ func NewIDGenerator(opts ...IDGeneratorOp) *IDGenerator {
 
 // ID returns the next platform.ID from an IDGenerator.
 func (g *IDGenerator) ID() platform.ID {
-	id := make(platform.ID, 8)
-	i := g.Generator.Next()
-	binary.BigEndian.PutUint64(id, i)
+	var id platform.ID
+	for {
```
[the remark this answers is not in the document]
leodido: :+1:

http/dashboard_service.go
goller: @leodido in this case, we probably need to create a different structure that does not contain the cell id and use it for the request.
See how I did it here:
https://github.com/influxdata/platform/blob/master/http/bucket_service.go#L191-L196
and
https://github.com/influxdata/platform/blob/master/http/bucket_service.go#L537

http/dashboard_service.go
goller: You'll need to apply the same technique here.

task/backend/meta.proto
```diff
@@ -37,7 +37,7 @@ message StoreTaskMetaRun {
   // now is the unix timestamp of the "now" value for the run.
   int64 now = 1;
   uint32 try = 2;
-  bytes run_id = 3 [(gogoproto.customname) = "RunID"];
+  uint64 run_id = 3 [(gogoproto.customname) = "RunID"];
```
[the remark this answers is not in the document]
leodido: :+1:

leodido:
After rebasing this PR on top of the branch there are new things that should be addressed.

For example

https://github.com/influxdata/platform/blob/871de35ff6e752e927d45b9a8355f99b5b64076d/http/write_handler.go#L153

The `ExplodePoints` expects two (byte slice) IDs.

https://github.com/influxdata/platform/blob/871de35ff6e752e927d45b9a8355f99b5b64076d/tsdb/explode.go#L22-L23

I cannot pass to it the encoded version of uint64 IDs since its length is 16, and as you can see, the `ExplodePoints` requires IDs with a length equal to 8.

id.go
```diff
@@ -1,37 +1,69 @@
...
 
 import (
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 )
 
+// IDStringLength is the exact length a string (or a byte slice representing it) must have in order to be decoded into a valid ID.
+const IDStringLength = 16
```
e-dard: `String` is redundant. `IDLength` will do

id.go
```diff
@@ -1,37 +1,69 @@
...
+
+// ErrInvalidID is the error thrown to notify invalid IDs.
+var ErrInvalidID = errors.New("invalid ID")
+
+// ErrInvalidIDLength is the error thrown to notify input does not match the wanted length.
+var ErrInvalidIDLength = errors.New("input must be an array of 16 bytes")
```
e-dard: If it were an array then the compiler would have caught the length mismatch. This must be checking a slice? Either way `input must have a length of 16` would be a better message.

id.go
e-dard: Actually this is really the length of a base 16 ID isn't it?

id.go
```diff
@@ -1,37 +1,69 @@
...
+	return 0
+}
+
 // Decode parses b as a hex-encoded byte-slice-string.
+//
+// It errors if the input byte slice does not have the correct length
+// or if it contains all zeros.
 func (i *ID) Decode(b []byte) error {
-	dst := make([]byte, hex.DecodedLen(len(b)))
+	if len(b) != IDStringLength {
```
e-dard: This implementation seems rather overly complicated and inefficient. How about something like:


```go
func (i *ID) Decode(b []byte) error {
 	if len(b) != IDStringLength { 
            return ErrInvalidIDLength
        }

	res, err = strconv.ParseUint(string(b), 16, 64)
	if err != nil {
            return err
        }
        
        if *i = ID(res); !i.Valid() {
            return ErrInvalidID
        }
        return nil 
}
```

e-dard: Further, if the string allocation is problematic then you can simply add:

```go
func unsafeBytesToString(in []byte) string {
	src := *(*reflect.SliceHeader)(unsafe.Pointer(&in))
	dst := reflect.StringHeader{
		Data: src.Data,
		Len:  src.Len,
	}
	s := *(*string)(unsafe.Pointer(&dst))
	return s
}
```

and then do `strconv.ParseUint(unsafeBytesToString(b), 16, 64)`

id.go
```diff
@@ -41,25 +73,50 @@ func (i *ID) DecodeFromString(s string) error {
 }
 
 // Encode converts ID to a hex-encoded byte-slice-string.
-func (i ID) Encode() []byte {
-	dst := make([]byte, hex.EncodedLen(len(i)))
-	hex.Encode(dst, i)
-	return dst
+//
+// It errors if the receiving ID holds its zero value.
+func (i ID) Encode() ([]byte, error) {
```
e-dard: This method can be boiled down to:

```go
func (i ID) Encode() ([]byte, error) {
    if !i.Valid() {
        return nil, ErrInvalidID
    }

    return strconv.AppendUint(nil, 42, 10), nil
}
```

Again, if allocations of the destination buffer are of concern then refactor `Encode` to accept the destination buffer, and pass that down into `AppendUint`

bolt/user_test.go
```diff
@@ -79,7 +79,7 @@ func TestBasicAuth(t *testing.T) {
 				users: []*platform.User{
 					{
 						Name: "user1",
-						ID:   platform.ID("0"),
+						ID:   platformtesting.MustIDFromString("aaaaaaaaaaaaaaaa"),
```
e-dard: I don't know where `MustIDFromString` is defined but I don't think the name is right. You don't need to specify the type in the method name when there is only really one type that's ever converted to an `ID`. Further, the method is actually about taking an ID that's in _base 16_ and converting it to base 10.

I think the method should be called something like `MustIDBase16` or just `MustID`. There might eventually be a sibling method `MustIDBytes`, but _only_ when there are two methods taking different types.

user.go
```diff
@@ -2,7 +2,7 @@ package platform
 
 import "context"
 
-// User is a user. 🎉
+// User is an user. 🎉
```
mark-rushakoff: `a user` was correct.

task/backend/meta.pb.go
```diff
@@ -1,6 +1,17 @@
...
+	Package backend is a generated protocol buffer package.
+
+	It is generated from these files:
+		meta.proto
+
+	It has these top-level messages:
+		StoreTaskMeta
+		StoreTaskMetaRun
+		StoreTaskMetaManualRun
+*/
```
mark-rushakoff: I think this comment, and the method changes below, were introduced because you're using a different version of protoc or the plugin than the previous committer. I don't know whose version is correct.

inmem/auth_service.go
```diff
@@ -85,7 +84,7 @@ func filterAuthorizationsFn(filter platform.AuthorizationFilter) func(a *platfor
 
 	if filter.UserID != nil {
 		return func(a *platform.Authorization) bool {
-			return bytes.Equal(a.UserID, *filter.UserID)
+			return a.UserID == *filter.UserID
```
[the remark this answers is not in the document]
leodido: Not checking for `Valid()` could cause `0 == 0` (both invalid IDs).

So that filter would return true but not find anything (since invalid IDs cannot exist).

Thus maybe we want to check by `Valid()` to avoid to go ahead in such edge situation. Maybe not. I have no preferences in this case.

task/backend/meta.pb.go
goller: Not sure... let's standardize protoc in a follow-on PR.

bolt/view.go
```diff
@@ -108,7 +112,7 @@ func (c *Client) FindView(ctx context.Context, filter platform.ViewFilter) (*pla
 func filterViewsFn(filter platform.ViewFilter) func(d *platform.View) bool {
 	if filter.ID != nil {
 		return func(d *platform.View) bool {
-			return bytes.Equal(d.ID, *filter.ID)
+			return d.ID == *filter.ID
```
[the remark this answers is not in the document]
goller: It turns out the validation checks on the data already stored in the database are not needed.  The ID validation is handled on load and unmarshal of the struct.

bolt/user_test.go
goller: @e-dard good point.  I like your MustIDBase16

id.go
```diff
@@ -41,25 +73,50 @@ func (i *ID) DecodeFromString(s string) error {
...
+	if !i.Valid() {
+		return nil, ErrInvalidID
+	}
+
+	b := make([]byte, hex.DecodedLen(IDStringLength))
+	binary.BigEndian.PutUint64(b, uint64(i))
+
+	dst := make([]byte, hex.EncodedLen(len(b)))
+	hex.Encode(dst, b)
```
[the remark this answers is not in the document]
goller: Hey @mark-rushakoff here is the performance difference (a benchmark will be committed as well):

```
benchmark               old ns/op     new ns/op     delta
BenchmarkIDEncode-4     46.1          174           +277.44%

benchmark               old allocs     new allocs     delta
BenchmarkIDEncode-4     2              3              +50.00%

benchmark               old bytes     new bytes     delta
BenchmarkIDEncode-4     24            40            +66.67%

```

id.go
goller: @e-dard I tried your solution here:

Without unsafe it is slower and take more memory:

```
benchmark               old ns/op     new ns/op     delta
BenchmarkIDDecode-4     40.7          50.3          +23.59%

benchmark               old allocs     new allocs     delta
BenchmarkIDDecode-4     1              1              +0.00%

benchmark               old bytes     new bytes     delta
BenchmarkIDDecode-4     8             16            +100.00%
```

With unsafe it is fast and no allocs:

```
benchmark               old ns/op     new ns/op     delta
BenchmarkIDDecode-4     40.7          31.8          -21.87%

benchmark               old allocs     new allocs     delta
BenchmarkIDDecode-4     1              0              -100.00%

benchmark               old bytes     new bytes     delta
BenchmarkIDDecode-4     8             0             -100.00%

```

id.go
goller: @e-dard that'll convert the ID, but, it doesn't pad with leading ascii zeros.

However, I have a benchmark in place for us to make this better. Perhaps a follow-on PR?

PR #540, opened 2018-08-02
refactor(task): Move task meta into the backend

lyondhill:
Task meta belongs with the rest of the task backend content.

When we moved all the shared interfaces and structures to backend we forgot one.
_What was the problem?_
The pb package was only used for store task meta
_What was the solution?_
Move the meta data into backend.

  - [ ] Rebased/mergeable
  - [x] Tests pass
  - [ ] idpe updates


task/backend/meta.proto
```diff
@@ -0,0 +1,20 @@
...
+
+// StoreTaskMeta is the internal state of a task.
+message StoreTaskMeta {
+  int32 max_concurrency = 1;
+  int64 last_completed = 2;
+  repeated StoreTaskMetaRun currently_running = 3;
+}
+
+message StoreTaskMetaRun {
+  int64 now = 1;
```
mark-rushakoff: This needs a more descriptive name, or at least a comment describing it. What are the units? Seconds, milliseconds, nanoseconds? What time does zero represent? `last_completed` above needs similar clarification.

lyondhill: Good point. I like the simple 'now' but I think it needs comments.

PR #541, opened 2018-08-03
docs(query): modify the flux readme to turn it into a getting started instead

### LINKED ISSUE #509: Simplify and update Flux README
Move the functions and their examples from the README into the SPEC.

Keep the README a simple targeted getting started guide

Explore using the `influx` command instead of raw curl command examples.

jsternberg:
For details, the user can always look at the spec, but a good getting
started guide that works is needed.

At the moment, this does not use the influx command. It does include
instructions for how to get started with docker since that is easiest to
setup and use.

Fixes #509.

### DIFF
--- a/query/README.md
+++ b/query/README.md
@@ -5,47 +5,68 @@ servers.
 
 `fluxd` runs on port `8093` by default
 
-### Specification
+## Specification
 
 A complete specification can be found in [SPEC.md](./docs/SPEC.md).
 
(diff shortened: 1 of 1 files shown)


jsternberg:
Yes. It is in the nightlies.

jsternberg:
I modified the curl command since the original one forgot `orgName`. The parameter doesn't matter at the moment so I just said that.

PR #542, opened 2018-08-03
Do for in-mem tasks

docmerlin:
This adds an Each method to the inmem task store so we can handle user and org deletions more easily in the enterprise version.


mark-rushakoff:
If this is for testing, what tests are going to use `Do`?

lyondhill:
I think im with mark. What value does this add or what problem does it fix?

docmerlin:
@mark-rushakoff @lyondhill this is for deletion of orgs and groups in Enterprise.

docmerlin:
Closing this.

PR #549, opened 2018-08-03
Add Grow()  functions to the table builder interface

### LINKED ISSUE #548: Add Grow()  functions to the table builder interface
a Grow() function will extend a column by the zero value for the type.  This is a good addition because under certain circumstances you can avoid several allocations compared to an append()  approach.

aanthony1243:
Closes #548


query/execute/table.go
```diff
@@ -481,6 +528,12 @@ func (b ColListTableBuilder) AppendTimes(j int, values []Time) {
 	col.data = b.alloc.AppendTimes(col.data, values...)
 	b.table.nrows = len(col.data)
 }
+func (b ColListTableBuilder) GrowTimes(j, n int) {
+	b.checkColType(j, query.TTime)
+	col := b.table.cols[j].(*timeColumn)
+	col.data = b.alloc.GrowTimes(col.data, n)
```
adamperlin: I just have one question: if  the `col.data` slice is grown, we're adding extra empty rows here based on the amount we're growing by. Do we want users of the builder to access these extra empty rows?

aanthony1243:
in the comments I added I note that it's the responsibility of the caller
to make sure all columns are the same size.

adamperlin:
To clarify further (I didn't phrase my comment well, sorry): in the case the slice is re-sized, its new **length** is `oldLen + 1.5*n`, i.e., `len(newSlice) == len(oldSlice) + 1.5 * n`.  So it's actually extending the column by `1.5 * n` zero values, not `n` zero values.

query/execute/allocator.go
```diff
@@ -71,6 +71,20 @@ func (a *Allocator) AppendBools(slice []bool, vs ...bool) []bool {
...
 }
 
+func (a *Allocator) GrowBools(slice []bool, n int) []bool {
+	newCap := len(slice) + n
+	if newCap < cap(slice) {
+		return slice[:newCap]
+	}
+	// grow capacity same way as built-in append
+	newCap = newCap*3/2 + 1
+	s := make([]bool, newCap)
```
adamperlin: Should this be `s := make([]bool, len(slice) + n, newCap)`? This is allocating with initial length but not specifying capacity. See below comment.

aanthony1243: yes,  good catch

PR #552, opened 2018-08-03
test: use zaptest logger

mark-rushakoff:
In places where we were explicitly passing a logger, or where we were
able to pass a logger via functional options, we now pass a zaptest
logger, which will call t.Logf, meaning you will see the log message if
the test fails or if you run go test -v.

I did not modify any of the places where we skip setting a logger with
the WithLogger method pattern, which I believe we've considered
deprecated now.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -864,7 +864,7 @@
   version = "v1.1.0"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
Closing because this is stale.

PR #556, opened 2018-08-06
Add sources to platform

desa:

This PR introduces sources to platform and removes various chronograf features related to v1 sources.


bolt/source.go
```diff
@@ -0,0 +1,273 @@
...
+
+// DefaultSource is the default source.
+var DefaultSource = platform.Source{
+	Default: true,
+	Name:    "autogen",
+	Type:    platform.SelfSourceType,
+}
+
+func init() {
+	if err := DefaultSource.ID.DecodeFromString("020f755c3c082000"); err != nil {
```
goller: Is this temporary?

bolt/source.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
+
+func (c *Client) findSourceByID(ctx context.Context, tx *bolt.Tx, id platform.ID) (*platform.Source, error) {
+	var s platform.Source
```
goller: Perhaps move this initialization to line 106 near the JSON unmarshal?

bolt/source.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if len(v) == 0 {
+		return nil, platform.ErrSourceNotFound
+	}
+
+	if err := json.Unmarshal(v, &s); err != nil {
+		return nil, err
+	}
+	if err := c.setServices(ctx, &s); err != nil {
+		c.Logger.Debug("could not set services on source", zap.Error(err))
```
goller: would you comment why this isn't an error?

http/influxdb/bucket.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		for _, rp := range rps {
+			d, err := time.ParseDuration(rp.Duration)
+			if err != nil {
+				return nil, 0, err
+			}
+			b := &platform.Bucket{
+				// TODO(desa): what to do about IDs?
+				RetentionPeriod: d,
+				Name:            fmt.Sprintf("%v.%v", db.Name, rp.Name),
```
goller: this is different than what we have been doing with dp/rp with a slash in the bucket.

However, I suspect dot is better.  Would you add an issue about bucket naming using dot rather than slash?

desa: 👍 We also changed this in the other PR so that `RetentionPolicyName` is not a separate attribute that we set for v1 sources.

desa: Opened https://github.com/influxdata/platform/issues/559

bolt/source.go
desa: Yeah, was hoping to making `0` as soon as we had the uint ID migration. I'll add a comment noting that it is temporary.

Issue #139, opened 2018-06-15
Add aggregate function


odd:
Some sketches:
```javascript
// alternative possible calling syntax
aggregate(every: 10m, fns: {max: (r) => max(), min: (r) => min(), mean: (r) => mean()})
aggregate(every: 10m, fns: {max, min, mean, myMean})
aggregate(every: 10m, fns: [max, min, mean, myMean])
aggregate(every: 10m, fns: (r) => ({max: max(), min: min(), mean: mean()}))

// implementation
aggregate = (table=<-, every, fns) => {
    tables = fns
        |> map(fn: (name, fn) =>
            table
                |> window(every: every)
                |> fn()
        )
    merge = (tables) => fns
        |> map(fn: (name, fn) => tables[name]._value)
    join(tables: tables, on: ["_time"], merge))
}
```

nathanielc:
Adding some more details:

Another similar  implementation of aggregate helper function
```
aggregate = (aggregates, table=<-, column="_value") =>  {
    tables = aggregates |> map(fn: (k,v) => {
        return { [k]: table |> v(columns:[column]) }
    })
    return join(tables: tables, except: [column])
}
```

Do we want to add `windowing` to these helper functions as well?



This issue will depend on these being complete:

https://github.com/influxdata/platform/issues/212 - object literals
https://github.com/influxdata/platform/issues/534 - remove time handling from aggregates
https://github.com/influxdata/platform/issues/249 - type inference
https://github.com/influxdata/platform/issues/535 - map of objects

nathanielc:
Adding example of how to call aggregate:

```

from(db:"foo")
    |> range(...)
    |> filter(...)
   // Produces a table with columns sum_value, mean_value, and count_value
    |> aggregate(aggregates:{sum,mean,count})

```

pauldix:
Might make sense to have this split up into two functions `apply` and `applyWindow`. The first would be defined like: `apply = (fns, columns= ["_value"], <-tables)` and the other would be something like:

```
applyWindow = (fns, columns=["_value"], every=0s, period=0s, <-tables) =>
    tables |> window(every:every, period:period) |> apply(fns, columns) |> window(every:inf)
```

Calling it aggregates is a little off because it's really about applying any functions to those tables.

PR #575, opened 2018-08-07
Add Duplicate function

### LINKED ISSUE #569: Add "duplicate" schema mutation
Duplicate a column

adamperlin:
Closes #569 

Adds a `duplicate` function to flux, along with relevant SPEC.

_What was the problem?_

A `duplicate` function was desired to make transpiling queries easier.

_What was the solution?_

Create an implementation for a `duplicate` function

  - [x] Rebased/mergeable
  - [x] Tests pass


adamperlin:
@nathanielc Let me know what you think. I added an `n` argument to allow for making more than one copy, but this may not be necessary. If not, I'll remove it.

query/functions/schema_mutators.go
```diff
@@ -318,6 +318,80 @@ func (m *DropKeepMutator) Mutate(ctx *BuilderContext) error {
...
+	for c := range m.Cols {
+		if execute.ColIdx(c, cols) < 0 {
+			return fmt.Errorf(`duplicate error: column "%s" doesn't exist`, c)
+		}
+	}
+	return nil
+}
+
+func (m *DuplicateMutator) Mutate(ctx *BuilderContext) error {
+	if err := m.checkColumnReferences(ctx.Cols()); err != nil {
```
aanthony1243: all of your logic here is OK, but I'm not convinced that this makes sense as a  Mutator type.   You could make a case for it, but if you had this code in its own duplicate.go  file it would be short + sweet and simple to maintain.  It has less in common with the other mutuators,  which is what makes me think this isn't really the place for it.  I'm open to debate on this point, though.

query/functions/schema_mutators.go
```diff
@@ -318,6 +318,80 @@ func (m *DropKeepMutator) Mutate(ctx *BuilderContext) error {
...
+		return nil, fmt.Errorf("invalid spec type %T", qs)
+	}
+
+	return &DuplicateMutator{
+		Cols: toStringSet(s.Cols),
+		N:    s.N,
+	}, nil
+}
+
+// TODO: make checkColumnReferences a separate utility function since all mutations use it
```
aanthony1243: if we keep this as a mutator type, let's resolve this TODO for this PR.

query/functions/schema_mutators.go
```diff
@@ -318,6 +318,80 @@ func (m *DropKeepMutator) Mutate(ctx *BuilderContext) error {
...
+			newColMap = append(newColMap, oldColMap[i])
+
+			if !dup {
+				break
+			}
+		}
+	}
+
+	ctx.TableColumns = newCols
+	ctx.TableKey = execute.NewGroupKey(keyCols, keyValues)
```
aanthony1243: as discussed, I don't think this is necessary in this case.  use the existing table group key.

query/functions/schema_mutators.go
```diff
@@ -318,6 +318,80 @@ func (m *DropKeepMutator) Mutate(ctx *BuilderContext) error {
...
+		keyed := keyIdx >= 0
+
+		// Structuring as a loop allows for duplicating `n` times
+		// if that becomes necessary
+		for count := int64(0); count < m.N+1; count++ {
+			if keyed {
+				keyCols = append(keyCols, c)
+				keyValues = append(keyValues, ctx.Key().Value(keyIdx))
+			}
+			newCols = append(newCols, c)
```
aanthony1243: I'm not sure what it will do to our data model to have 2 columns with the same name.  For example, if we duplicate `a`  and then later want to rename `a`, which one do we rename?  the answer is probably the one earlier in the table, but that may not be what we want.  It also makes it basically impossible to rename the duplicate column without first renaming the original.  

did you run this design past Paul or Nathaniel?   I think we probably need to suffix the new column name to make it unique.

adamperlin: Ok, sounds good. I haven't yet run the design past Paul or Nathaniel. I'll link them in this thread though.

adamperlin:
@pauldix @nathanielc I'd love some feedback on the `duplicate` design SPEC.

query/docs/SPEC.md
```diff
@@ -901,6 +901,34 @@ Additionally exactly two columns must be provided to the `columns` property.
...
+#### Duplicate 
+Duplicate will duplicate specified columns in a data table.
+Given a list of columns and a number of copies to make, it will make 
+copies of those columns and add them to the returned data table. 
+
+Duplicate has the following properties:
+
+* `columns` array of strings
+    An array of columns to duplicate
+* `n` int
```
nathanielc: I think we want to change the API here. Duplicate should take a map of columns, where the key is a new column label and the value an existing column value. We want to avoid naming schemes if possible and allow the user to be explicit about the name.  The `n` parameter is not needed. If they want to duplicate a single column multiple times use multiple entries in the columns map.

The use case for duplicate is for aggregations to preserve data through the aggregation step so it is normalized. 
Meaning aggregate will drop all columns that are not being aggregated and that are not part of the group key. This means that `_time` will almost always be dropped. So doing `|> duplicate(columns: {_stop:"_time"})` after an aggregate will add the `_time` column back on the data using the value of the group column `_stop`.

query/functions/schema_mutators.go
```diff
@@ -225,19 +225,19 @@ func NewDropKeepMutator(qs query.OperationSpec) (*DropKeepMutator, error) {
 	return m, nil
 }
 
-func (m *DropKeepMutator) checkColumnReferences(cols []query.ColMeta) error {
-	if m.DropCols != nil {
-		for c := range m.DropCols {
-			if execute.ColIdx(c, cols) < 0 {
-				return fmt.Errorf(`drop error: column "%s" doesn't exist`, c)
-			}
+func checkColumnsBoolMap(toCheck map[string]bool, tableCols []query.ColMeta) error {
```
nathanielc: What does this function do?

nathanielc:
We also need to define what happens if you duplicate a column in the group key. Does it get added to the group key? I think not, it cannot change the grouping so there is no value in automatically adding it to the group key.

query/docs/SPEC.md
pauldix: I'd like to flesh out more about what the use case is here. Can we expand with some code examples to show different scenarios? Is the aggregate one the only case we can think of? Why duplicate the `_stop` column rather than just renaming it to `_time`? Also, it seems far more readable/clear if the definition was simply `duplicate(column:"_stop", to:"_time")`. What is the use case for copying a column multiple times?

aanthony1243: I talked to @adamperlin about the `n` parameter.  He put it in there because it was trivial to extend 'copy once' to 'copy n', but he did admit that it might not make sense to do that.  

I think I like that duplicate would duplicate only one column at a time.  it's a clean definition, and simple enough that calling it multiple times if needed is probably not too much of a burden.

nathanielc: @pauldix This comes out of a discussion we had a while back about aggregates not managing the `time` aspect and leaving that to the user.

The desire to duplicate the `_stop` column as the `_time` column is to help the "it works by default" mantra. Various functions operate on columns and in all cases the columns on which they operate are configurable but also have a good default. 

In the case of the `window` function the default column on which to window the data is `_time`. The both the `range` and `window` functions make use of the `_start` and `_stop` columns by default as well.

Since  an aggregate is commonly an aggregate along the `_time` dimension, the `_time` column will be lost during the aggregate operation. In order to make preserving the expected default `_time` column the `duplicate` operation can be used to duplicate on of the group key times as the `_time` column, typically this is the `_stop` column. Meaning that the `_time` associated with an aggregate is the upper bound time of the aggregated data. The user could choose to use `_start` or anything else custom they may have.

In summary the known use case is to duplicate a single column after an aggregate operation.

Then in order to make this easier for users to manage we plan to create a helper aggregate function https://github.com/influxdata/platform/issues/139

nathanielc: As a follow up, we can easily rewrite multiple duplicate operations in to a single transformation under the hood, so if we make the user write out `duplicate` multiple times there will be no performance impact.

adamperlin: If our usage case is primarily to duplicate a single column, then should we just go with the single column syntax that @pauldix suggested, `duplicate(column:"_time", to: "_stop")`? The implementation can still be written to work with multiple columns. I think it might be even more readable to rename `to` to `as`, so we would have:

```js
from(db: "db")
   |> range(...)
   |> aggregateOperation(...)
   |> duplicate(column: "_stop", as: "_time")
```

pauldix: @adamperlin yeah that looks good.

pauldix:
@nathanielc I would say that it shouldn't alter the group key. Maybe the `group`, `window`, and `join` functions are the only ones that can alter a group key?

adamperlin:
@nathanielc @aanthony1243 Added the new implementation. However, it would be nice to have a fruitful example for the SPEC, potentially an aggregate operation where a time column would need to be duplicated. @nathanielc could you give a concrete example of the usage case you were describing?

nathanielc:
@adamperlin Let's just give a trivial example of duplicating a column. Its not hard to understand what it does, I do not think it needs the context of aggregate functions to be understood. The spec does not have to answer why it is useful.

PR #578, opened 2018-08-07
feat(task): connect the api to tasks

lyondhill:
We need a coordination layer to facilitate the communication in the api to the task service.
The api was also not connected in main to the http service.

### DIFF
--- a/cmd/idpd/main.go
+++ b/cmd/idpd/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/influxdata/platform/http"
 	"github.com/influxdata/platform/kit/prom"
 	"github.com/influxdata/platform/query"
+	_ "github.com/influxdata/platform/query/builtin"
 	"github.com/influxdata/platform/query/control"
 	"github.com/influxdata/platform/query/execute"
 	"github.com/influxdata/platform/task"
@@ -136,8 +137,7 @@ func platformF(cmd *cobra.Command, args []string) {
 	{
 		boltStore, err := taskbolt.New(c.DB(), "tasks")
 		if err != nil {
-			logger.Error("failed opening task bolt", zap.Error(err))
-			os.Exit(1)
+			logger.Fatal("failed opening task bolt", zap.Error(err))
 		}
 
 		executor := taskexecutor.NewQueryServiceExecutor(logger, queryService, boltStore)
--- a/cmd/idpd/main.go
+++ b/cmd/idpd/main.go
@@ -7,6 +7,7 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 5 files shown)


cmd/idpd/main.go
```diff
@@ -18,6 +19,14 @@ import (
 	"github.com/influxdata/platform/chronograf/server"
 	"github.com/influxdata/platform/http"
 	"github.com/influxdata/platform/kit/prom"
+	"github.com/influxdata/platform/query"
+	"github.com/influxdata/platform/query/control"
```
mark-rushakoff: I think we'll need `_ "github.com/influxdata/platform/query/builtin"` here too. Not sure whether it will do anything now but may as well include it since we're in main.go.

cmd/idpd/main.go
```diff
@@ -108,6 +117,38 @@ func platformF(cmd *cobra.Command, args []string) {
...
+			AsyncQueryService: control.New(config),
+		}
+	}
+
+	var taskSvc platform.TaskService
+	{
+		boltStore, err := taskbolt.New(c.DB(), "tasks")
+		if err != nil {
+			logger.Error("failed opening task bolt", zap.Error(err))
+			os.Exit(1)
```
mark-rushakoff: Prefer `logger.Fatal` over `logger.Error(...); os.Exit`.

http/platform_handler.go
```diff
@@ -89,6 +91,11 @@ func (h *PlatformHandler) ServeHTTP(w nethttp.ResponseWriter, r *nethttp.Request
 		return
 	}
 
+	if strings.HasPrefix(r.URL.Path, "/v1/tasks") {
+		fmt.Println("handled?")
```
mark-rushakoff: Remove print.

task/backend/coordinator/coordinator.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return id, err
+	}
+
+	task, err := c.st.FindTaskByID(ctx, id)
+	if err != nil {
+		return id, err
+	}
+
+	opt, err := options.FromScript(script)
```
mark-rushakoff: This should be the first call in CreateTask, so we don't create a task if the script or options are invalid.

task/backend/coordinator/coordinator.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if err := c.sch.ClaimTask(task, meta.LastCompleted, &opt); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *Coordinator) ListTasks(ctx context.Context, params backend.TaskSearchParams) ([]backend.StoreTask, error) {
```
mark-rushakoff: There's a few methods here that are basically just delegating the call to the underlying store, so maybe embedding the backend.Store is a better fit here so that we don't have to write those calls explicitly. But it's probably fine as-is.

task/backend/coordinator/coordinator_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	task, err = timeoutSelector(createChan)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if task.Script != newScript {
+		t.Fatal("task sent to scheduler doesnt match task created")
+	}
+
```
mark-rushakoff: Remove blank line.

PR #579, opened 2018-08-08
Implement chronograf v2 dashboards in platform

desa:
This PR introduces v2 dashboards and replaces the dashboards that were previously implemented in platform.

- [x] Corresponding PR (https://github.com/influxdata/idpe/pull/1469)
- [x] Rebase onto master
- [x] Squash commits
- [x] Swagger for dashboards and views
- [x] Swagger for sources


bolt/dashboard.go
```diff
@@ -4,29 +4,22 @@ import (
...
+	"strconv"
 
-	"github.com/coreos/bbolt"
+	bolt "github.com/coreos/bbolt"
 	"github.com/influxdata/platform"
 )
 
 var (
-	dashboardBucket = []byte("dashboardsv1")
+	dashboardV2Bucket = []byte("dashboardsv3")
```
goller: v2 vs v3

chronograf/server/TODO.go
```diff
@@ -1,14 +0,0 @@
-package server
```
desa: 😢

desa:
@goller when you have a chance, this could use a review.

http/dashboard_service.go
```diff
@@ -218,497 +272,268 @@ func (h *DashboardHandler) handlePatchDashboard(w http.ResponseWriter, r *http.R
...
+	if err := json.NewDecoder(r.Body).Decode(req.cell); err != nil {
 		return nil, err
 	}
 
-	return &postDashboardCellRequest{
-		DashboardCell: c,
-		DashboardID:   i,
-	}, nil
+	qp := r.URL.Query()
+	if cv := qp.Get("copyView"); cv != "" {
```
goller: Why have a copyView?  Why not GET and then POST?

if we want to do this we would put it at `/cells` with something like this:

```json
{
    "copyFrom": "link"
}
```

you aren't acting on the cell id you are acting on cells with information from another cell.

http/dashboard_service.go
```diff
@@ -218,497 +272,268 @@ func (h *DashboardHandler) handlePatchDashboard(w http.ResponseWriter, r *http.R
...
-	if !bytes.Equal(req.Cell.ID, cid) {
-		return nil, fmt.Errorf("url cell_id does not match id on provided cell")
-	}
-
 	return req, nil
 }
 
-// handleDeleteDashboardCell is the HTTP handler for the DELETE /v1/dashboards/:id/cells/:cell_id route.
-func (h *DashboardHandler) handleDeleteDashboardCell(w http.ResponseWriter, r *http.Request) {
+// handlePuttDashboardCell creates a dashboard cell.
```
goller: handlePutt should be handlePut

http/dashboard_service.go
```diff
@@ -218,497 +272,268 @@ func (h *DashboardHandler) handlePatchDashboard(w http.ResponseWriter, r *http.R
...
-	hc := newClient(u.Scheme, s.InsecureSkipVerify)
-
-	resp, err := hc.Do(req)
-	if err != nil {
-		return err
-	}
-
-	return CheckError(resp)
-}
+// handlePatchDashboardCell creates a dashboard cell.
```
goller: Patch shouldn't be for creating a cell and why do we need put and patch for cells?

http/swagger.yml
```diff
@@ -31,8 +31,429 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /views:
+    post:
+      tags:
+        - Views
+      summary: Create a view
```
goller: would you add a bit about what view is?

http/swagger.yml
```diff
@@ -1433,3 +1858,359 @@ components:
...
+            self:
+              type: string
+        views:
+          type: array
+          items:
+            $ref: "#/components/schemas/View"
+    Cell:
+      type: object
+      properties:
+        x:
```
goller: should we really have cells?  How would I modify one cell without also having to modify other cells coordinates?

Shouldn't cells layout information like this be in the dashboard if you have to modify more than one at a time?

http/dashboard_service.go
desa: > Why have a copyView? Why not GET and then POST?

So this came out of the need to "clone" cells in a dashboard.

At first we had two requests

1. Create a new view
2. Create a cell that uses that view

The reason we put it into a single operation was so that in failure cases we don't end up with stale views. Things like A user creates a view and the sever crashes or the user refreshes the page before the cell with the view was created.

So we settled on having the user specify whether they would like to make a copy of a cells view or use the view directly. It's my understanding that there is currently not a requirement to allow two cells to have the same view, but this will be a future requirement.

> if we want to do this we would put it at /cells with something like this:
>```
>{
>    "copyFrom": "link"
>}
>```
>you aren't acting on the cell id you are acting on cells with information from another cell.

Had something like this originally, but it felt odd. Specifically because it's not the cell that we're copying and the new cell that we create will have completely distinct attributes, except for it's view. Happy to go back to it if you think that it's a better idea.

http/dashboard_service.go
desa: Fixed.

http/dashboard_service.go
desa: > Patch shouldn't be for creating a cell 
Copy pasta

> why do we need put and patch for cells?
The `PUT` is to `/dashboard/:id/cells` and replaces all of a dashboards cells.
The `PATCH` updates a single cell.

desa: fixed comment

http/swagger.yml
desa: How's `A view contains information about the visual representation of data`?

http/swagger.yml
```diff
@@ -31,8 +31,602 @@ paths:
...
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+  '/dashboards/{dashboardID}/cells/{cellID}':
+    patch:
+      tags:
+        - Cells
+      summary: Update a dashboards cell
```
goller: change this to only update names and specify that it is the only field to update _because_ we don't want conflicts in x,y,w,h

http/swagger.yml
```diff
@@ -1433,3 +2031,417 @@ components:
...
+          type: string
+          description: The reference to a view from the views API
+    Cells:
+      type: array
+      items:
+        $ref: "#/components/schemas/Cell"
+    Dashboard:
+      properties:
+        links:
+          $ref: "#/components/schemas/Links"
```
goller: update this to have the correct links

http/swagger.yml
```diff
@@ -31,8 +31,602 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+  /sources/{sourceID}/health:
```
goller: document the top-level /health

desa:
@goller made the changes we talked about 😃

PR #591, opened 2018-08-08
feat(task): create validation layer for TaskService

lyondhill:


task/validator.go
```diff
@@ -0,0 +1,40 @@
...
+func checkPermission(ctx context.Context, perm platform.Permission) error {
+	auth, err := platcontext.GetAuthorization(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !platform.Allowed(perm, auth.Permissions) {
+		return errors.New("unauthorized")
+	}
+}
```
mark-rushakoff: Missing a return at the end here.

auth.go
```diff
@@ -62,6 +62,8 @@ const (
 	UserResource = resource("user")
 	// OrganizationResource represents the org resource actions can apply to.
 	OrganizationResource = resource("org")
+	// TaskResource represents the task resource actions can apply to.
+	TaskResource = resource("task")
```
mark-rushakoff: I'm not sure if this is okay, or if the task resource needs to be scoped like the bucket resource below.  Conversation happening in https://github.com/influxdata/platform/issue_comments#issuecomment-411576446.

task/validator.go
```diff
@@ -0,0 +1,40 @@
...
+	platcontext "github.com/influxdata/platform/context"
+)
+
+func NewValidator(ts platform.TaskService) platform.TaskService {
+	return &TaskServiceValidator{
+		TaskService: ts,
+	}
+}
+
+type TaskServiceValidator struct {
```
mark-rushakoff: I'm not sure that this type needs to be exported.

task/validator.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/influxdata/platform"
+	platcontext "github.com/influxdata/platform/context"
+)
+
+func NewValidator(ts platform.TaskService) platform.TaskService {
```
mark-rushakoff: Update this to take a logger to associate with the validator.

task/validator.go
```diff
@@ -0,0 +1,40 @@
...
+	if err := checkPermission(ctx, platform.Permission{Action: platform.CreateAction, Resource: platform.TaskResource}); err != nil {
+		return err
+	}
+
+	return ts.TaskService.CreateTask(ctx, t)
+}
+
+// TODO(lh): add permission checking for the all the platform.TaskService functions.
+
+func checkPermission(ctx context.Context, perm platform.Permission) error {
```
mark-rushakoff: A few changes on how this works:

- Change the name to reflect "Create"
- Change to a method on Validator
- Return a bool instead of an error; on error inside the method, log the error and return false
- Callers of the new method should return ErrUnauthorized

task/validator.go
lyondhill: Im not sure if a logger is necessary in this context. Like the PlatformAdaptor this is a invisible layer that acts as a filter.

task/validator.go
lyondhill: This check is intended to reduce repeated code. It is not a subset of the 'create' function but more a permission check used by of the TaskService functions.

task/validator.go
mark-rushakoff: The logger is necessary. We need a record of why authorizations fail.

lyondhill: I agree. That should be done in the request handler.. right?

mark-rushakoff: On principle, you don't expose to the user the specific reason why an authentication failed; the user should see a generic message, not a specific "wrong username" or "wrong password" or "valid username and password but invalid resource".  But the system operators should be able to see the specific reason, hence a plain log until we start doing audit logging.

lyondhill: I agree that we dont want to inform the user. On previous projects I would return specific errors to the handler like ErrUnauthorized, ErrInvalidUser, ErrBadPassword etc.. Then the handlers logged the error and returned 404 not found. This was made possible because all the 'auth' type errors could be asserted to a AuthError type. 
```
type AuthError interface {
  error
  AuthError() string
}
```
This allowed authentication errors to be handled in 1 place and using 1 set of logic.
Is there a reason that pattern wouldn't work here?

mark-rushakoff: I see what you mean now. Custom typed errors have been pretty infrequent in our codebases, but that pattern seems like a good fit for authorization errors. That works for me.

lyondhill: I can setup the code that way and make the changes to the create handler.

auth.go
```diff
@@ -64,6 +64,11 @@ const (
 	OrganizationResource = resource("org")
 )
 
+// TaskResource represents the task resource scoped to an organization.
+func TaskResource(orgID ID) resource {
+	return resource(fmt.Sprintf("org/%s/task", orgID))
```
mark-rushakoff: I like this format, it's really clear.

http/errors.go
```diff
@@ -16,6 +16,11 @@ const (
 	errorHeaderMaxLength = 256
 )
 
+type AuthError interface {
```
mark-rushakoff: Auth is ambiguous - is it authz, authn, or both? 

Can you add godoc for this type? Maybe like:

AuthzError is returned for authorization errors. When this error type is returned, the user can be presented with a generic "authorization failed" error, but the system can log the underlying AuthzError() so that operators have insight into what actually failed with authorization.

task/validator.go
```diff
@@ -0,0 +1,53 @@
...
+
+// TODO(lh): add permission checking for the all the platform.TaskService functions.
+
+func validatePermission(ctx context.Context, perm platform.Permission) error {
+	auth, err := platcontext.GetAuthorization(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !platform.Allowed(perm, auth.Permissions) {
```
mark-rushakoff: What do you think about replacing `platform.Allowed` with a new method that returns an AuthError indicating the first requested permission that missed and the Authorization ID? Since platform.Allowed just returns a boolean, it seems difficult to track down why an operation was denied.

http/errors.go
lyondhill: OO very good suggestion. I will make this update.

task/validator.go
lyondhill: Sounds good.

lyondhill: In this case we are asking if the auth contains a single permission we can only fail on that permission. But I think I found a reasonable alternative since validator knows the auth and the failure I think we can build a better response in the validator.

PR #604, opened 2018-08-10
new duration literals

### LINKED ISSUE #558: Syntax error produced with Flux range function
When I enter this query I get a syntax error:
`from(db:"telegraf") |> range(start:-7d) |> yield()`
```
Error: 1:38 (37): no match found, expected: "!=", "!~", ")", ",", ".", "//", "<", "<=", "==", "=~", ">", ">=", "and"i, "empty"i, "in"i, "ms", "not empty"i, "ns", "or"i, "startswith"i, "us", "|>", "µs", "μs", [ \t\r\n], [*/], [+-], [0-9] or [smh]
```
According to the SPEC `-7d` is a valid duration literal, so it seems like this should work.

Note: if I use `range(start:-1000h)` then it works.

jlapacik:
Closes #558 

_What was the problem?_

The Flux grammar did not support day and week duration literals.

_What was the solution?_

I defined new tokens for `d` and `w` in a duration literal in the Flux grammar file. I also delayed converting an abstract duration literal type to a concrete duration Go type. That conversion now takes place during the construction of a semantic graph.

To test:
* rebuild the project by running `make` from the project root
* `cd` to the bin/darwin directory
* start the Flux cli via `./influx repl`
* Enter the new duration literals to see that they are correctly parsed and interpreted

    ```
    >1d
    >24h0m0s
    >
    >1w
    >168h0m0s
    ```

  - [ ] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/query/ast/ast.go
+++ b/query/ast/ast.go
@@ -791,12 +791,19 @@ func (l *RegexpLiteral) Copy() Node {
 	return nl
 }
 
+// Duration is a pair consisting of length of time and the unit of time measured.
+// It is the atomic unit from which all duration literals are composed.
+type Duration struct {
+	Magnitude int64  `json:"magnitude"`
+	Unit      string `json:"unit"`
+}
+
 // DurationLiteral represents the elapsed time between two instants as an
 // int64 nanosecond count with syntax of golang's time.Duration
 // TODO: this may be better as a class initialization
 type DurationLiteral struct {
 	*BaseNode
-	Value time.Duration `json:"value"`
+	Values []Duration `json:"values"`
 }
 
 // Type is the abstract type
--- a/query/ast/json.go
+++ b/query/ast/json.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
-	"time"
 )
 
 func (p *Program) MarshalJSON() ([]byte, error) {
@@ -729,36 +728,12 @@ func (l *DurationLiteral) MarshalJSON() ([]byte, error) {
 	raw := struct {
 		Type string `json:"type"`
 		*Alias
-		Value string `json:"value"`
 	}{
 		Type:  l.Type(),
 		Alias: (*Alias)(l),
-		Value: l.Value.String(),
 	}
 	return json.Marshal(raw)
 }
-func (l *DurationLiteral) UnmarshalJSON(data []byte) error {
-	type Alias DurationLiteral
-	raw := struct {
-		*Alias
-		Value string `json:"value"`
-	}{}
-	if err := json.Unmarshal(data, &raw); err != nil {
-		return err
-	}
-
-	if raw.Alias != nil {
-		*l = *(*DurationLiteral)(raw.Alias)
-	}
-
-	value, err := time.ParseDuration(raw.Value)
-	if err != nil {
-		return err
-	}
-	l.Value = value
-	return nil
-}
-
 func (l *DateTimeLiteral) MarshalJSON() ([]byte, error) {
 	type Alias DateTimeLiteral
 	raw := struct {
--- a/query/parser/flux.peg
+++ b/query/parser/flux.peg
@@ -362,6 +362,12 @@ MinuteUnits
 HourUnits
   = "h"
 
+DayUnits
+  = "d"
+
+WeekUnits
+  = "w"
+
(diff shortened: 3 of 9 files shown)


query/ast/ast.go
```diff
@@ -811,6 +814,24 @@ func (l *DurationLiteral) Copy() Node {
 	return nl
 }
 
+type SingleDurationLiteral struct {
```
jsternberg: Instead of as a separate AST types, can we build this into the `DurationLiteral` itself? You could do something like this:

```
type Duration struct {
    Mag *IntegerLiteral `json:"magnitude"`
    Unit string `json:"unit"`
}

type DurationLiteral struct {
    *BaseNode
    Values []Duration `json:"values"`
}
```

I feel like this will show up better when marshaling and it will be more simple since I don't think there's a place in the grammar where we accept only a single duration literal. If there is, we can add that as a base node later.

query/semantic/json.go
```diff
@@ -753,6 +753,18 @@ func (l *DurationLiteral) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (l *SingleDurationLiteral) MarshalJSON() ([]byte, error) {
```
wolffcm: So there is no method needed to unmarshal `SingleDurationLiteral`?  Does it always get unmarshal'd to a `DurationLiteral`?

query/parser/flux.peg
```diff
@@ -370,14 +376,18 @@ DurationUnits
       / SecondUnits
       / MinuteUnits
       / HourUnits
+      / DayUnits
+      / WeekUnits
```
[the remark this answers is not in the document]
jlapacik: Yeah days and weeks were only outlined in the spec, so that's why I only added those. We can create a new issue if we want to add more.

query/semantic/graph.go
```diff
@@ -1324,10 +1344,42 @@ func analyzeDateTimeLiteral(lit *ast.DateTimeLiteral, declarations DeclarationSc
 	}, nil
 }
 func analyzeDurationLiteral(lit *ast.DurationLiteral, declarations DeclarationScope) (*DurationLiteral, error) {
+	var duration time.Duration
+	for _, literal := range lit.Values {
+		dur, err := analyzeSingleDurationLiteral(literal, declarations)
```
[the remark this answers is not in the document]
jlapacik: You're correct, that is allowed. I would say let's not place any restrictions on it, as it doesn't introduce any unexpected behavior. But that's just me.

query/ast/json.go
```diff
@@ -889,6 +877,8 @@ func unmarshalNode(msg json.RawMessage) (Node, error) {
 		node = new(RegexpLiteral)
 	case "DurationLiteral":
 		node = new(DurationLiteral)
+	case "SingleDurationLiteral":
+		node = new(DurationLiteral)
```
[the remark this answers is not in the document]
jlapacik: Yes it should be.

query/ast/ast.go
jlapacik: @jsternberg do you mean something like this instead of having an integer literal type inside of a single duration?

```
type Duration struct {
    Mag int64
    Unit string
}
```

jsternberg: Either one. I think yours might be better.

query/parser/types.go
```diff
@@ -16,6 +16,26 @@ func toIfaceSlice(v interface{}) []interface{} {
 	return v.([]interface{})
 }
 
+func toDuration(literal *singleDurationLiteral) (time.Duration, error) {
```
jsternberg: This still needs to be a combination of duration literals. Unfortunately, we can't necessarily concatenate duration literals without losing some of the semantic meaning. We're going to likely add month and year literals in at some point so we should just prepare for that now.

That means this should take an array of `*singleDurationLiteral` values and convert them to an array or `Duration` values with `Duration` being the struct we talked about earlier.

jlapacik: Updated.

query/ast/ast.go
```diff
@@ -791,12 +791,17 @@ func (l *RegexpLiteral) Copy() Node {
 	return nl
 }
 
+type Duration struct {
```
jsternberg: Can you document this before merging?

PR #605, opened 2018-08-10
[WIP] Add storage layer

e-dard:
This is a WIP. I will update the description accordinly.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -513,8 +513,8 @@
   version = "v1.0"
 
 [[projects]]
-  branch = "master"
-  digest = "1:f99f5fc44ba4eda20698e4477665a5a525d9b0ba15fdda78f5c62f87a481eb89"
+  branch = "platform"
+  digest = "1:cfbcc5e88b3631fe7778f5513e3c45013a330442e86b4f4f3bc67badb177d381"
   name = "github.com/influxdata/influxdb"
   packages = [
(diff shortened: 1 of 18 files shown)


e-dard:
**NOT READY TO REVIEW OR MERGE**

This PR is basically a place-holder of some initial work to bring a storage layer into Platform. Some of the things that it is missing:

 1) There is no consumer service. We don't yet know what sort of durable log we will be using in platform. The storage layer will need to consume this log. That service is not present in this PR.
 2) The read service is not wired up or configured as it's not clear if we will just import the read service on the platform branch from OSS or implement one in here. @stuartcarnie can you chime in on that?
3) The retention service needs to be set on the storage service. I think we can use the retention service I have brought in via this PR, but we will need to use a `bolt.Client` to provide access to the `platform.BucketService`.
4) We might want to rethink how we use the local store, or if we want to just embed parts of the `tsdb.Store` or even just use its `Read` and `GroupRead` methods. Needs some thinking...
5) None of the Prometheus metrics are wired up.
6) As we continue to get this working we need to think about making sure that the storage layer is constructed in way that will allow us to embed it in our other products... Consider that we will have different consumer services and so on.
7) None of this is wired up to a binary. I'm not sure how/where we are planning on doing that yet.

e-dard:
Closed in favour of #929 and others.

PR #609, opened 2018-08-11
Add Stream Context to Executor

### LINKED ISSUE #369: Add range bounds to "stream" context
There are two parts:

1. Add explicit stream context to execution engine.
2. Add range bounds context to streams

adamperlin:
Closes #369 

Adds bounds to the planner/executor at the stream level. In the planner, each procedure now has its own bounds, which are inherited from parent procedures.
_What was the problem?_
Having bounds global to the planner did not allow for enough flexibility.
_What was the solution?_
Add explicit stream context with bounds to the planner and execution engine. 
  - [x] Rebased/mergeable
  - [x] Tests pass


query/plan/physical.go
```diff
@@ -134,10 +133,34 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
...
+			pr.Bounds = pr.Bounds.Intersect(bounded.TimeBounds(), now)
+		}
+
+		// If pr is a root procedure and has 0 bounds,
+		// that doesn't necessarily mean it's unbounded because range operations aren't always
+		// pushed down.
+		// If it has parents AND 0 bounds,
+		// or if it's a root node with no children and no parents AND  zero bounds,
+		// we return an error
+		if (len(pr.Parents) > 0 || (len(pr.Parents) == 0 && len(pr.Children) == 0)) &&
```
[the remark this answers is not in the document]
adamperlin: I should clarify this logic a bit and maybe make the comment clearer. The checks are only necessary in the case that range bounds are **not** pushed down into the root `from` node (i.e., if the root node is `fromCSV` or some other source). In that case, the parent won't have any bounds (because it won't inherit any), but its children might, so we have to wait to check the children before erroring (and error if **any child node** has 0 bounds). Also, If we have a root node that's unbounded and also isolated with no children, then that's also invalid and we error. Basically, what this should be is: 
```go
if (!isRootNode(pr) || !hasChildren(pr)) && pr.Bounds.Empty() { ... }
```

adamperlin:
@nathanielc I've made review changes + added what we talked about. Ready for another look

query/execute/executor.go
```diff
@@ -109,10 +118,16 @@ func (es *executionState) createNode(ctx context.Context, pr *plan.Procedure, no
 	if n, ok := nodes[pr.ID]; ok {
 		return n, nil
 	}
+
 	// Build execution context
 	ec := executionContext{
 		es: es,
+		streamContext: newStreamContext(Bounds{
+			Start: values.Time(pr.Bounds.Start.Time(es.p.Now).UnixNano()),
```
nathanielc: The logic to resolve time is now in multiple places. Can you make a helper method for this like `resolveTime(now time.Time, t query.Time) Time`? Then use it here and in the ResolveTime method of the executionContext?

query/plan/procedure.go
```diff
@@ -106,15 +107,65 @@ type BoundsSpec struct {
...
+	if o.Stop.Time(now).After(b.Stop.Time(now)) {
 		u.Stop = o.Stop
 	}
+
+	return
+}
+
+// Intersect returns the intersection of two bounds. If there is no intersection or if
+// either of the bounds have zeroes, it will return a zero-valued BoundsSpec along with `false`.
+func (b BoundsSpec) Intersect(o BoundsSpec, now time.Time) (i BoundsSpec, exists bool) {
```
nathanielc: Why have the `exists` bool? I say we still return the `EmptyBounds` value here since empty bounds are well defined. This way intersection and unions can be composed as expected. 

Meaning if I can do  `A.Intersect(B).Union(C)` and if A and B do not intersect I can still union that empty result with C, without having to have a separate code path.

I may have mislead you in saying we didn't need the EmptyBounds var. What I meant is we would not need it as a default value for the procedure bounds. But it is still useful for making intersect play nice.

query/time_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+	"time"
+
+	"github.com/influxdata/platform/query"
+)
+
+func TestTime(t *testing.T) {
+	t.Log(query.MinTime.Time(time.Time{}))
+	t.Log(query.MaxTime.Time(time.Time{}))
+}
```
nathanielc: Can you remove this test?

query/plan/physical_test.go
```diff
@@ -809,6 +896,190 @@ func TestPhysicalPlanner_Plan(t *testing.T) {
...
+							},
+							TimeCol: "_time",
+						},
+						Bounds: plan.BoundsSpec{
+							Start: query.Time{
+								IsRelative: true,
+								Relative:   -1 * time.Hour,
+							},
+
+							Stop: query.Now,
```
nathanielc: nit, can you remove these empty lines to make the tests shorter, since they are already long enough ;) ?

query/plan/physical_test.go
```diff
@@ -1070,11 +1363,15 @@ func PhysicalPlanTestHelper(t *testing.T, lp *plan.LogicalPlanSpec, want *plan.P
...
 
+	//	log.Println("GOT: ", got)
+	//	log.Println("WANT: ", want)
+
 	if !cmp.Equal(got, want, plantest.CmpOptions...) {
 		t.Log("Logical:", plan.Formatted(lp))
 		t.Log("Want Physical:", plan.Formatted(want))
 		t.Log("Got  Physical:", plan.Formatted(got))
 		t.Errorf("unexpected physical plan -want/+got:\n%s", cmp.Diff(want, got, plantest.CmpOptions...))
+
```
nathanielc: nit, remove this line.

query/plan/physical_test.go
```diff
@@ -1070,11 +1363,15 @@ func PhysicalPlanTestHelper(t *testing.T, lp *plan.LogicalPlanSpec, want *plan.P
 		t.Fatal(err)
 	}
 
+	//	log.Println("GOT: ", got)
+	//	log.Println("WANT: ", want)
```
nathanielc: Remove these comments.

query/plan/physical.go
```diff
@@ -169,8 +191,12 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
 		}
 	}
 
-	if p.plan.Bounds.Start.IsZero() && p.plan.Bounds.Stop.IsZero() {
-		return nil, errors.New("unbounded queries are not supported. Add a 'range' call to bound the query.")
+	for _, yield := range p.plan.Results {
+		if pr, ok := p.plan.Procedures[yield.ID]; ok {
+			if pr.Bounds.HasZero() {
+				return nil, errors.New("unbounded queries are not supported. Add a 'range' call to bound the query.")
```
nathanielc: Can you update this error message to indicate which result is unbounded using the result's name? The name is the map key.

query/plan/intersect_union_test.go
```diff
@@ -0,0 +1,505 @@
...
+				},
+				Stop: query.Time{
+					IsRelative: true,
+					Relative:   2 * time.Hour,
+				},
+			},
+		},
+		{
+			name: "one zero, one not",
+			a: plan.BoundsSpec{
```
nathanielc: Can you make this explicit? Meaning use `b: plan.BoundsSpec{}` and same for want.

query/plan/intersect_union_test.go
```diff
@@ -0,0 +1,505 @@
...
+					Absolute: time.Unix(100, 0),
+				},
+				Stop: query.Now,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.a.Union(tt.b, time.Now())
```
nathanielc: Instead of using `time.Now` here can you make the `now` time a property of the test?

query/plan/intersect_union_test.go
```diff
@@ -0,0 +1,505 @@
...
+				Stop: query.Time{
+					Absolute: time.Unix(200, 0),
+				},
+			},
+			wantExists: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, gotExists := tt.a.Intersect(tt.b, time.Now())
```
nathanielc: Same here do not use `time.Now` but make `now` be defined by the test explicitly. Using `time.Date` is a good way to write readable dates for test cases.

query/plan/intersect_union_test.go
```diff
@@ -0,0 +1,505 @@
...
+				Stop: query.Time{
+					Absolute: time.Unix(200, 0),
+				},
+			},
+			want: plan.BoundsSpec{
+				Start: query.Time{
+					Absolute: time.Unix(150, 0),
+				},
+				Stop: query.Time{
+					Absolute: time.Unix(200, 0),
```
nathanielc: For example this test would read better if time.Date was used instead of time.Unix.

query/plan/procedure.go
adamperlin: Ah ok. Thanks for the clarification. I'll add empty bounds back in.

adamperlin:
@nathanielc changes addressed, should be ready for a last look.

adamperlin:
@nathanielc done.

query/plan/procedure.go
```diff
@@ -101,20 +102,90 @@ type PushDownRule struct {
...
 
+var EmptyBoundsSpec = BoundsSpec{
+	Start: query.Now,
+	Stop:  query.Now,
+}
+
+// IsEmpty reports whether the given bounds
+// are equal to EmptyBoundsSpec, the specified 'empty' bounds.
+func (b BoundsSpec) IsEmpty(now time.Time) bool {
+	return b.Start == b.Stop || b.Start.Time(now).After(b.Stop.Time(now))
```
nathanielc: I think  this needs to be `b.Start.Time(now).Equal(b.Stop.Time(now)` Can you add a test for that case? I.e Start is relative now and Stop is absolute now.

adamperlin:
@nathanielc latest comments addressed. My apologies for forgetting that test case.

PR #621, opened 2018-08-13
refactor(iocounter): extract iocounter from the query package

jsternberg:
The iocounter package isn't specific to the query engine so it is being
removed from the query package and put into a `pkg` folder which is used
to contain utility packages that are useful, but don't make sense as
packaged with a specific area.


jsternberg:
This is not relevant anymore.

PR #622, opened 2018-08-13
cleanup(query): remove the idfile package

jsternberg:
It is not used anywhere and the entire project compiles with it removed.
It was probably used somewhere at some point and then stopped being
used.


jsternberg:
Not relevant anymore. This package seems to have been removed in both platform and flux by someone else, possibly when we transitioned flux to its own repository.

PR #623, opened 2018-08-13
cleanup(query): remove the query/tracing package

jsternberg:
Tracing queries is something we will need, but this package doesn't do
anything and isn't used so it would be better to add this when we start
working on it. Otherwise, it sends an incorrect message to programmers
who are looking at the directories.


jsternberg:
This doesn't appear to be relevant anymore.

Issue #620, opened 2018-08-13
Add erroring test cases to Flux tests


nathanielc:
We need to be able to assert that a query can error using our `functions/query_test.go` framework.

PR #627, opened 2018-08-14
Add errors as expected output in test framework

### LINKED ISSUE #533: Allow for expected errors in query test framework
Currently, the query test framework will fail if a query throws an error. This prevents writing tests for flux where errors are the expected output. Once #505 is merged (which affects the test framework significantly), this change should be made.

adamperlin:
Closes #533 

This is a very small change to the test framework. Instead of failing tests if they error, we compare the error message to our expected output file so we can deliberately test flux error handling.

_What was the problem?_
There was no way to expect errors as test ouput in `query_test.go`, which we should be able to do if we want to verify all flux functionality.

_What was the solution?_
Treat error output the same way as query output, and compare it against our expected data.

  - [x] Rebased/mergeable
  - [x] Tests pass


adamperlin:
Closes #620

query/functions/query_test.go
```diff
@@ -209,9 +203,11 @@ func QueryTestCheckSpec(t testing.TB, pqs query.ProxyQueryService, req *query.Pr
 
 	var buf bytes.Buffer
 	_, err := pqs.Query(context.Background(), &buf, req)
+	// If we have an error, we still want compare it against the .out.csv file.
+	// This allows us to expect errors as output.
 	if err != nil {
-		t.Errorf("failed to run query: %v", err)
-		return
+		buf.Reset()
```
aanthony1243: @nathanielc if results.Error() is not null, then the encoder will encode the error to buf.  However, pqs.Query()  will also return the value of results.Error().  This makes it hard for us to detect what type of error is being returned, and to know whether we should reset the buffer or let the encoder run as normal.  

@adamperlin   we may need to write a test-specific version of ProxyQueryServiceBridge the only returns errors resulting from attempting to run the query, or in running the encoder.  If the results contain an error, we'll let that go through naturally to the output and return a nil error.

query/functions/testdata/drop_after_rename.out.csv
```diff
@@ -1 +1 @@
-error: drop error: column "old" doesn't exist
+drop error: column "old" doesn't exist
```
aanthony1243: after above fix, we'll have to regenerate these.

query/functions/query_test.go
adamperlin: Ok, that makes sense.

adamperlin:
@aanthony1243 This ended up being a more involved change. After some digging, I realized that there was actually **no way** to differentiate between encoding errors and query execution errors. @nathanielc suggested adding a separate `EncodingError` type in order to fix this. `(*DelimitedMultiResultEncoder).Encode` checks the error type whenever it calls an encoding function, and encodes the result only if it `isn't` an `EncodingError`.

query/csv/result.go
```diff
@@ -688,7 +688,7 @@ func (e *ResultEncoder) Encode(w io.Writer, result query.Result) (int64, error)
 			}
 
 			if err := writeSchema(writer, &e.c, row, cols, tbl.Empty(), tbl.Key(), result.Name(), tableIDStr); err != nil {
-				return err
+				return query.NewEncoderError(err.Error())
```
aanthony1243: this looks great!  I had to double-check `query/influxql/result.go`  since there's a JSON encoder in there.  The good news is that that one is good to go, it only returns errors in the case of JSON encoding problems.  all other errors are embedded in the JSON.

query/result.go
```diff
@@ -157,10 +172,19 @@ type flusher interface {
 
 func (e *DelimitedMultiResultEncoder) Encode(w io.Writer, results ResultIterator) (int64, error) {
 	wc := &iocounter.Writer{Writer: w}
+
 	for results.More() {
 		result := results.Next()
 		if _, err := e.Encoder.Encode(wc, result); err != nil {
-			return wc.Count(), err
+			// If we have an error that's from
```
aanthony1243: on line 134 of the original source, there's a comment for the MultiResultEncoder interface that's now wrong.  Can you update it to say  "Returns the number of bytes written to w and any error resulting from the encoding process.  Errors obtained from the results object should be encoded to w and then discarded."

adamperlin: yep, will do.

adamperlin:
I decided to change another outdated comment for the `DelimitedMultiResultEncoder` as well.

query/result.go
```diff
@@ -157,10 +175,19 @@ type flusher interface {
...
+			// encoding specifically, return it
+			if _, ok := err.(*EncoderError); ok {
+				return wc.Count(), err
+			}
+			// Otherwise, the error is from query execution,
+			// so we encode it instead.
+			e.Encoder.EncodeError(wc, err)
```
nathanielc: We need to return any potential err from encoding the error here.

PR #631, opened 2018-08-14
Generated assets  normally ignored, consider them only when needed

fntlnz:
Related to the sprint item in the Chronograf's migration project, not related to any issue.

_What was the problem?_
`TODO.go` files are a problem when building locally because one generally have the generated files and this leads to conflicts in function names.
There's the need of a way to avoid considering `TODO.go` files when they are not needed.

_What was the solution?_
Ignore `TODO.go` files by default, only consider them when needed with the `ignoretodo` build tag like in CI.

  - [x] Rebased/mergeable
  - [ ] Tests pass


desa:
What happens if we clone the repo and run a bare `go install ./...` on a clone?

fntlnz:
@desa Thanks for the feedback, changed a bit the way it works now, here are all the cases I considered:

# Cases
## 1. New clone and no generated assets

- `make` works and tells the compiler to not use assets files but use the placeholder instead
- `go install ./...` or any `go` command line works because it uses the placeholder by default

## 2. Generated assets

- `make` works and tells the compiler to use the assets by passing the tag `assetsplaceholder` 
- `go install ./...` or any `go` command still works but still does not consider the assets by default, if one wants them can pass the mentioned tag OR use the makefile

## 3. Assets partially generated
Like if one generated only `canned` but not `server` and `dist`:

- `make` will break because it will use the tags since it finds a generated file but will fail because it does not have the definitions for the others. I think this is the right approach so that one can fix the inconsistency by seeing that a file is missing.
- The other normal `go` tools still work because they do not consider the assets by default

# Approach

The approach used to solve this was to abstract the generated function names in order to be able to create a file for each case `assets_placeholder.go` and `assets_generated.go` that are mutually excluded using the `assetsplaceholder` tag. The `assets_placeholder.go` are normally included while `assets_generated.go` are normally excluded. The makefile includes the tag to include `assets_generated.go` if it detects that assets are generated.

fntlnz:
We could also re-enable the `generate` target with this.

chronograf/canned/asset_placeholder.go
```diff
@@ -0,0 +1,18 @@
+// +build !assetsplaceholder
+
+package canned
+
+import "errors"
+
+// The functions defined in this file are placeholders
+// until we decide how to get the finalized Chronograf assets in platform.
```
fntlnz: Maybe this comment can be removed now

fntlnz:
@desa PTAL

chronograf/canned/asset_generated.go
```diff
@@ -0,0 +1,11 @@
+// +build assetsplaceholder
+
+package canned
+
+func GeneratedAsset(name string) ([]byte, error) {
```
desa: Does this rename cause an issue when we call `go generate ./...`?

fntlnz: It doesn't impact that from what I can see, have you had such experience?

fntlnz:
Rebased

fntlnz:
@desa PTAL again

fntlnz:
Closing in favor of https://github.com/influxdata/platform/pull/773

PR #632, opened 2018-08-14
Update run schema

### LINKED ISSUE #617: api: update schema for task Run
Runs should include task IDs.

We probably don't need `queuedAt`, as the scheduling implementation doesn't queue and pull; it executes as soon as a run-slot is free and the schedule is due to run. Maybe "lag" would be more appropriate (e.g. it was scheduled to run at 1 PM exactly but started running 100ms after). But it's not clear what lag would mean on retries or re-runs of historical time ranges.

I'm not sure what `startTime` and `stopTime` are, as they don't have a description in the swagger. I don't think `startTime` is the `now` value of a run, but we should report that.

jademcgough:
Closes #617

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -794,13 +794,13 @@ paths:
           schema:
             type: string
             format: date-time
-          description: filter runs to those queued after this time
+          description: filter runs to those scheduled after this time, RFC3339
         - in: query
           name: beforeTime
           schema:
             type: string
             format: date-time
-          description: filter runs to those queued before this time
+          description: filter runs to those scheduled before this time, RFC3339
       responses:
         '200':
           description: a list of task runs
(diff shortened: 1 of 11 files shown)


http/swagger.yml
```diff
@@ -1236,33 +1236,44 @@ components:
...
           enum: [
-            "queued",
+            "scheduled",
             "executing",
             "failed",
             "success"
           ]
-        queuedAt:
+        scheduledFor:
+          description: Time run is scheduled to begin, RFC3339.
```
mark-rushakoff: This description isn't quite right. "scheduled to begin" sounds like there's a correlation between wall time and this option, which isn't necessarily the case if the `delay` option is present or if it's a re-run. Maybe `Time used for run's "now" option` would be more accurate?

http/swagger.yml
```diff
@@ -1236,33 +1236,44 @@ components:
...
           format: date-time
         error:
           $ref: "#/components/schemas/Error"
         log:
           readOnly: true
-          description: A url to a relevant log.
+          description: Link to the full logs for a run.
           type: string
-      required: [queuedAt, status]
+      required: [scheduledFor]
```
mark-rushakoff: The user doesn't create runs directly, and I don't know how the `required` section affects the schema in swagger, so I'm not sure whether a `required` section makes sense here.

jademcgough: Agreed, I'm fine with removing this.

http/swagger.yml

jademcgough:
I'll get CI green before this is merged.

PR #633, opened 2018-08-14
Pivot() function 

### LINKED ISSUE #508: Implement Collate
See #455 for design

aanthony1243:
Closes #508 

_What was the problem?_
_What was the solution?_

  - [X] Rebased/mergeable
  - [X] Tests pass


query/functions/pivot.go
```diff
@@ -0,0 +1,366 @@
...
+	cache := execute.NewTableBuilderCache(a.Allocator())
+	d := execute.NewDataset(id, mode, cache)
+	t := NewPivotTransformation(d, cache, s)
+	return t, d, nil
+}
+
+type pivotTransformation struct {
+	d          execute.Dataset
+	cache      execute.TableBuilderCache
+	spec       PivotProcedureSpec
```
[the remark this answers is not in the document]
aanthony1243: a few functions already do this.  it saves a few keystrokes, and makes the data access more uniform.

query/functions/pivot_test.go
```diff
@@ -0,0 +1,149 @@
...
+						{execute.Time(2), 4.0, "m1", "f2"},
+					},
+				},
+			},
+			want: []*executetest.Table{
+				{
+					KeyCols: nil,
+					ColMeta: []query.ColMeta{
+						{Label: "_time", Type: query.TTime},
+						{Label: "value_m1_f1", Type: query.TFloat},
```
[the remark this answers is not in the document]
aanthony1243: good catch.  I forgot to run this test before I pushed.

query/functions/pivot.go
```diff
@@ -0,0 +1,366 @@
...
+	for colIDX, v := range tbl.Cols() {
+		if _, ok := colKeyIndex[v.Label]; !ok && v.Label != t.spec.ValueCol {
+			// the columns we keep are: group key columns not in the column key and row key columns
+			if tbl.Key().HasCol(v.Label) {
+				colMap[newIDX] = colIDX
+				newIDX++
+				keyCols = append(keyCols, tbl.Cols()[colIDX])
+				cols = append(cols, tbl.Cols()[colIDX])
+				keyValues = append(keyValues, tbl.Key().LabelValue(v.Label))
+			} else if _, ok := rowKeyIndex[v.Label]; ok {
```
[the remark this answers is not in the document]
aanthony1243: it's OK if rowKey and groupKey overlap since all of the rowKey columns will be retained in the output table.

query/functions/pivot.go
```diff
@@ -0,0 +1,366 @@
...
+		t.nextRow = 0
+	}
+
+	tbl.Do(func(cr query.ColReader) error {
+		for row := 0; row < cr.Len(); row++ {
+			rowKey := ""
+			colKey := t.spec.ValueCol
+			for j, c := range cr.Cols() {
+				if _, ok := rowKeyIndex[c.Label]; ok {
+					rowKey += valueToStr(cr, c, row, j)
```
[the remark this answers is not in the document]
aanthony1243: I think this one's ok.  it's clear in the code that since the key values are of different types, they must be marshalled into a single hashable type.

query/functions/pivot.go
```diff
@@ -0,0 +1,366 @@
...
+			// we know the col key;
+			//  0.  If we've not seen the colKey before, then we need to add a new column and backfill it.
+			if _, ok := t.colKeyMaps[newKey.String()][colKey]; !ok {
+				newCol := query.ColMeta{
+					Label: colKey,
+					Type:  valueColType,
+				}
+				builder.AddCol(newCol)
+				growColumn(builder, newCol.Type, t.nextCol, builder.NRows())
+				// don't forget to keep track of table width
```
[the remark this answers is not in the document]
aanthony1243: removed the comment instead.

query/functions/pivot_test.go
```diff
@@ -0,0 +1,149 @@
...
+						{execute.Time(2), 4.0, "m1", "f2"},
+					},
+				},
+			},
+			want: []*executetest.Table{
+				{
+					KeyCols: nil,
+					ColMeta: []query.ColMeta{
+						{Label: "_time", Type: query.TTime},
+						{Label: "_value_m1_f1", Type: query.TFloat},
```
nathanielc: I think this would be better as `m1_f1` instead of `_value_m1_f1`. 

Also is the `_` delimiter configurable by the user? If not that is fine. But we do need to specify that a `_` delimiter is used when concatenating the values.

query/functions/pivot_test.go
```diff
@@ -0,0 +1,149 @@
...
+						{Label: "_value_m1_f2", Type: query.TFloat},
+					},
+					Data: [][]interface{}{
+						{execute.Time(1), 1.0, 2.0},
+						{execute.Time(2), 3.0, 4.0},
+					},
+				},
+			},
+		},
```
nathanielc: Can you add a few more test cases?

I am interested in seeing test cases on:

*  the duplicate row behavior,
* where a column is dropped,
* and where the valueCol and/or colKey contain group key columns.


I'd also like to see the happy path test case where the defaults are used to transform raw _field _value data from a `from` call into a wide table.

query/functions/pivot_test.go
aanthony1243: _value  is referring to the source column from the pivot.   So if the column was named `mean`  we would have  `mean_m1_f1`.   It gives context to what's being stored and where it came from.  It does make it a bit repetitive though.

query/functions/pivot_test.go
aanthony1243: yes, I was planning on writing more tests.

query/functions/pivot.go
```diff
@@ -0,0 +1,389 @@
...
+	ValueCol string   `json:"value_col"`
+}
+
+var pivotSignature = query.DefaultFunctionSignature()
+
+var alignFieldsBuiltin = `
+// alignFields collects fields in each measurement with the same time stamp and aligns them in a single row. 
+// input requirements: table should have at minimum _time, _measurement, _field and _value columns and _measurement 
+// should be in the group key.  
+alignFields = (table=<-) => pivot(table: table, rowKey:["_time"], colKey: ["_field"], valueCol: "_value")
```
nathanielc: Should this helper have a call to `group` in order to ensure the data is grouped correctly?

query/functions/testdata/alignFields.in.csv
```diff
@@ -0,0 +1,28 @@
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string
```
nathanielc: Where is the `alignField.out.csv`?

PR #635, opened 2018-08-14
fix(query/control): fix the queueing metric

jsternberg:
We decremented all, but forgot to decrement queueing when it was set and
the query spec could not be validated.

### DIFF
--- a/query/control/controller_test.go
+++ b/query/control/controller_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/platform/query/mock"
 	"github.com/influxdata/platform/query/plan"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


query/control/controller_test.go
```diff
@@ -96,7 +152,7 @@ func TestController_EnqueueQuery_Failure(t *testing.T) {
 	}
 }
 
-func TestController_CancelQuery(t *testing.T) {
+func TestController_ExecuteDeadlock(t *testing.T) {
```
nathanielc: What is this test doing? Is it trying to test the deadlock we just found? If so what is it part of this PR?

jsternberg: I must have put that test code accidentally here. I'll remove it.

PR #641, opened 2018-08-15
Add query examples to flux SPEC

### LINKED ISSUE #546: Ensure every function in the query spec has an example
Every function should have a working example that is possible with the default telegraf/influxdb/fluxd setup described in the query readme file.

Many of these examples can be taken from the removed lines in the query readme. We're removing a lot of those examples in the readme so the readme itself is relatively easy to read, but we should either move the example to the query spec or create a new example that we know works.

It would also be good to ensure that the readme examples are included as query tests too if we can do that.

adamperlin:
Closes #546 

In the README, there were example queries for nearly every built-in. Those examples have been removed from the README, so they've been moved to the SPEC. Additional built-in examples which weren't in the README have also been added.

_What was the problem?_
We need query examples for those learning the language, and they need to live somewhere in this repo. 
_What was the solution?_
Use the SPEC for query examples.

  - [x] Rebased/mergeable
  - [x] Tests pass


query/docs/SPEC.md
```diff
@@ -935,29 +963,55 @@ Percentile has the following properties:
...
 For each aggregated column, it outputs the skew of the non null record as a float.
 
+Example:
+
+```
+from(db:"telegraf")
+    |> filter(fn: (r) => r["_measurement"] == "system" AND
+               r["_field"] == "uptime")
+    |> range(start:-12h)
+    |> sort(cols:["region", "host", "value"])
```
aanthony1243: not a skew example

query/docs/SPEC.md
```diff
@@ -988,21 +1042,48 @@ All selector operations have the following properties:
...
+Example:
+`from(db:"telegraf") |> range(start:-1m) |> group(by:["host"]) |> first()`
+
 ##### Last
 
 Last is a selector operation.
 Last selects the last non null record from the input table.
 
+Example:
+`from(db: "telegraf") |> last()`
```
aanthony1243: users will wonder why first() has the extra calls to range and group and this doesn't.   probably just make both consistent one way or the other.

query/docs/SPEC.md
```diff
@@ -988,21 +1042,48 @@ All selector operations have the following properties:
...
 Min selects the minimum record from the input table.
 
+Example: 
+
+```
+from(db:"telegraf")
+    |> filter(fn: (r) => r[ "_measurement"] == "cpu" AND r["_field" ] == "usage_system")
+    |> range(start:-12h)
+    |> window(every:10m, period: 5m)
+    |> min()
```
aanthony1243: same here. max and min are related.  I think it's best if their examples are consistently similar.

query/docs/SPEC.md
```diff
@@ -1034,6 +1125,17 @@ Filter has the following properties:
...
 
+Example: 
+
+```
+from(db:"telegraf")
+    |> filter(fn: (r) => r["_measurement"]=="cpu" AND
+                r["_field"] == "usage_system" AND
+                r["service"] == "app-server")
+    |> range(start:-12h)
+    |> max()
```
aanthony1243: drop max()  and put range before filter.  Mostly just to make the example clearer and more focused on the filtering.

query/docs/SPEC.md
```diff
@@ -1066,6 +1170,17 @@ Map has the following properties:
...
 
+Example:
+```
+from(db:"telegraf")
+    |> filter(fn: (r) => r["_measurement"]=="cpu" AND
+                r["_field"] == "usage_system" AND
+                r["service"] == "app-server")
+    |> range(start:-12h)
+    // Square the value
+    |> map(fn: (r) => r._value * r._value)
```
aanthony1243: let's add a second example where map creates a new table, such as: 
|> map(fn: (r) => {_time: r._time, app-server: r["service"]})

query/docs/SPEC.md
```diff
@@ -1084,6 +1199,14 @@ Range has the following properties:
     Specifies the exclusive newest time to be included in the results.
     Defaults to "now"
 
+Example:
+```
+from(db:"telegraf")
+    |> filter(fn: (r) => r["_measurement"] == "cpu" AND
+               r["_field"] == "usage_system")
+    |> range(start:-12h, stop: -15m)
```
aanthony1243: add a second example that uses date/time stamps.

query/docs/SPEC.md
```diff
@@ -1123,6 +1246,8 @@ Drop a list of columns: `drop(columns: ["host", "_measurement"])`
 
 Drop all columns matching a predicate: `drop(fn: (col) => col =~ /usage*/)`
 
+TODO(adamperlin): Add complete example
```
aanthony1243: finish todo

query/docs/SPEC.md
```diff
@@ -1193,6 +1331,9 @@ Examples:
     group(except:["_time", "region", "_value"]) // group records by all other columns except for _time, region, and _value
     group(by:[]) // group all records into a single group
     group(except:[]) // group records into all unique groups
+```
+from(db: "telegraf") |> range(start: -30m) |> group(by: ["host", "_measurement"])
```
aanthony1243: let's add an example using `except`  and put a comment with each example explaining what the grouping will be afterward.

query/docs/SPEC.md
```diff
@@ -1377,6 +1541,13 @@ Difference has the following properties:
 * `columns` list strings
     columns is a list of columns on which to compute the difference.
 
+```
+from(db: "telegraf")
+    |> range(start: -5m)
+    |> filter(fn: (r) => r._measurement == "cpu" and r._field == "usage_user")
+    |> difference()
+    |> limit(n: 1)
```
aanthony1243: drop limit

query/docs/SPEC.md
```diff
@@ -935,29 +963,48 @@ Percentile has the following properties:
...
+	|> percentile(p: 0.99)
+```
+
 ##### Skew
 
 Skew is an aggregate operation.
 For each aggregated column, it outputs the skew of the non null record as a float.
 
+Example:
+`from(db: "telegraf") |> range(start: -30m, stop: -15m) |> skew()`
```
aanthony1243: Add a filter line in here just like above for percentile.  This will get all fields from telegraf and not all of them are compatible with skew

query/docs/SPEC.md
```diff
@@ -935,29 +963,48 @@ Percentile has the following properties:
...
+Example:
+`from(db: "telegraf") |> range(start: -30m, stop: -15m) |> stddev()`
+
 ##### Sum
 
 Stddev is an aggregate operation.
 For each aggregated column, it outputs the sum of the non null record.
 The output column type is the same as the input column type.
 
+`from(db: "telegraf") |> range(start: -30m, stop: -15m) |> sum()`
```
aanthony1243: Same for spread, stddev and sum.  sorry I missed this the first time around.  Just add a filter to each query and these are all fine as-is.

query/docs/SPEC.md
aanthony1243: this comment moved somehow?  this sort example looks fine.

query/docs/SPEC.md
```diff
@@ -1194,6 +1364,20 @@ Examples:
...
+    |> group(by: ["host", "_measurement"])
+```
+All records are grouped by "host" and "_measurement".
+
+```
+from(db: "telegraf")
+    |> range(start: -30m)
+    |> group(except: ["_time"])
+```
+All records are grouped together by "host" and "_measurement", except not if they have the same "_time"
```
aanthony1243: All records are grouped by the set of all columns in the table, excluding "_time".  For example, if the table has columns ["_time", "host", "_measurement", "_field", "_value"]  then the group key would be ["host", "_measurement", "_field", "_value"]

query/docs/SPEC.md
```diff
@@ -1223,6 +1407,14 @@ Window has the following properties:
...
+```
+from(db:"telegraf")
+    |> range(start:-12h)
+    |> window(every:10m)
+    |> max()
+```
+
 [IMPL#319](https://github.com/influxdata/platform/query/issues/319) Remove concept of Bounds from tables
 
 #### Collate
```
aanthony1243: you will need to rebase with master to get my updates to the spec.

query/docs/SPEC.md
adamperlin: sounds good - thanks for catching.

query/docs/SPEC.md
```diff
@@ -1383,9 +1540,18 @@ Drop has the following properties:
...
+```
+from(db: "telegraf")
+	|> range(start: -5m)
+	|> drop(columns: ["host", "_measurement"])
+```
+Drop columns matching a predicate:
+```
+from(db: "telegraf")
+	|> range(start: -5m)
+    |> drop(fn: (col) => col =~ /usage*/)
```
nathanielc: The spacing is off here.

PR #643, opened 2018-08-15
refactor(task): add EffectiveCronString method to Options

mark-rushakoff:
The logic used to be only in the scheduler, but now Options exposes it.
This is another step towards making StoreTaskMeta schedule-aware to
enable creating manually requested runs.


task/options/options.go
```diff
@@ -125,3 +118,62 @@ func FromScript(script string) (Options, error) {
...
+	var errs []string
+	if o.Name == "" {
+		errs = append(errs, "name required")
+	}
+
+	cronPresent := o.Cron != ""
+	everyPresent := o.Every != 0
+	if cronPresent == everyPresent {
+		// They're both present or both missing.
+		errs = append(errs, "must specify exactly one of cron and every")
```
docmerlin: This error is a bit awkward, should probably be "must specify exactly one of either cron or every" using the English or instead of the logical or.  The current err uses the logical `and` which makes it a bit ambiguous in English.

task/options/options.go
```diff
@@ -125,3 +118,62 @@ func FromScript(script string) (Options, error) {
...
+		errs = append(errs, "retry must be at least 1")
+	} else if o.Retry > maxRetry {
+		errs = append(errs, fmt.Sprintf("retry exceeded max of %d", maxRetry))
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("invalid options: %s", strings.Join(errs, ", "))
```
docmerlin: 👍

task/options/options.go
mark-rushakoff: Updated to your suggested wording.

Issue #212, opened 2018-07-03
SPEC: add syntax for array and object literals


adamperlin:
From ifql created by [jsternberg](https://github.com/jsternberg) : influxdata/ifql#347  

There is no specification for what an array literal is and what an object literal is. I think this is also needed because object literals and blocks are opened using the same symbol so we need to differentiate how the two are different in the language syntax.

nathanielc:
Adding some details here:

Arrays are enclosed in `[]` and delimited with commas.

Objects are enclosed in `{}` and delimited with commas. 

Trailing commas are valid for both.

The object syntax has three variants:

Basic object literal
```
{
    foo: "bar",
    jim: "bob",
   every: 10s,
}
```

Implicit keys literal
```
{ sum, mean, count }
// the above is equivalent to 
{ sum: sum, mean: mean, count: count }
```

Keys as an expression:
```
var k = "mykey"

{ [k]: "value" }
// the above evaluates to this literal
{ "mykey": "value" }
```

nathanielc:
@pauldix Also suggested that we could diverge from the JS syntax and use string interpolation for key expressions. 

```
k = "mykey"
{ "{k}": "value"}
```

jlapacik:
This requires error handling to be implemented in the parser. I'm placing this back on the backlog and will pick back up when https://github.com/influxdata/flux/issues/305 is added.

PR #646, opened 2018-08-15
design spec for to function

jlapacik:
Addresses this issue https://github.com/influxdata/idpe/issues/1382
Addresses this PR https://github.com/influxdata/idpe/pull/1410

This commit proposes a new design for the `to` operation.

_What was the problem?_

The current `to` operation doesn't write its data points correctly when its input table contains a `_field` column. In particular, calling

    from(bucket:"my-bucket") |> range(start:-1h) |> to(bucket:"new-bucket", org:"my-org")

should write the exact same data points from my-bucket to new-bucket. Currently it does not.

_What was the solution?_

Propose a new function signature and design for the `to` function such that the default behavior is as specified above, while at the same time giving the user maximum control for how they want to write out their data.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/docs/SPEC.md
+++ b/query/docs/SPEC.md
@@ -685,6 +685,7 @@ Details about their arguments and behavior can be found in the Operations sectio
 * min
 * percentile
 * range
+* to
 * sample
 * set
 * shift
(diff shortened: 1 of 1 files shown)


query/docs/SPEC.md
```diff
@@ -1401,6 +1402,65 @@ Shift has the following properties:
...
 
+
+#### To
+
+The To operation takes data from a stream and writes it to a bucket.
+To has the following properties:
+
+* `bucket` string  
+    The bucket to which data will be written.
+* `org` string
```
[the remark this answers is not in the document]
jlapacik: I am behind on that as well but we can definitely add any such options if necessary. @pauldix  will probably be able to give more insight.

query/docs/SPEC.md
```diff
@@ -1401,6 +1402,65 @@ Shift has the following properties:
...
+    Function that takes a record from the input table and returns an object.  
+    For each record from the input table `fn` returns on object that maps output field key to output value
+
+| Property | Required | Default |
+| -------- | -------- | ------- |
+| bucket        | Yes | N/A |
+| org           | Yes | N/A |
+| time_column   | No | "_time" |
+| tag_columns   | No | All columns of type string excluding all value columns and the _field column if present |
+| fn            | No | (r) => return { r._field: r._value } |
```
nathanielc: The function syntax needs to be `(r) => return { [r._field]: r._value }` See https://github.com/influxdata/platform/issues/212

query/docs/SPEC.md
```diff
@@ -1401,6 +1402,65 @@ Shift has the following properties:
...
+    The bucket to which data will be written.
+* `org` string  
+    The organization ID of the above bucket.
+* `time_column` string  
+    The time column of the output
+* `tag_columns` list of strings  
+    The tag columns of the output
+* `fn` function(record) object  
+    Function that takes a record from the input table and returns an object.  
+    For each record from the input table `fn` returns on object that maps output field key to output value
```
nathanielc: Should this be called the `fieldFn` or something, to make it clear that the key value tuples returned are always field values?

query/docs/SPEC.md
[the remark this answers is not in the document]
nathanielc: Writing to another cluster is an additive change, i.e. new parameter will be added to the function. I am OK with that being out of scope for now. But if @pauldix  wants us to spec that right now we can.

query/docs/SPEC.md
```diff
@@ -1401,6 +1402,64 @@ Shift has the following properties:
...
+* `timeColumn` string  
+    The time column of the output.  
+    **Default:** `"_time"`
+* `tagColumns` list of strings  
+    The tag columns of the output.  
+    **Default:** All columns of type string, excluding all value columns and the `_field` column if present.
+* `fieldFn` function(record) object  
+    Function that takes a record from the input table and returns an object.  
+    For each record from the input table `fieldFn` returns on object that maps output field key to output value.  
+    **Default:** `(r) => return { [r._field]: r._value }`
```
nathanielc: I missed it the first time but this does not need the `return` keyword and should be `(r) => ({ [r._field]: r._value })`

jlapacik:
I have added optional `host` and `token` parameters to the function signature.

PR #647, opened 2018-08-16
spec outlining packages, namespaces, and imports

### LINKED ISSUE #510: Design how imports/namespaces work in Flux
The current SPEC does not define how import works.

Things in scope:

* Add syntax for `imports` into Flux
* Import builtin packages
* Using the imported package as a "namespace"
* What does the import path mean?

Things not in scope:

* Given an external import path, how do I get that package?
* Importing external packages
* Defining external packages
* Package versioning

Things to explore:

* Do we need to support major version numbers in the import path?

jlapacik:
Closes #510 

This commit specifies how packages, namespaces, and imports work in Flux

_What was the problem?_

Flux currently doesn't have a notion of packages or a way to import external Flux source code.

_What was the solution?_

Propose a design for packages and namespaces as well as the behavior of package imports.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/docs/SPEC.md
+++ b/query/docs/SPEC.md
@@ -94,7 +94,7 @@ Examples:
 The following keywords are reserved and may not be used as identifiers:
 
     and    import  not  return
-    empty  in      or
+    empty  in      or   package
 
 [IMPL#256](https://github.com/influxdata/platform/issues/256) Add in and empty operator support  
 [IMPL#334](https://github.com/influxdata/platform/issues/334) Add "import" support  
@@ -456,7 +456,7 @@ Flux is lexically scoped using blocks:
 1. The scope of an option identifier is the options block.
 2. The scope of a preassigned (non-option) identifier is in the universe block.
 3. The scope of an identifier denoting a variable or function at the top level (outside any function) is the package block.
-4. The scope of a package name of an imported package is the file block of the file containing the import declaration.
+4. The scope of the name of an imported package is the file block of the file containing the import declaration.
 5. The scope of an identifier denoting a function argument is the function body.
 6. The scope of a variable assigned inside a function is the innermost containing block.
 
@@ -465,8 +465,6 @@ While the identifier of the inner assignment is in scope, it denotes the entity 
 
 Option identifiers have default assignments that are automatically defined in the _options block_.
 Because the _options block_ is the top-level block of a Flux program, options are visible/available to any and all other blocks.
-However option values may only be reassigned or overridden in the explicit block denoting the main (executable) package.
-Assignment of option identifiers in any non-executable package is strictly prohibited.
 
 The package clause is not a assignment; the package name does not appear in any scope.
 Its purpose is to identify the files belonging to the same package and to specify the default package name for import declarations.
@@ -569,12 +567,66 @@ Examples:
     baz = (y=<-) => // function body elided
     foo() |> bar() |> baz() // equivalent to baz(x:bar(y:foo()))
 
+### Program
+
+A Flux program is a sequence of statements defined by
+
+    Program = [PackageStatement] [ImportList] StatementList .
+    ImportList = { ImportStatement } .
+
 ### Statements
 
 A statement controls execution.
 
-    Statement = OptionStatement | VarAssignment | ReturnStatement |
-                ExpressionStatement | BlockStatment .
+    Statement = OptionStatement | VarAssignment |
+                ReturnStatement | ExpressionStatement | BlockStatment .
+
+#### Package statement
+
+    PackageStatement = "package" identifier .
+
+A package statement defines a package block.
+Package names must be valid Flux identifiers.
+The package statement must be the first statement of every Flux source file.
+If a file does not declare a package statement, all identifiers in that file will belong to the special _main_ package.
+
+##### package main
+
+The _main_ package is special for a few reasons:
+
+1. It defines the entrypoint of a Flux program
+2. It cannot be imported
+3. All query specifications produced after evaluating the _main_ package are coerced into producing side effects
+
+#### Import statement
+
+    ImportStatement = "import" [identifier] `"` unicode_char { unicode_char } `"`.
+
+Associated with every package is a package name and an import path.
+The import statement takes a package's import path and brings all of the identifiers defined in that package into the current scope.
+The import statment defines a namespace through which to access the imported identifiers.
+By default the identifer of this namespace is the package name unless otherwise specified.
+For example, given a variable `x` declared in package `foo`, importing `foo` and referencing `x` would look like this:
+
+```
+import "import/path/to/package/foo"
+
+foo.x
+```
+
+Or this:
+
+```
+import bar "import/path/to/package/foo"
+
+bar.x
+```
+
+A package's import path is always absolute.
+Flux does not support relative imports.
+Assigment into the namespace of an imported package is not allowed.
+A package cannot access nor modify the identifiers belonging to the imported packages of its imported packages.
+Every statement contained in an imported package is evaluated.
 
 #### Option statements
 
@@ -650,6 +702,15 @@ Examples:
     1 + 1
     f()
     a
+
+### Side Effects
+
+Side effects can occur in two ways.
+
+1. By reassigning builtin options
+2. By calling a function that produces side effects
+
+A function produces side effects when it is explicitly declared to have side effects or when it calls a function that itself produces side effects.
 
 ### Built-in functions
 
@@ -926,12 +987,13 @@ Examples:
 The execution of a query is separate and distinct from the execution of Flux the language.
 The input into the query engine is a query specification.
 
-The output of an Flux program is a query specification, which then may be passed into the query execution engine.
+The output of a Flux program is a query specification, which then may be passed into the query execution engine.
 
 ### Query specification
 
 A query specification consists of a set of operations and a set of edges between those operations.
 The operations and edges must form a directed acyclic graph (DAG).
+A query specification produces side effects when at least one of its operations produces side effects.
 
 #### Encoding
 
@@ -1047,6 +1109,8 @@ Most operations output one table for every table they receive from the input str
 
 Operations that modify the group keys or values will need to regroup the tables in the output stream.
 
+An operation produces side effects when it is constructed from a function that produces side effects.
+
 ### Built-in operations
 
 #### From
@@ -1095,6 +1159,8 @@ Yield has the following properties:
 
 Example:
 `from(bucket: "telegraf/autogen") |> range(start: -5m) |> yield(name:"1")`
+
+**Note:** The `yield` function produces side effects.
 
 #### Aggregate operations
 
@@ -2140,6 +2206,8 @@ _tag1=a hum=55.3,temp=100.1 0005
 _tag1=a hum=55.4,temp=99.3 0006
 _tag1=a hum=55.5,temp=99.9 0007
 ```
+
+**Note:** The `to` function produces side effects.
 
 #### Type conversion operations
 


query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
-[IMPL#317](https://github.com/influxdata/platform/query/issues/317) Add package/namespace support
+### Packages and namespaces
+
+Packages group identifier assignments into namespaces.
+Access to an identifier assigned in an external package requires referencing that identifier's namespace.
+Access to an identifier assigned in the current package does not require any such reference.
+
+A package defines a namespace via the package statement
+
+    package name
```
nathanielc: Can you define this statement using the EBNF grammar? Then provide and example like this as well.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+
+Here _name_ defines the namespace for all identifiers in the this package.
+Any identifier defined in this package will be referenced via the syntax _name_._identifier_ in all external packages.
```
nathanielc: This is not true, when importing the package the namespace can be renamed in the context of the importing namespace. So this is true only most of the time.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+That is, only the _main_ package can be executed.
+
+Package names must be valid Flux identifiers.
+
+### Importing packages
+
+Associated with every package is a namespace and a path.
+Namespaces can be imported by another package using that package's path and the import statement as follows:
+
+    import "path/to/package"
```
nathanielc: Same here, can you define the import statement using the EBNF grammar? Then include this example.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+
+Once imported, variables and other identifiers defined in the imported package may be accessed using their namespace and the dot operator.
+For example, given a variable `x` declared in the package `foo`, importing `foo` and referencing `x` would look like this:
+
+```
+import "path/to/package/foo"
+
+variable = foo.x
```
nathanielc: I think this example would be more clear without the assignment.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+```
+
+A package's import path is always absolute.
+Flux does not support relative imports.
+The _main_ namespace is the only namespace that cannot be imported.
+Assignment cannot be made inside of an imported namespace.
+Namespaces are containers for identifier assignments only.
```
nathanielc: I am not sure what this means.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+Because the import statement is not an assignment, a package cannot access nor modify the imported namespaces of its imported namespaces.
```
nathanielc: Lets not state why, but only what. i.e. `A package cannot access nor modify the imported namespaces of its imported namespaces.`

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+An imported namespaces can be renamed via the syntax:
+
+    import new_name_for_foo "path/to/package/foo"
```
nathanielc: Make sure to include this option in the import statement grammar.

query/docs/SPEC.md
```diff
@@ -414,7 +414,53 @@ The package clause is not a assignment; the package name does not appear in any
...
+
+A package defines a namespace via the package statement
+
+    package name
+
+Here _name_ defines the namespace for all identifiers in the this package.
+Any identifier defined in this package will be referenced via the syntax _name_._identifier_ in all external packages.
+
+All identifer assignments in Flux must occur inside of a package.
+The package statement must be the first line of every Flux source file.
```
pauldix: I assume this isn't true for something that is just a one-off script that is meant to be run as `main`?

jlapacik:
I don't think we want to specify how an import path gets resolved just yet. This spec is about namespaces and what their import behavior should be. So I think then a package should be defined as a 2-tuple of (namespace, import path). Where the import path comes from and how it gets resolved will be decided later.

pauldix:
@jlapacik ok that makes sense, but that surfaces another question for me. This might only be applicable to imports from supported SCMs (like github), but I was thinking that the import should include a version. Like:

```
import "github.com/pauldix/a", ">=1.0.0"
import "github.com/pauldix/b", "1.1.2"
import "github.com/pauldix/c", "master"
import "github.com/pauldix/d", "ce6195150a7931844acea48d531b212c35be86ed"
```

Might also make sense for imports from specific paths. But then we'd need a version in the package definition, like:

```
package foo "0.1.0"
```

Of course, that puts it on the package developer to increment their version and be responsible. It's how RubyGems did things back in the day.

jlapacik:
@pauldix are you saying you want package versions to be required? As in every package statement must have a version associated with it?

pauldix:
@jlapacik that was my initial thought. But it's pretty crazy and out there so I'm interested in hearing what others think of it. Best idea or worst idea?

pauldix:
Although it would be annoying to have to increment the version in every file in a package. So that would be something that we'd want to be able to declare once.

nathanielc:
> Although it would be annoying to have to increment the version in every file in a package. So that would be something that we'd want to be able to declare once.

This is my biggest concern with versions being part of the import statement. That is why typically there is an external file the indicates which versions of the things need to be used. Then that file extends to all imports in the "package", but if a package is a single file it is not that important anymore. 

For now though I want to leave versions out of scope for this PR.  We do not plan to support external imports yet, only builtin packages will be importable. 

When we go to specify how external imports will work, we will need to figure out versions at that point as well.

pauldix:
@nathanielc sounds good, let's leave this one for internal packages. Helps so we can start to get those built and move things around. We can have the nerd fight later on versioning and external packages :)

query/docs/SPEC.md
```diff
@@ -573,8 +570,57 @@ Examples:
...
+The package import path is used by external packages to import and access the identifiers defined in a package.
+
+Access to an identifier assigned in an external package requires referencing the name of the package containing said identifier.
+Access to an identifier assigned in the current package does not require any such reference.
+
+The package statement must be the first line of every Flux source file.
+If a file does not declare a package statement, all identifiers in that file will belong to the special _main_ package.
+The _main_ package cannot be imported.
+The _main_ package defines the entrypoint of a Flux program.
+That is, only the _main_ package can be executed.
```
nathanielc: What does it mean to be `executed`? I think this needs some more details around side effects etc. Maybe just reference the `main` package here but then in the scopes sections talk about how the `main` package is different.

query/docs/SPEC.md
```diff
@@ -573,8 +570,57 @@ Examples:
...
+
+PackageStatement = "package" identifier .
+
+A package statement defines a package block.
+A package has a name and an import path.
+The package name acts as a namespace for all identifiers defined in that package.
+The package import path is used by external packages to import and access the identifiers defined in a package.
+
+Access to an identifier assigned in an external package requires referencing the name of the package containing said identifier.
+Access to an identifier assigned in the current package does not require any such reference.
```
nathanielc: These statements feel odd. 

Why do I have to state this?
>Access to an identifier assigned in the current package does not require any such reference.

Shouldn't that be assumed already based on the discussion about identiers and scopes?

> Access to an identifier assigned in an external package requires referencing the name of the package containing said identifier.

This statement is misleading since it won't always be the package name that is used to identify the package when imported.

 Maybe how you access identifiers from within a package should be part of the import statement discussion. Then you can state that access happens by using the package name by default or the specific 'as' name if provided.

query/docs/SPEC.md
```diff
@@ -573,8 +570,57 @@ Examples:
...
+
+Package names must be valid Flux identifiers.
+
+#### Import statement
+
+ImportStatement = "import" [identifier] `"` unicode_char { unicode_char } `"`.
+
+Packages can be imported by other packages and their identifiers accessed via the imported package's name.
+This is achieved through the import statement.
+The import statement also allows renaming of the imported package, in which case all identifiers will be reference by the new package name.
```
nathanielc: This sounds great. Let's be careful to not make it sound like we are renaming the package itself. I see the package name and the access identifier as separate things, with the exception that the default access identifier is the package name.

jlapacik:
I added a sub-section to the package statement section describing the main package. I also added a tiny section on side effects and which statement produce them.

query/docs/SPEC.md
```diff
@@ -572,8 +570,55 @@ Examples:
 
 A statement controls execution.
 
-    Statement = OptionStatement | VarAssignment | ReturnStatement |
-                ExpressionStatement | BlockStatment .
+    Statement = PackageStatement | ImportStatement | OptionStatement |
```
nathanielc: It needs to be the first Statement correct? Then it cannot belong in the Statement production rule as that means it can appear anywhere. You will need to update the Program production rule to have an optional PackageStatement.

query/docs/SPEC.md
```diff
@@ -572,8 +570,55 @@ Examples:
...
+                VarAssignment | ReturnStatement | ExpressionStatement | BlockStatment .
+
+#### Package statement
+
+    PackageStatement = "package" identifier .
+
+A package statement defines a package block.
+Package names must be valid Flux identifiers.
+The package statement must be the first line of every Flux source file.
```
nathanielc: Should we say first `line` or first `statement`? Meaning do we care about extra whitespace at the beginning of the file?

query/docs/SPEC.md
```diff
@@ -572,8 +570,55 @@ Examples:
...
+If a file does not declare a package statement, all identifiers in that file will belong to the special _main_ package.
+
+##### package main
+
+The _main_ package is special for a few reasons:
+
+1. It defines the entrypoint of a Flux program
+2. It cannot be imported
+3. All composite query operations defined in the _main_ package are coerced into producing side effects
```
nathanielc: I like this phrasing `are coerced into producing side effects` what are `composite query operations`? Do we define that anywhere else in the spec? If not lets give a name to functions that produce query operations. Then we can refer to it here.

I think in the section about `Query Spec` we should talk about side effects and give a name to functions that produce a query operation.

query/docs/SPEC.md
```diff
@@ -572,8 +570,55 @@ Examples:
...
+
+```
+import bar "import/path/to/package/foo"
+
+bar.x
+```
+
+A package's import path is always absolute.
+Flux does not support relative imports.
+Assignment cannot be made inside of an imported package.
```
nathanielc: This bit kind of reads like packages are not allowed to have assignment statements. Maybe we should say `Assigment into an imported package identifier is not allowed`?

query/docs/SPEC.md
```diff
@@ -650,6 +695,15 @@ Examples:
     f()
     a
 
+### Side Effects
+
+Side effects can occur in two ways in Flux.
```
nathanielc: `Side effects can occur in two ways`.

query/docs/SPEC.md
```diff
@@ -650,6 +695,15 @@ Examples:
...
+
+1. By reassigning builtin options
+2. By calling a function that produces side effects
+
+In Flux a function produces side effects if it is explicitly declared to have side effects or if it calls a function that itself produces side effects.
```
nathanielc: Remove the `In Flux` bits.

nathanielc: Also ` produces side effects if it` -> ` produces side effects when it`. I like using `when` instead of `if` because it reads more like statement of fact instead of a procedure for how to do something. Just a small nit.

query/docs/SPEC.md
jlapacik: Yes should be first statement.

query/docs/SPEC.md
```diff
@@ -441,6 +441,10 @@ In addition to explicit blocks in the source code, there are implicit blocks:
 4. Each file has a _file block_ containing all Flux source text in that file.
 5. Each function literal has its own _function block_ even if not explicitly declared.
 
+**Note:** The syntax of a _file block_ is defined as:
+
+    FileBlock = [PackageStatement] StatementList
```
nathanielc: I know we refer to the file block here, but lets maybe just call this the Program rule? And the Program rule should look like this:

```
    Program = [PackageStatement] [ImportList] StatementList
```

Where the ImportStatement is not part of the normal sets of statements. Meaning all import must come before any other statements but after the package statement.

jlapacik: Updated.

PR #648, opened 2018-08-16
Using query request struct to query resources

### LINKED ISSUE #594: Use query.Request struct in sources
Currently to query sources, we use a custom query structure. This should be updated to be the request structure defined here:

https://github.com/influxdata/platform/blob/846455a487f074d35b21da51365d921d99d1739d/query/query.go#L83-L95

fntlnz:
TODO:
- [x] Use the query.Request struct for query requests to sources
- [x] Add Flux and spec mappings 
- [x] SourceQueryService skeleton 
- [x] Implement the `Query` function in the `SourceQueryService` to proxy to the external sources
~- [ ] Debug on `query/query.go:203` - dialect non taken from the request body for some reason.~
~- [ ] Check if the reworked external query handler still works and respect requirements `http/external_query_handler.go:44`~
~- [ ] Support for influxql mappingsnot yet done `http/source_service.go:94`~
- [ ] Work on the specific ProxyQueryServices as summarized [here](https://github.com/influxdata/platform/pull/648/commits/fbe18aa14181bd8c7e558164b271c0a746398ee9)

Closes #594

  - [x] Rebased/mergeable
  - [x] Tests pass


fntlnz:
I think that an early review on the overall design from you, @nathanielc would be useful on this.

http/query_service.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	queryPath = "/v1/query"
+	queryPath = "/query"
```
nathanielc: This is not correct. The query service is expected to exist on `/v1/query`. This path is internal and not the external path.

source/source.go
```diff
@@ -1,8 +1,12 @@
-package platform
```
nathanielc: Why does source need its own package? It seems small enough to live at the top level.

http/source_query_service.go
```diff
@@ -0,0 +1,22 @@
...
+
+// TODO(fntlnz): remove this before merge
+import (
+	"context"
+
+	"github.com/influxdata/platform/query"
+	"github.com/influxdata/platform/source"
+)
+
+type SourceQueryService struct {
```
nathanielc: Can you add a comment here explaining what this is? It looks to me like this struct implements the QueryService by consuming a Source and assuming the source can be contacted via HTTP?

The QueryService is currently an internal only service. The ProxyQueryService is the only publicly exposed service, so its likely that this service needs to use the ProxyService instead? I am not sure what the overall design is here so I may be confusing things.

http/source_service.go
```diff
@@ -63,14 +67,18 @@ type SourceHandler struct {
...
 }
 
 // NewSourceHandler returns a new instance of SourceHandler.
 func NewSourceHandler() *SourceHandler {
 	h := &SourceHandler{
-		Router: httprouter.New(),
-		Logger: zap.NewNop(),
+		Router:           httprouter.New(),
+		Logger:           zap.NewNop(),
+		CompilerMappings: query.CompilerMappings{},
```
nathanielc: Mapping should be set in the `main.go` of the binary, exposing what kind of queries you want to handle.

source/source.go
fntlnz: had some import cycles that have been lately solved, will put it back under http  if I don't reintroduce import cycles again

fntlnz:
@nathanielc the goal is to call query sources using the builtin structures in query instead of custom ones, see the issue: https://github.com/influxdata/platform/issues/594

fntlnz:
This is the kind of http call I'm doing to platform's api to trigger the query endpoint for sources:

```
curl -X POST \
  http://127.0.0.1:9999/v2/sources/027dc84508cf1000/query \
  -d '{
    "compiler_type": "flux",
    "compiler": {
    	"dialect_type": "csv",
        "dialect" : {
          "header": true,
          "annotations": ["datatype"]
        },
        "query": "from(db: \"testmod\") |> range(start: -50m)"
    }
}'
```

http/query_service.go
fntlnz: Yes you're right, also I don't need this anymore, will revert this change

fntlnz:
The request I do to idpd now looks like this: 

```
curl -X POST \
  http://127.0.0.1:9999/v2/sources/027dc84508cf1000/query \
  -H 'Content-Type: application/json' \
  -d '{
	"request": {
	    "compiler_type": "flux",
    	"compiler": {
        	"query": "from(db: \"testmod\") |> range(start: -10d)"
		}	
	},
    "dialect_type": "csv",
    "dialect" : {
		"noheader": false,
		"annotations": ["datatype"]
	}
}'
```

fntlnz:
@desa explained me some good parts of this today since I didn't understand the specific differences of v1 (influxd + flux) vs v1 (only influxd) and v2 (external and self). I summarized all the explainations in comments under this commit (https://github.com/influxdata/platform/pull/648/commits/fbe18aa14181bd8c7e558164b271c0a746398ee9)

fntlnz:
Worked a bit more with @desa we now have most of the use cases working

- v1 with flux
- v1 with influxql
- v2 with influxql


- v1 with spec (should be easy as using the right compiler)
- v2 with flux
- self (for now will be the same as v2 using localhost)

http/source_proxy_service.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type SourceProxyQueryService struct {
+	InsecureSkipVerify bool
+	URL                string
+	source.SourceFields
+}
+
+func (s *SourceProxyQueryService) Query(ctx context.Context, w io.Writer, req *query.ProxyRequest) (int64, error) {
+	switch req.Request.Compiler.(type) {
```
desa: use `req.Request.Compiler.CompilerType()`

fntlnz: use `req.Request.Compiler.CompilerType()` here instead

http/influxdb/source_proxy.go
```diff
@@ -0,0 +1,104 @@
...
+type SourceProxyQueryService struct {
+	InsecureSkipVerify bool
+	URL                string
+	source.SourceFields
+	source.V1SourceFields
+}
+
+func (s *SourceProxyQueryService) Query(ctx context.Context, w io.Writer, req *query.ProxyRequest) (int64, error) {
+	switch req.Request.Compiler.CompilerType() {
+	case "flux":
```
fntlnz: Avoid using string literals here

http/source_proxy_service.go
```diff
@@ -0,0 +1,97 @@
...
+		return 0, err
+	}
+	defer resp.Body.Close()
+	if err := CheckError(resp); err != nil {
+		return 0, err
+	}
+	return io.Copy(w, resp.Body)
+}
+
+func (s *SourceProxyQueryService) queryInfluxQL(ctx context.Context, w io.Writer, req *query.ProxyRequest) (int64, error) {
```
desa: Should error if not valid response format.

http/source_proxy_service.go
```diff
@@ -0,0 +1,97 @@
...
+	compiler := req.Request.Compiler.(*influxql.Compiler)
```
fntlnz: Check if this panics

http/influxdb/source_proxy.go
```diff
@@ -0,0 +1,104 @@
...
+	hreq, err := http.NewRequest("POST", u.String(), nil)
+	if err != nil {
+		return 0, err
+	}
+
+	// TODO(fntlnz): configure authentication methods username/password and stuff
+	hreq = hreq.WithContext(ctx)
+
+	params := hreq.URL.Query()
+	compiler := req.Request.Compiler.(*influxql.Compiler)
```
desa: add `ok` check here

fntlnz:
This is ready for another review @goller @nathanielc ! Thanks :tada:

source/source.go
goller: @fntlnz I'd rather this stay as top level package

http/source_proxy_service.go
```diff
@@ -0,0 +1,97 @@
...
+	var body bytes.Buffer
+	if err := json.NewEncoder(&body).Encode(req); err != nil {
+		return 0, err
+	}
+
+	hreq, err := http.NewRequest("POST", u.String(), &body)
+	if err != nil {
+		return 0, err
+	}
+	hreq.Header.Set("Authorization", s.Token)
```
goller: I think there is a helper function to set this header, but, it needs "Token :" + s.Token

http/external_query_handler.go
```diff
@@ -25,97 +24,42 @@ type ExternalQueryHandler struct {
...
-				return fmt.Errorf(`unknown dialect annotation type: %s`, a)
-			}
-		}
-
-		switch request.Dialect.DateTimeFormat {
-		case "RFC3339", "RFC3339Nano":
-		default:
-			return fmt.Errorf(`unknown dialect date time format: %s`, request.Dialect.DateTimeFormat)
+	default:
+		orgName := r.FormValue("organization")
```
goller: This should be a query parameter and not a multipart form.  the body of this message is a JSON body and should not both JSON and the parameters in a multipart way.

cmd/idpd/main.go
```diff
@@ -189,6 +193,10 @@ func platformF(cmd *cobra.Command, args []string) {
 
 		sourceHandler := http.NewSourceHandler()
 		sourceHandler.SourceService = sourceSvc
+		query.AddCompilerMappings(sourceHandler.CompilerMappings)
+		influxql.AddCompilerMappings(sourceHandler.CompilerMappings, mock.NewDBRPMappingService())
```
goller: Shoudl this be a mock?

fntlnz: Yes, the compiler does need a dbprp mapping service for deserialization but after it will not be used since this is not the point where dbrp is applied.

Maybe we could write a specific dummy mapping service for here or just avoid the need to do that in deserialization.

Open to ideas

http/external_query_handler.go
fntlnz: This is only triggered when the message is not a `application/json`. The code is not original from me, if you look at the diff it is the same behavior as it was before my change, I only changed the `application/json` behavior to support the `query.ProxyRequest` struct.

goller:
Ok, I’d like this to be switched to just the query parameter

cmd/idpd/main.go
goller: Just put a TODO to fix and file the issue.  I'm ok with it temporarily.

http/influxdb/source_proxy.go
```diff
@@ -0,0 +1,104 @@
...
+	source.SourceFields
+	source.V1SourceFields
+}
+
+func (s *SourceProxyQueryService) Query(ctx context.Context, w io.Writer, req *query.ProxyRequest) (int64, error) {
+	switch req.Request.Compiler.CompilerType() {
+	case influxql.CompilerType:
+		return s.influxQuery(ctx, w, req)
+	case query.FluxCompilerType:
+		return s.fluxQuery(ctx, w, req)
```
nathanielc: This part is confusing to me. The idea behind the `query.Request` type is that a single implementation can service both by using the Compiler interface. I get that the Source proxies to a backend implementation, but why does that implementation not also use the query.Request type?

http/external_query_handler.go
```diff
@@ -25,97 +24,42 @@ type ExternalQueryHandler struct {
...
-			Header         *bool    `json:"header"`
-			Delimiter      string   `json:"delimiter"`
-			CommentPrefix  string   `json:"commentPrefix"`
-			DateTimeFormat string   `json:"dateTimeFormat"`
-			Annotations    []string `json:"annotations"`
-		}
-	}{}
-
+func decodeQueryRequest(r *http.Request, req *query.ProxyRequest, orgSvc platform.OrganizationService, cmap query.CompilerMappings, dmap query.DialectMappings) error {
+	req.WithCompilerMappings(cmap)
```
nathanielc: I am having a hard time following all the changes to the API, can you update the swagger doc for these changes so I can see how it impacts the API?

http/influxdb/source_proxy.go
fntlnz: Because in this way the user can pass a `dialect` and a `dialect_type` so that we can pass it to the backend service or reformat the data ourselves. For now we are just informing the backend service about that and not reformatting anything.

fntlnz:
@nathanielc 

> Change the public/external query API to facilitate this change. Its not clear how the API is changing other than it seems to now be using the query.Request type.

It is not using `query.Request` but the goal is to use `query.ProxyRequest` which contains a `query.Request`


> What backend services are being abstracted behind a source?

Influxql and flux backends for both v1 and v2.

> Why does implementing the SourceProxyQueryService require different backends instead of using the query.Request down through the layers.

Because there's no unified endpoint to make the calls against.

For example for `v1` you need to use `source.URL` if the `compiler_type` is `influxql` and you will use `influxdb`as backend but if you specify `flux` as `compiler_type` it will use `source.V1SourceFields.FluxURL` and call a `fluxd` daemon instead.


> Why is there no SourceProxyQueryService interface when I see at least two implementations of it? Maybe seeing this interface would help me understand what is being changed.

Because it implements the `query.ProxyQueryService` interface already and only that

https://github.com/influxdata/platform/blob/3e7a04e88718ba8b2a4dd130e96f2ebcb7c05d9e/query/query.go#L30-L34

nathanielc:
>Influxql and flux backends for both v1 and v2.

When you say InfluxQL do you mean transpiled InfluxQL or InfluxQL to v1 `/query` endpoint or both? If both how do you differentiate them?

fntlnz:
@nathanielc It's just InfluxQL to v1 `/query` now

cmd/idpd/main.go
fntlnz: Reasoned a bit more on this, since that dbrp mapping is not used and the receiver accepts an interface we can just pass nil to obtain the same behavior.

http/source_proxy_service.go
fntlnz: The only thing I found related to this is the parser and the scheme does not include the `:` . Anyway thanks for pointing this out, I did the same thing at line 86 but at some point this was confusing because one usually takes the token from the caller and parses it while in this case we take it from the source and construct it.

http/external_query_handler.go
fntlnz: switched as query parameter.

`text/plain` requests  going directly to `fluxd` looks like this now:

```
curl  -XPOST --data-binary "query=from(db: \"testmod\") |> range(start: -10d)" http://127.0.0.1:8093/query\?organization\=smerdo                                                              ,result,table,_start,_stop,_time,_value,_field,_measurement,server_name
,result,table,2018-08-14T11:36:21.91889644Z,2018-08-24T11:36:21.91889644Z,2018-08-21T10:58:17.939725773Z,5,body_bytes_sent,mymeasures,myserver                                                     ,result,table,2018-08-14T11:36:21.91889644Z,2018-08-24T11:36:21.91889644Z,2018-08-21T12:54:35.61523857Z,0,body_bytes_sent,mymeasures,myserver                                                      ,,,2018-08-14T11:36:21.91889644Z,2018-08-24T11:36:21.91889644Z,2018-08-21T10:58:17.939725773Z,239,bytes_sent,mymeasures,myserver                                                                   
,,,2018-08-14T11:36:21.91889644Z,2018-08-24T11:36:21.91889644Z,2018-08-21T12:54:35.61523857Z,178,bytes_sent,mymeasures,myserver                                                                    
```

This does not impact sources since sources uses `application/json` that is passing `query.ProxyRequest` as for purpose of this pr.

http/swagger.yml
```diff
@@ -1039,6 +1034,57 @@ paths:
...
+            - $ref: "#/components/schemas/FluxCompiler"
+            - $ref: "#/components/schemas/InfluxQLCompiler"
+            - $ref: "#/components/schemas/SpecCompiler"
+    QueryProxyRequest:
+      description: generic query request format for proxied queries
+      type: object
+      properties:
+        request:
+          $ref: "#/components/schemas/QueryRequest"
+        dialect:
```
nathanielc: Looks like `dialect_type` is missing.

http/swagger.yml
```diff
@@ -1039,6 +1034,57 @@ paths:
                 $ref: "#/components/schemas/Error"
 components:
   schemas:
+    SpecCompiler:
+      description: the spec compiler (documenting this will be a bit of work, for that reason we'll leave it sparse)
+      type: object
+      properties:
+        spec:
+          type: object
```
nathanielc: There is a ref already for the Spec, you should be able to just point to it from here.

fntlnz:
@goller Me and @desa made work the v1 endpoints with the new specification we defined together in swagger and removed the import cycle.


Example for influxql
```
curl -X POST \
  http://127.0.0.1:9999/v2/sources/0282fd7e1e268000/query \
  -H 'Content-Type: application/json' \
  -d '{
	"query": "select * from mymeasures",
	"type": "influxql",
	"db": "testmod",
	"dialect": {}
}'
```

Example for flux

```
curl -X POST \
  http://127.0.0.1:9999/v2/sources/0282fd7e1e268000/query \
  -H 'Content-Type: application/json' \
  -d '{
	"query": "from(db: \"testmod\") |> range(start: -10d)",
	"type": "flux",
	"dialect": {
		"delimiter": ","
	}
}'
```


The remaining part now is to do the same in the v2 source (the backend part is already done).

PR #652, opened 2018-08-17
add internal system buckets for orgs

jademcgough:
This PR adds per-org system buckets (which we'll be using to store task run information).


organization.go
```diff
@@ -4,8 +4,9 @@ import "context"
 
 // Organization is a organization. 🎉
 type Organization struct {
-	ID   ID     `json:"id"`
-	Name string `json:"name"`
+	ID       ID     `json:"id"`
+	Name     string `json:"name"`
+	BucketID ID     `json:"-"` // internal bucket for system use
```
mark-rushakoff: I'm not sure about this - it looks like this means there is only one system bucket in an organization. I understand the plan is to eventually support multiple system buckets within an organization. Although right now we're only targeting task logs, the anticipation is that we will eventually need different system buckets with different retention policies.

I guess we could add new fields if and when new system buckets are introduced. The original spec said that we'd have a reserved range of IDs for system buckets, but having the bucket IDs on the Organization type still lets us avoid doing a separate lookup.

cmd/influx/organization.go
```diff
@@ -70,6 +86,26 @@ func organizationCreateF(cmd *cobra.Command, args []string) {
 	w.Flush()
 }
 
+func createInternalBucket(o *platform.Organization) (*platform.Bucket, error) {
+	bucketS := &http.BucketService{
+		Addr:  flags.host,
+		Token: flags.token,
+	}
+
+	bucket := &platform.Bucket{
```
goller: What do you think of putting this definition as a global variable in bucket.go at the top-level?  That way other people would be able to see it and use it in the code

organization.go
goller: I don't think we need to state the bucketid here for it because we'll have a list of "known" buckets.

cmd/influx/organization.go
```diff
@@ -70,6 +86,26 @@ func organizationCreateF(cmd *cobra.Command, args []string) {
...
+		OrganizationID:  o.ID,
+		Name:            "internal",
```
goller: I'm thinking this should be named "task-logs"

bucket.go
```diff
@@ -13,6 +13,7 @@ type Bucket struct {
 	Name                string        `json:"name"`
 	RetentionPolicyName string        `json:"rp,omitempty"` // This to support v1 sources
 	RetentionPeriod     time.Duration `json:"retentionPeriod"`
+	Internal            bool          `json:"-"` // Internal buckets are filtered from http requests
```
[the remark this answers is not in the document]
jademcgough: 👍 yep, good idea

jademcgough:
Okay, this is ready for another review @goller.

I'm no longer storing a reference to the bucket within an org, and instead setting the bucket ID from a concatenation of the org ID and an enum value. This feels sort of clunky to me, but guarantees that there won't be any conflicts between these buckets and others with randomly generated IDs. Snowflake will always generate IDs of the same length, and these will be that length plus the enum.

PR #653, opened 2018-08-17
Encoding Error Tests

### LINKED ISSUE #651: Add test cases to make sure `EncoderError` is returned from `Encode` methods
As of #533,  `csv/result.go` to return a specialized `EncoderError` if any errors are encountered while encoding. This allows us to differentiate between errors that occur during the encoding of query results and errors from query execution. To verify that this change is valid, we should unit test `(*DelimitedMultiResultEncoder).Encode` and `(*csv.ResultEncoder).Encode)` to make sure the appropriate type of error is returned.

adamperlin:
Closes #651 

Adds test cases to ensure the correct error types are returned from csv encoding.
_What was the problem?_
#533 didn't have any extra test coverage.
_What was the solution?_
Add test coverage specific to encoding errors vs. other errors.
  - [x] Rebased/mergeable
  - [x] Tests pass


adamperlin:
@nathanielc changed the test code and refactored encoder error checking. How does this look?

query/result.go
```diff
@@ -137,18 +138,13 @@ type MultiResultEncoder interface {
 	Encode(w io.Writer, results ResultIterator) (int64, error)
 }
 
-type EncoderError struct {
-	msg string
+type encoderError interface {
```
nathanielc: One small change, it makes sense to have this interface public and give it a comment. That way implementations of encoders will know it exists.

PR #654, opened 2018-08-17
feat: hook up zap tracer with runtime tracing

zeebo:
Go 1.11 adds runtime tracing integration features that map well
to the concepts in opentracing. When tracing is not enabled, these
calls are cheap (some allocs and a conditional).

This will allow us to get tracing information out of the new
/debug/pprof/trace endpoint and analyze it with go tool trace.

|        | old        | new        | delta   | confidence      |
| ------ | ---------- | ---------- | ------- | --------------- |
| time   | 489ns ± 1% | 575ns ± 3% | +17.64% | (p=0.000 n=7+8) |
| alloc  | 368B ± 0%  | 424B ± 0%  | +15.22% | (p=0.000 n=8+8) |
| allocs | 6.00 ± 0%  | 8.00 ± 0%  | +33.33% | (p=0.000 n=8+8) |


zap/tracer_older.go
```diff
@@ -0,0 +1,13 @@
+// +build !go1.11
+
+package zap
+
+import "context"
+
+type task struct{}
+
+func newTask(pctx context.Context, name string) (context.Context, *task) {
+	return context.Background(), new(task)
```
mark-rushakoff: This looks like it should return `pctx` rather than a new context.

Issue #613, opened 2018-08-13
task: support re-run over time range


mark-rushakoff:
Task backend issue to support #611.

The current scheduler assumes time is always moving forward. We need to support re-runs of a requested time range. Supporting this case should simplify retries of a single run.

mark-rushakoff:
Here's roughly how I see this being implemented:

- New task/backend.Store method like `ManuallyRunTimeRange(taskID platform.ID, start, stop int64) error`, that creates a new [StoreTaskMetaManualRun](https://github.com/influxdata/platform/blob/be7a3e56dee504b5eaf61ce25e899aa4dc8578ba/task/backend/meta.pb.go#L167-L178) entry in the appropriate StoreTaskMeta. The field is write only for this commit or PR.
- Update [(*StoreTaskMeta).CreateNextRun](https://github.com/influxdata/platform/blob/be7a3e56dee504b5eaf61ce25e899aa4dc8578ba/task/backend/meta.go#L63-L64): a "naturally due" run takes precedence, then if there is a manual run queued, add it to stm.CurrentlyRunning, and set [RangeStart, RangeEnd, and RequestedAt fields](https://github.com/influxdata/platform/blob/be7a3e56dee504b5eaf61ce25e899aa4dc8578ba/task/backend/meta.pb.go#L112-L117) so that we can tell it was a run from a manual request, not a naturally scheduled run. The `StoreTaskMetaManualRun.LatestCompleted` field needs to move forward to indicate that we've progressed in the queue. If this was the last run of the queue, remove the `StoreTaskMetaManualRun` entry.

Then we're waiting on https://github.com/influxdata/platform/issues/611 to connect this to the rest of the API.

mark-rushakoff:
Closed in #698. 

#611 is still outstanding.

PR #656, opened 2018-08-17
StoreTaskMeta cleanup and support for task option delay

### LINKED ISSUE #571: task: respect delay option
Follows #572.

The delay option is used to compensate for data that may arrive late. For example, if you are relatively sure that data will not arrive later than ten seconds past its timestamp, you could set the task option `delay: 10s`.

When a delay is set, it must be positive and representable in whole seconds - `1500ms` is not valid, `1h` is okay. I would prefer to reject `delay: 0s` if possible; if you want no delay, omit the option.

When delay is present, the provided `cron` or `every` can no longer be interpreted literally. Given the option `every: 1h`, normally that would mean schedule at 1:00, 2:00, 3:00, etc. But if option `every: 1h` has `delay: 5m` alongside, then we need to _schedule_ at 1:05, 2:05, 3:05, etc., but we need to rewrite the `now` option to be 1:00, 2:00, 3:00, etc.

This will likely require a comprehensive set of unit tests to make sure our internal views of provided cron schedules are correct.

If it seems easier for us to fork the cron library we're using, that's an option we can take.

mark-rushakoff:
This is a fairly wide refactor that pushes schedule determination responsibility into StoreTaskMeta, and then replaces Store.CreateRun with Store.CreateNextRun. The work has been organized into a logical progression of commits.

The old behavior of Store.CreateRun was to just create a run for whatever timestamp was requested. The new behavior of Store.CreateNextRun is to track the runs that have been created and enforce that runs are created in the correct chronological order (via the StoreTaskMeta type).

As part of this change, a task's `delay` option is respected when scheduling a task.

This change also updates StoreTaskMeta to include new fields to handle manually requested runs (#613). Other than adding those fields, I have not yet begun work on that. Supporting manual re-runs should be a small change after this is merged.

Fixes #571.

### DIFF
--- a/task/backend/inmem_store.go
+++ b/task/backend/inmem_store.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/snowflake"
@@ -67,6 +68,7 @@ func (s *inmem) CreateTask(_ context.Context, org, user platform.ID, script stri
 		Status:         string(TaskEnabled),
 		LastCompleted:  scheduleAfter,
 		EffectiveCron:  o.EffectiveCronString(),
+		Delay:          int32(o.Delay / time.Second),
 	}
 
 	return id, nil
--- a/task/backend/inmem_store.go
+++ b/task/backend/inmem_store.go
@@ -233,38 +233,6 @@ func (s *inmem) Close() error {
 	return nil
 }
 
-// CreateRun adds `now` to the task's metaData if we have not exceeded 'max_concurrency'.
-func (s *inmem) CreateRun(ctx context.Context, taskID platform.ID, now int64) (QueuedRun, error) {
-	queuedRun := QueuedRun{}
-
-	stm, ok := s.runners[taskID.String()]
-	if !ok {
-		return queuedRun, errors.New("taskRunner not found")
-	}
-
-	if len(stm.CurrentlyRunning) >= int(stm.MaxConcurrency) {
-		return queuedRun, errors.New("MaxConcurrency reached")
-	}
-
-	runID := s.idgen.ID()
-
-	running := &StoreTaskMetaRun{
-		Now:   now,
-		Try:   1,
-		RunID: runID,
-	}
-
-	stm.CurrentlyRunning = append(stm.CurrentlyRunning, running)
-	s.mu.Lock()
-	s.runners[taskID.String()] = stm
-	s.mu.Unlock()
-
-	queuedRun.TaskID = taskID
-	queuedRun.RunID = runID
-	queuedRun.Now = now
-	return queuedRun, nil
-}
-
 func (s *inmem) CreateNextRun(ctx context.Context, taskID platform.ID, now int64) (RunCreation, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
--- a/task/backend/inmem_store.go
+++ b/task/backend/inmem_store.go
@@ -265,6 +265,28 @@ func (s *inmem) CreateRun(ctx context.Context, taskID platform.ID, now int64) (Q
 	return queuedRun, nil
 }
 
+func (s *inmem) CreateNextRun(ctx context.Context, taskID platform.ID, now int64) (RunCreation, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	stm, ok := s.runners[taskID.String()]
+	if !ok {
+		return RunCreation{}, errors.New("task not found")
+	}
+
+	makeID := func() (platform.ID, error) {
+		return s.idgen.ID(), nil
+	}
+	rc, err := stm.CreateNextRun(now, makeID)
+	if err != nil {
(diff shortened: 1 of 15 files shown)


task/backend/scheduler.go
```diff
@@ -4,25 +4,23 @@ import (
...
 
 var ErrRunCanceled = errors.New("run canceled")
 var ErrTaskNotClaimed = errors.New("task not claimed")
 
 // DesiredState persists the desired state of a run.
 type DesiredState interface {
-	// CreateRun returns a run ID for a task and a now timestamp.
-	// If a run already exists for taskID and now, CreateRun must return an error without queuing a new run.
-	CreateRun(ctx context.Context, taskID platform.ID, now int64) (QueuedRun, error)
+	// CreateNextRun requests the next run from the desired state, occurring no later than the Unix timestamp now.
```
docmerlin: why "no later than"?  That can't be right.

task/backend/scheduler.go
```diff
@@ -451,46 +361,48 @@ func (r *runner) IsIdle() bool {
...
 
-		r.tt.StartRun(next)
+	// Create a new child logger for the individual run.
+	// We can't do r.logger = r.logger.With(zap.String("run_id", qr.RunID.String()) because zap doesn't deduplicate fields,
+	// and we'll quickly end up with many run_ids associated with the log.
+	runLogger := r.logger.With(zap.String("run_id", qr.RunID.String()), zap.Int64("now", qr.Now))
 
-		go r.executeAndWait(qr, runLogger)
+	// TODO(mr): this used to record metrics or something?
+	// r.tt.StartRun(next)
```
docmerlin: delete commented code?

task/backend/scheduler_test.go
```diff
@@ -4,57 +4,31 @@ import (
...
-	"time"
 
 	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/kit/prom"
 	"github.com/influxdata/platform/kit/prom/promtest"
 	_ "github.com/influxdata/platform/query/builtin"
 	"github.com/influxdata/platform/task/backend"
 	"github.com/influxdata/platform/task/mock"
-	"github.com/influxdata/platform/task/options"
+	"go.uber.org/zap/zaptest"
```
docmerlin: 👍

task/backend/store.go
```diff
@@ -57,6 +57,28 @@ func (r RunStatus) String() string {
...
 
+// RunNotYetDueError is returned from CreateNextRun if a run is not yet due.
+type RunNotYetDueError struct {
+	// DueAt is the unix timestamp of when the next run is due.
+	DueAt int64
+}
+
+func (e RunNotYetDueError) Error() string {
+	return "run not due until " + time.Unix(e.DueAt, 0).Format(time.RFC3339)
+}
```
docmerlin: Good way to do this.

task/backend/inmem_store.go
```diff
@@ -226,36 +233,26 @@ func (s *inmem) Close() error {
 	return nil
 }
 
-// CreateRun adds `now` to the task's metaData if we have not exceeded 'max_concurrency'.
```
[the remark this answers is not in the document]
mark-rushakoff: Just didn't want to copy the comment from the interface. The deleted comment was stale.

task/backend/meta.go
```diff
@@ -25,3 +28,69 @@ func (stm *StoreTaskMeta) FinishRun(runID platform.ID) bool {
 	}
 	return false
 }
+
+// CreateNextRun attempts to update stm's CurrentlyRunning slice with a new run
+// whose now time is no later than the given now.
```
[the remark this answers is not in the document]
mark-rushakoff: Incoming, clearer version of this comment:

> // CreateNextRun attempts to update stm's CurrentlyRunning slice with a new run.
> // The new run's now is assigned the earliest possible time according to stm.EffectiveCron,
> // that is later than any in-progress run and stm's LastCompleted timestamp.
> // If the run's now would be later than the passed-in now, CreateNextRun returns a RunNotYetDueError.
> //
> // makeID is a function provided by the caller to create an ID, in case we can create a run.
> // Because a StoreTaskMeta doesn't know the ID of the task it belongs to, it never sets RunCreation.Created.TaskID.
> //
> // TODO: if a run is not yet due, and stm contains a manual run request, create a run for the manual request.

task/backend/scheduler.go
mark-rushakoff: Incoming updated comment:

> CreateNextRun requests the next run from the desired state, delegating to (*StoreTaskMeta).CreateNextRun.
> This allows the scheduler to be "dumb" and just tell DesiredState what time the scheduler thinks it is,
> and the DesiredState will create the appropriate run according to the task's cron schedule,
> and according to what's in progress and what's been finished.
>
> If a Run is requested and the cron schedule says the schedule isn't ready, a RunNotYetDueError is returned.

task/backend/scheduler.go
mark-rushakoff: Good catch - deleted, and other comments updated, in 66e97c938.

PR #664, opened 2018-08-20
Add BucketsAccessed method to OperationSpec and Authorizer interface

### LINKED ISSUE #565: query: internal API to identify affected buckets of a script
As part of validating authorizations on tasks, we need to validate that the user has appropriate read and write permissions on affected buckets. While it's probably possible to iterate through a query spec and extract the arguments from the `from` or `to` functions, it would be better if there were a consistent API so that task code wouldn't have to hardcode a function list.

wolffcm:
Closes #565

Add a method to OperationSpec to get the buckets accessed by an operation.

Add an Authorization interface that can verify user is authorized to access buckets read/written by query.

_What was the problem?_
_What was the solution?_

  - [x] Rebased/mergeable
  - [ ] Tests pass


query/authorizer.go
```diff
@@ -0,0 +1,63 @@
...
+			return errors.Wrapf(err, "Could not find bucket %v", writeBucketFilter)
+		}
+
+		reqPerm := platform.WriteBucketPermission(bucket.ID)
+		if ! platform.Allowed(reqPerm, auth.Permissions) {
+			return errors.New("No write permission for bucket: \"" + bucket.Name + "\"")
+		}
+	}
+
+	// TODO: log pass/fail decision to audit logger
```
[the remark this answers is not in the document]
wolffcm: That work has been deferred to a separate issue:
https://github.com/influxdata/platform/issues/666
I think that ultimately the parameter won't be of type Logger, so I'll remove it.

query/functions/function_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+	"github.com/influxdata/platform/query/values"
+	"github.com/influxdata/platform/query"
+)
+
+// TestBucketAwareOperationSpec verifies that any functions with parameters
+// named "bucket" or "bucketID" implement interface BucketAwareOperationSpec.
+func TestBucketAwareOperationSpec(t *testing.T) {
```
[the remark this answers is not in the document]
wolffcm: I asked Nathaniel about this.  He suggested just making the OperationSpec have a `bucketsAccessed` method, which means more boilerplate but avoids this fragility.  I'll push another commit that does this.

query/operation.go
```diff
@@ -69,6 +70,9 @@ type NewOperationSpec func() OperationSpec
 type OperationSpec interface {
 	// Kind returns the kind of the operation.
 	Kind() OperationKind
+
+	// BucketsAccessed returns which buckets are read and which are written by th operation
+	BucketsAccessed() (readBuckets, writeBuckets []platform.BucketFilter)
```
nathanielc: Now that we are using this for PreAutherize checks only I am OK if we move this method back to is own interface. That way we can reduce the boiler plate of new operations. And bugs of omission will not result in a vulnerability but only in a false positive on the pre auth check.

query/preauthorizer.go
```diff
@@ -0,0 +1,63 @@
...
+
+// PreAuthorizer provides a method for ensuring that the buckets accessed by a query spec
+// are allowed access by the given Authorization.  This is a pre-check provided as a way for
+// callers to fail early for operations that are not allowed.  However, it's still possible
+// for authorization to be denied at runtime even if this check passes.
+type PreAuthorizer interface {
+	PreAuthorize(ctx context.Context, spec *Spec, auth *platform.Authorization) error
+}
+
+// NewAuthorizer creates a new PreAuthorizer
```
nathanielc: Update the comment here with the new name. 

FYI, if you use the Go renaming tools, it will rename these comments for you. How to use the renaming depends on your IDE of choice but the renaming functions all end up calling the Go guru utility which will handle comments for you.

query/operation.go
wolffcm: Yes, I think that makes sense.  For new operations, there will be a lot of new code, perhaps some of it copy and pasted, so I think it only slightly reduces the chance of an error of omission putting `BucketsAccessed` in the general interface.  And less boilerplate is good.

query/preauthorizer.go
wolffcm: I didn't realize that the Go toolchain does renaming, that's neat and I will try it.

PR #670, opened 2018-08-20
chore(ci): parallelize js and go tests

### LINKED ISSUE #599: ci: use Circle 2.0 workflow to run JS and Go tests in parallel
Currently circle.yml just runs `make test` which builds some dependencies, then runs the JS tests, then runs the Go tests. We should be able to speed this up a bit with a [Circle Workflow](https://circleci.com/docs/2.0/workflows/) to run some things in parallel, but this means we would need to run more granular steps than `make test`.

/cc @nathanielc who probably knows the most about our Makefile.

lukevmorris:
Closes #599

This PR teases apart the golang and js build steps so that they can be run in parallel in circleci, or locally with `make test -j2`. Here's a screenshot of what the new ci workflow looks like:

screen shot 2018-08-20 at 2 25 42 pm

On master, these two will converge into a final `deploy` job which will build an image and push to quay.


lukevmorris:
@mark-rushakoff @nathanielc let me know if these Makefile changes look ok to you.

Makefile
```diff
@@ -123,12 +121,14 @@ generate:
 	# $(GO_GENERATE) ./chronograf/server/...
 	# $(GO_GENERATE) ./chronograf/canned/...
 
-jstest:
+test-js: .jsdep
 	cd chronograf/ui && yarn test --runInBand
 
-test: all jstest
+test-go: all
```
goller: instead of depending on all, it's probably best just to depend on .godep and generate

mark-rushakoff:
Looks fine to me, but I'll defer to @nathanielc since he's been much more involved with the Makefile than I have.

lukevmorris:
@goller and I paired for a bit at the end of the day to do some cleanup of the Makefile.

query/promql/promql.go
```diff
@@ -1,4 +1,3 @@
-// Package promql implements a promql parser to build flux query specifications from promql.
```
mark-rushakoff: I doubt you intended to delete this line.

Can you be sure to rebase before merging?

lukevmorris: Rebased just now, and I'll remember to click the Rebase + Merge button.

This was removed as a result of running `make`, so I assumed someone had added this comment to a generated file. I can add back.

lukevmorris:
@mark-rushakoff @nathanielc Let me know if there's anything else I need to do to get this merged. Thanks!

.circleci/config.yml
```diff
@@ -1,14 +1,20 @@
...
-      # specify the version
       - image: circleci/golang:1.10.3-node-browsers
+    working_directory: /go/src/github.com/influxdata/platform
+    steps:
+      - checkout
+      - run: make test-js
 
+  gotest:
+    docker:
+      - image: circleci/golang:1.10.3-node-browsers
```
mark-rushakoff: Can this be Go without node-browsers, and can it be Go 1.11?

.circleci/config.yml
```diff
@@ -1,14 +1,20 @@
 version: 2
 jobs:
-  build:
+  jstest:
     docker:
-      # specify the version
       - image: circleci/golang:1.10.3-node-browsers
```
mark-rushakoff: Please switch this to Go 1.11 too.

PR #673, opened 2018-08-21
update links to open issues in spec

### LINKED ISSUE #545: Remove closed issues from the query spec and update links to point to the correct url
The query spec has a lot of links that only make sense when you point them to the archived ifql repository. We migrated and closed some of the issues, but the links were only modified to `s@ifql@platform/query` and so all of the links are broken.

Go through fixing the links and removing the entries that are closed.

jlapacik:
Closes #545 

Updating all links and issue numbers in the current query spec

_What was the problem?_

Links and issue numbers of yet-to-be-implemented functionality in the current spec point to the old ifql repo

_What was the solution?_

Update the links to point to the current repo

  - [ ] Rebased/mergeable
  - [ ] Tests pass


query/docs/SPEC.md
```diff
@@ -1039,7 +1035,7 @@ Missing values are represented with a special _null_ value.
 The _null_ value can be of any data type.
 
 
-[IMPL#219](https://github.com/influxdata/platform/query/issues/219) Design how nulls behave
+[IMPL#353](https://github.com/influxdata/platform/issues/353) Design how nulls behave
```
adamperlin: I think the root issue here is actually #300

PR #675, opened 2018-08-21
Bounds Refactor

### LINKED ISSUE #639: Update`plan.BoundsSpec` to use absolute time bounds
The `plan.BoundsSpec` defined in `query/plan/procedure.go` represents the time bounds the planner handles internally. Currently, it can hold both relative and absolute time bounds because flux functions like `range` use this type to store their own bounds. However, the planner's time bounds should not be relative, it only makes bounds logic more complicated. Instead, we should update the `plan.BoundsSpec` struct to contain `values.Time` (always absolute), and add an additional `query.Bounds` struct which uses the `query.Time` type (which can be either relative or absolute) which flux functions can use instead.

adamperlin:
Closes #639 

Moves the task of relative time resolution to the planner. The planner's `BoundsSpec` struct now uses `values.Time` as opposed to `query.Time`. A new `query.Bounds` struct was added which uses `query.Time`. It is now used by flux functions instead of the `BoundsSpec`.

_What was the problem?_
Deferring time resolution past the planner meant that the current `now` context had to be passed around whenever bounds operations were performed while planning, which made the code unnecessarily messy.
_What was the solution?_
Resolve bounds directly in the planner, once for each set of bounds.

  - [x] Rebased/mergeable
  - [x] Tests pass


query/execute/executor.go
```diff
@@ -128,13 +128,19 @@ func (es *executionState) createNode(ctx context.Context, pr *plan.Procedure, no
 		return n, nil
 	}
 
+	// Add explicit stream context if bounds are set on this node
+	var streamContext streamContext
+	if pr.Bounds != nil {
+		streamContext = newStreamContext(Bounds{
```
nathanielc: This code makes it read like the `streamContext` is a pointer. Instead why not just set `streamContext.Bound` to the correct bounds if the procedure bounds exist?
```
if pr.Bounds != nil {
    streamContext.Bounds = ...
}
```

query/plan/physical.go
```diff
@@ -60,6 +67,24 @@ func NewPlanner() Planner {
...
+func resolveTime(qt query.Time, now time.Time) values.Time {
+	return values.ConvertTime(qt.Time(now))
+}
+
+func ToBoundsSpec(bounds query.Bounds, now time.Time) (*BoundsSpec, error) {
+	if bounds.HasZero() {
+		return nil, errors.New("bounds contain zero time")
+	}
+	if bounds.IsEmpty(now) {
+		return nil, errors.New("bounds are empty")
```
nathanielc: Why are empty bounds an error?

query/plan/physical.go
```diff
@@ -136,26 +161,31 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
 		// The bounds of the current procedure are always the union
 		// of the bounds of any parent procedure
 		pr.DoParents(func(parent *Procedure) {
-			if pr.Bounds.HasZero() {
+			if pr.Bounds != nil && parent.Bounds != nil {
```
nathanielc: I think this is simpler if you return early if `parent.Bounds == nil`. That way you only have to check it once.

query/plan/physical.go
```diff
@@ -136,26 +161,31 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
...
 		// from its parents, and its own bounds.
 		if bounded, ok := pr.Spec.(BoundedProcedureSpec); ok {
-			if pr.Bounds.HasZero() {
-				pr.Bounds = bounded.TimeBounds()
-			} else {
-				newBounds := pr.Bounds.Intersect(bounded.TimeBounds(), now)
-				if newBounds != EmptyBoundsSpec {
+			convertedBounds, err := ToBoundsSpec(bounded.TimeBounds(), pr.plan.Now)
+			if err != nil {
+				return nil, errors.Wrap(err, "invalid time bounds provided")
```
nathanielc: Can you add the context of which procedure caused the error?

query/plan/physical.go
```diff
@@ -136,26 +161,31 @@ func (p *planner) Plan(lp *LogicalPlanSpec, s Storage) (*PlanSpec, error) {
...
+			}
+
+			if pr.Bounds != nil {
+				newBounds := pr.Bounds.Intersect(convertedBounds)
+				if !newBounds.IsEmpty() {
```
nathanielc: Why do we ignore the empty bounds?

query/plan/procedure.go
```diff
@@ -105,88 +111,81 @@ type ProcedureKind string
...
-	Stop:  query.Now,
+	Start: values.Time(0),
+	Stop:  values.Time(0),
 }
 
-// IsEmpty reports whether the given bounds
-// are empty, i.e., if start >= stop.
-func (b BoundsSpec) IsEmpty(now time.Time) bool {
-	return b.Start.Time(now).Equal(b.Stop.Time(now)) || b.Start.Time(now).After(b.Stop.Time(now))
+func NewEmptyBoundsSpec() *BoundsSpec {
```
nathanielc: Why is this function useful? Maybe just make the `EmptyBoundsSpec` var a pointer type itself?

query/plan/physical.go
adamperlin: just changed per our conversation

adamperlin:
@nathanielc Addressed your review comments. I also added a commit removing flux functions' dependence on the `(*Administration.ResolveTime)` function because it seemed confusing to do bounds resolution both there and in the planner with the same end result.

query/execute/executor.go
```diff
@@ -128,13 +128,24 @@ func (es *executionState) createNode(ctx context.Context, pr *plan.Procedure, no
...
+	var streamContext streamContext
+	if pr.Bounds != nil {
+		if !pr.Bounds.IsEmpty() {
+			streamContext.bounds = &Bounds{
+				Start: pr.Bounds.Start,
+				Stop:  pr.Bounds.Stop,
+			}
+		} else {
+			// If bounds are empty, standardize them to the default EmptyBounds
+			streamContext.bounds = EmptyBounds
```
nathanielc: Why is this needed? All consumers should be using IsEmpty to check if bounds are empty instead of comparing against the EmptyBounds object.

query/functions/window.go
```diff
@@ -185,10 +185,17 @@ func createWindowTransformation(id execute.DatasetID, mode execute.AccumulationM
 	} else {
 		start = a.ResolveTime(s.Window.Start)
 	}
+
+	bounds := a.StreamContext().Bounds()
+	// TODO(adamperlin): could also use empty bounds by default in this case
+	if bounds == nil {
+		return nil, nil, errors.New("nil bounds passed to window")
+	}
```
nathanielc: EmptyBounds are different than nil bounds. The current behavior of returning an error is correct. We will need to validate that all `window` procedures have non nil bounds, but that change can be part of the empty windows change since that is when it matters. 

Can you remove the TODO? This code is correct as is.

adamperlin:
@nathanielc Should be ready for a final look. I changed the IgnoreBounds parameter of the window function as part of this PR, but I can move those changes over to another branch if you'd like. I'll resolve the merge conflicts if those changes stay part of this PR.

query/bounds.go
```diff
@@ -0,0 +1,24 @@
+package query
+
+import "time"
+
+type Bounds struct {
+	Start Time
+	Stop  Time
+}
+
+var EmptyBounds = Bounds{
```
nathanielc: As far as I can tell this is only used in tests. If so lets define it in the tests package.

query/execute/bounds.go
```diff
@@ -26,6 +26,11 @@ var AllTime = Bounds{
 	Stop:  MaxTime,
 }
 
+var EmptyBounds = &Bounds{
```
nathanielc: Same here.

PR #685, opened 2018-08-23
feat: Add histogramQuantile function

nathanielc:


query/functions/histogram_quantile.go
```diff
@@ -0,0 +1,295 @@
...
+	} else {
+		s.CountColumn = execute.DefaultValueColLabel
+	}
+
+	if col, ok, err := args.GetString("upperBoundColumn"); err != nil {
+		return nil, err
+	} else if ok {
+		s.UpperBoundColumn = col
+	} else {
+		s.UpperBoundColumn = "upperBound"
```
aanthony1243: I'm leaning toward saying this parameter should be required.  upperBound is a column name that may not occur naturally in data sets, so the user will be providing a value here possibly every time they call the function.

query/functions/histogram_quantile.go
```diff
@@ -0,0 +1,295 @@
...
+) execute.Transformation {
+	return &histogramQuantileTransformation{
+		d:     d,
+		cache: cache,
+		spec:  *spec,
+	}
+}
+
+func (t histogramQuantileTransformation) RetractTable(id execute.DatasetID, key query.GroupKey) error {
+	// TODO
```
aanthony1243: need todo this?

query/functions/testdata/percentile.flux
```diff
@@ -1,6 +1,6 @@
 from(db: "test")
     |> range(start: 2018-05-22T19:50:26Z)
     |> group(by: ["_measurement", "_start"])
-    |> percentile(p:0.75, method:"exact_selector")
+    |> percentile(percentile:0.75, method:"exact_selector")
     |> map(fn: (r) => {_time: r._time, percentile: r._value})
-    |> yield(name:"0")
\ No newline at end of file
+    |> yield(name:"0")
```
aanthony1243: can we add a querytest for quantile?  if you cut off your real-data query from the monitoring server to right before the quantile call, you can use that output as input directly to the quantile function, should be quick + painless.   then run  `prepcsvtests . quantile`  to generate the output.

query/functions/histogram_quantile.go
nathanielc: I was too, so I'll change it.

PR #687, opened 2018-08-23
Parse duration literals

### LINKED ISSUE #253: Parse duration literals
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#311  

From the spec:

A duration literal is a representation of a length of time.
It has an integer part and a duration unit part.
Multiple duration may be specified together and the resulting duration is the sum of each smaller part.

    duration_lit        = { int_lit duration_unit } .
    duration_unit       = "ns" | "u" | "µ" | "ms" | "s" | "m" | "h" | "d" | "w" .

| Units    | Meaning                                 |
| -----    | -------                                 |
| ns       | nanoseconds (1 billionth of a second)   |
| us or µs | microseconds (1 millionth of a second)  |
| ms       | milliseconds (1 thousandth of a second) |
| s        | second                                  |
| m        | minute (60 seconds)                     |
| h        | hour (60 minutes)                       |
| d        | day (24 hours)                          |
| w        | week (7 days)                           |

Durations represent a fixed length of time.
They do not change based on time zones or other time related events like daylight savings or leap seconds.

Examples:

    1s
    10d
    1h15m // 1 hour and 15 minutes
    5w

wolffcm:
Closes #253

_What was the problem?_
_What was the solution?_

  - [x] Rebased/mergeable
  - [ ] Tests pass


query/parser/types.go
```diff
@@ -341,30 +341,41 @@ func regexLiteral(chars interface{}, text []byte, pos position) (*ast.RegexpLite
...
+	}
+
+	if otherParts == nil {
+		slice := make([]*singleDurationLiteral, 1, 10)
+		slice[0] = sdl
+		return slice, nil
+	} else {
+		others := otherParts.([]*singleDurationLiteral)
+		return append(others, sdl), nil
+	}
```
jlapacik: Can remove else here

query/parser/flux.peg
```diff
@@ -344,11 +344,136 @@ DateTimeLiteral
       return datetime(c.text, c.pos)
     }
 
+// The order of the choices in this rule and those below matters.
+// Minutes (m) must appear before milliseconds (ms) and months (mo),
+// otherwise "10ms" will be parsed as 10 minutes and an extraneous "s".
```
jsternberg: Random comment on this. Most lexer generators recognize this and check for the longest matching. Does our current lexer/parser generator not do that?

Additionally, is the comment here correct? I looked at the issue this closes and the spec and this isn't part of it. The spec doesn't mention that durations parts cannot be repeated and it doesn't mention a specific order. I like this method better. I think we should be more strict about the duration literals and require them in a specific order, but that's not in the spec so we would need to propose a change to the spec first.

query/parser/flux.peg
```diff
@@ -344,11 +344,136 @@ DateTimeLiteral
...
+      / MinuteDuration
+      / NanoSecondDuration
+      / SecondDuration
+      / WeekDuration
+      / YearDuration
+    ) {
+      return durationLiteral(durations, c.text, c.pos)
+    }
+
+YearDuration
```
jsternberg: Because of the above comment, I don't think this much of a change is needed. I think you just need to add the extra literals we have missed and then update the duration units section.

query/parser/flux.peg
wolffcm: I can confirm that it does not choose the longest matching.  It just chooses the first match.  I was surprised by this too, but the docs for pigeon state it clearly.  Some of the tests that use ms durations started failing after my change, which was how I discovered I need to reorder them.

query/docs/SPEC.md
```diff
@@ -163,6 +163,7 @@ Examples:
 A duration literal is a representation of a length of time.
 It has an integer part and a duration unit part.
 Multiple durations may be specified together and the resulting duration is the sum of each smaller part.
+When several durations are specified together, larger units must appear before smaller ones, and there can be no repeated units.
```
jsternberg: Can you make some mention that the table below is in the order of smallest to largest? That way it's explicit about the ordering.

wolffcm: How about if I just reverse the order of the rows in the table?

PR #689, opened 2018-08-24
Materialize Empty Windows

### LINKED ISSUE #669: Window function should create empty windows for the procedure bounds

adamperlin:
Closes #669 

Changes the logic of the window function to generate all windows which intersect the window procedure's bounds, even windows which no records fall in.

_What was the problem?_
The window function wouldn't create empty tables for windows without data, which meant that functions like `count` would be unable to aggregate return data for all windows.

_What was the solution?_
Create a table for every window that intersects the window procedure's bounds, even if that table is empty.

  - [x] Rebased/mergeable
  - [x] Tests pass


query/functions/window.go
```diff
@@ -286,12 +286,20 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
...
+	allBounds := t.generateWindowsWithinBounds()
+
 	return tbl.Do(func(cr query.ColReader) error {
 		l := cr.Len()
 		for i := 0; i < l; i++ {
 			tm := cr.Times(timeIdx)[i]
-			bounds := t.getWindowBounds(tm)
-			for _, bnds := range bounds {
+
+			// TODO: avoid creating empty tables more than once to get better performance
```
aanthony1243: I think we need to work this out before we merge back.  The performance of this will be very poor if the number of bounds is ~ the number of points.

query/functions/window.go
```diff
@@ -286,12 +286,20 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
 		keyColMap = append(keyColMap, len(keyColMap))
 	}
 
+	// Abort processing if no data will match bounds
+	if t.bounds.IsEmpty() {
+		return nil
+	}
+
+	allBounds := t.generateWindowsWithinBounds()
```
nathanielc: This should be done only once, when the transformation is created.

query/functions/window.go
```diff
@@ -286,12 +286,20 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
...
+
 	return tbl.Do(func(cr query.ColReader) error {
 		l := cr.Len()
 		for i := 0; i < l; i++ {
 			tm := cr.Times(timeIdx)[i]
-			bounds := t.getWindowBounds(tm)
-			for _, bnds := range bounds {
+
+			// TODO: avoid creating empty tables more than once to get better performance
+			for _, bnds := range allBounds {
```
nathanielc: Instead of iterating over all bounds we can search the existing list of allBounds since it can be sorted.

query/functions/window.go
```diff
@@ -286,12 +286,20 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
 		keyColMap = append(keyColMap, len(keyColMap))
 	}
 
+	// Abort processing if no data will match bounds
+	if t.bounds.IsEmpty() {
+		return nil
+	}
+
+	allBounds := t.generateWindowsWithinBounds()
+
```
nathanielc: I expected a loop outside the tbl.Do function that would iterator over allBounds and create the tables.

Then within tbl.Do you would just assign each point to the already existing tables.

query/functions/window.go
```diff
@@ -344,44 +360,31 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
...
-		}
-
-		if bnds.Start < t.bounds.Start {
-			bnds.Start = t.bounds.Start
-		}
-
-		// Check bounds again since we just clamped them.
-		if bnds.Contains(now) {
-			bounds = append(bounds, bnds)
-		}
```
nathanielc: Hmm, I know I said not to clamp them but I am now reconsidering that. I think we do want to clamp them. It would be misleading to say that a window represents interval X when in reality it only contains data for Y which is smaller than X. Maybe we need an option for this? i.e. `clamp: true/false` Then we can have a discussion about what the default is.

query/functions/window.go
```diff
@@ -342,21 +348,59 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
...
+	vs := make([]values.Value, len(keyCols))
+	for j, c := range keyCols {
+		cols[j] = c
+		switch c.Label {
+		case t.startColLabel:
+			vs[j] = values.NewTimeValue(bnds.Start)
+		case t.stopColLabel:
+			vs[j] = values.NewTimeValue(bnds.Stop)
+		default:
+			vs[j] = tbl.Key().Value(keyColMap[j])
```
aanthony1243: we can code around this loop. talked in person.

aanthony1243:
just need to regenerate the tests when the emptywindows parameter is done, and I think this is good to go.

adamperlin:
@aanthony1243 done. I also added another test for empty windows specifically.

query/functions/window.go
```diff
@@ -286,33 +308,36 @@ func (t *fixedWindowTransformation) Process(id execute.DatasetID, tbl query.Tabl
...
+		key := t.newWindowGroupKey(tbl, keyCols, bnds, keyColMap)
+		builder, created := t.cache.TableBuilder(key)
+		if created {
+			for _, c := range newCols {
+				builder.AddCol(c)
+			}
+		}
+	}
+
+	err := tbl.Do(func(cr query.ColReader) error {
```
nathanielc: A small simplification is to just return directly from here i.e. `return tbl.Do(...`

query/promql/promql.go
```diff
@@ -1,4 +1,3 @@
-// Package promql implements a promql parser to build flux query specifications from promql.
```
nathanielc: This should not be changed.

PR #692, opened 2018-08-24
docs(http): add write endpoint documentation

goller:
_What was the problem?_

Design for the write endpoint

_What was the solution?_

Swagger it up

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -5,6 +5,76 @@ info:
 servers:
   - url: /v1
 paths:
+  /write:
+    post:
+      tags:
+        - Write
+      summary: write time-series data into influxdb
+      parameters:
+        - in: header
+          name: Accept-Encoding
+          description: when present, its value indicates compression is applied to the entity-body
+          schema:
+            type: string
+            description: specifies that the line protocol in the body is compressed with gzip
+            enum:
+              - gzip
+        - in: query
+          name: precision
+          description: specifies the precision for the unix timestamps within the body line-protocol
+          schema:
+            type: string
+            default: ns
+            description: specifies the unit of time
+            enum:
+              - ns
+              - us
+              - u
+              - ms
+              - s
+              - m
+              - h
+      responses:
+        '204':
+          description: write data is correctly formatted and accepted for writing to the database.
+        '400':
+          description: line protocol poorly formed and no points were written.  Response can be used to determine the first malformed line in the body line-protocol.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/LineProtocolError"
+        '401':
+          description: token does not have sufficient permissions to write to this organization and bucket.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        '403':
+          description: no token was sent and they are required.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        '413':
+          description: write has been rejected because the payload is too large. Error message returns max size supported.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/LineProtocolLengthError"
+        '503':
+          description: server is temporarily unavailable to accept writes.  The Retry-After header describes when to try the write again.
+          headers:
+            Retry-After:
+              description: A non-negative decimal integer indicating the seconds to delay after the response is received.
+              schema:
+                type: integer
+                format: int32
+        default:
+          description: internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
   /ping:
     servers:
       - url: /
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
...
+              - ns
+              - us
+              - u
+              - ms
+              - s
+              - m
+              - h
+      responses:
+        '204':
+          description: write data is correctly formatted and accepted for writing to the database.
```
desa: *nit* `%s/database/bucket/`

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Routes"
+  /write:
+    post:
+      tags:
+        - Write
+      summary: write time-series data into influxdb
+      parameters:
+        - in: header
```
danielnelson: Are there any other headers a client ought to use?  I'm thinking about `Content-Type`, `Content-Length` and `Accept`.

Current Telegraf implementation for comparison: we use `Content-Type: text/plain; charset=utf-8` and do not send Content-Length or Accept.  Content-Length is not sent now because we stream the metrics and don't know the length, but this is probably not that helpful in the recommended batch sizes.

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
...
+          name: Accept-Encoding
```
danielnelson: Accept-Encoding controls the response, but what about Content-Encoding for the request?

http/swagger.yml
danielnelson: Previous /write endpoint would send several "warnings" as 4xx responses: "points beyond retention policy" and "hinted handoff queue not empty".  These always should have been 2xx responses IMO but we should consider if we still need anything like this.  For Telegraf, we just ignore these warnings anyway so 204 is good.

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
...
+          description: when present, its value indicates compression is applied to the entity-body
+          schema:
+            type: string
+            description: specifies that the line protocol in the body is compressed with gzip
+            enum:
+              - gzip
+        - in: query
```
danielnelson: Do we need a query param for buckets?

http/swagger.yml
goller: bucket is required

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        '403':
+          description: no token was sent and they are required.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        '413':
```
goller: Add 429 with a retry-after that is when one is overquota.

http/swagger.yml
goller: Fixed in d4c95cc

http/swagger.yml
goller: Fixed in d4c95cc

http/swagger.yml
goller: Fixed in d4c95cc

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Routes"
+  /write:
+    post:
```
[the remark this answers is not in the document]
goller: @danielnelson Would `PUT` have different semantics than a `POST` to write?

http/swagger.yml
goller: Fixed in 0da3fa9

http/swagger.yml
```diff
@@ -15,6 +15,76 @@ paths:
...
+      summary: write time-series data into influxdb
+      parameters:
+        - in: header
+          name: Accept-Encoding
+          description: when present, its value indicates compression is applied to the entity-body
+          schema:
+            type: string
+            description: specifies that the line protocol in the body is compressed with gzip
+            enum:
+              - gzip
```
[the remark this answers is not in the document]
goller: Fixed in 52e71f9

http/swagger.yml
goller: @danielnelson I mixed this up. I don't think we need an accept-encoding just a content-encoding.

goller: Fixed in c85517f

http/swagger.yml
goller: @danielnelson I'll put all 3 in.  I think content-type is good as I imagine a future when we allow people sending arrow buffers.

http/swagger.yml
goller: HH not empty shouldn't be an error to me.  We have other ways of hinting the client in this specification with the 503 Retry-After.

Points beyond retention policy is a bit trickier.  I can understand if the database just sends those points to /dev/null but in some ways if I accidentally write a timestamp of 0, I want to know why my points are showing up. 

@pauldix do you think the database should send an error to the user if points are outside the retention policy?

pauldix: @goller yes definitely, if the timestamp of the data is older than the RP, the write should fail and send a 400.

goller: @pauldix @danielnelson I'm thinking 422 with a response stating the line containing the timestamp error and also the retention duration.

goller: Added in 42f6b27

jademcgough:
https://github.com/influxdata/platform/pull/682 has the write path as `/v1/organizations/:org_spec/buckets/:bucket_spec/write`, with org and bucket in the path, not in the query. Which one are we going to use?

goller:
@jademcgough I'm thinking
/v2/write?org=goller&bucket=telegraf

I'll add the org parameter

goller:
@jademcgough I added the org parameter in 6b6aad3

http/swagger.yml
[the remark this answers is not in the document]
goller: In talking to @danielnelson about this, we'll not put `PUT` in.

http/swagger.yml
goller: Accepting data doesn't tell you how the data is going to be stored because it can be further processed in the write pipeline (e.g. field conflicts, downsampling, etc)

Issue #611, opened 2018-08-13
api: define how to re-run task over time range


mark-rushakoff:
We have an endpoint to retry a single run: https://github.com/influxdata/platform/blob/6a861f2648a3a52da9f022bb892a29ba98d72a74/http/swagger.yml#L855-L885

But we don't yet have an endpoint that says "re-run this task over the time range [start, end]". We need that endpoint to support manual re-runs of a task for a known backfilled time range.

docmerlin:
I was thinking something like:
```yaml
  '/tasks/{taskId}/runs/{begin}/{end}/retry':
    post:
      tags:
        - Tasks
      summary: Retry a task run on interval [begin, end)
      parameters:
        - in: path
          name: taskId
          schema:
            type: string
          required: true
          description: task ID
        - in: path
          name: begin
          schema:
            type: string
            format: date-time
          required: true
          description: beginning of retry interval
        - in: path
          name: end
          schema:
            type: string
            format: date-time
          required: true
          description: ending of retry interval
```

goller:
Talking to @jademcgough and @mark-rushakoff about this:

Think this should be a  new action,`POST`, on the /runs endpoint with new parameters about a specific new run rather than retry.  A retry should be a simple way to just retry a specific run rather than modify it.

PR #698, opened 2018-08-24
Support manual runs in task storage and scheduler

mark-rushakoff:
This is separated into 3 pretty logical commits, adding support for manual runs to task/backend.Store, task/backend.StoreTaskMeta, and task/backend.Scheduler.

Updates #613. Then waiting on #611, then will need to finish gluing everything together.


task/backend/scheduler.go
```diff
@@ -226,6 +226,9 @@ func (s *outerScheduler) PrometheusCollectors() []prometheus.Collector {
 
 // taskScheduler is a lightweight wrapper around a collection of runners.
 type taskScheduler struct {
+	// Reference to outerScheduler.now. Must be accessed atomically.
+	now *int64
```
lyondhill: I still feel like its a bit strange to pass around a 'now' object. I wonder if it would be valuable to pass a interface that handles the atomic access internally. Probably a change for a separate PR.

task/backend/scheduler.go
```diff
@@ -242,6 +245,7 @@ type taskScheduler struct {
 	nextDueMu     sync.RWMutex // Protects following fields.
 	nextDue       int64        // Unix timestamp of next due.
 	nextDueSource int64        // Run time that produced nextDue.
+	hasQueue      bool         // Whether there is a queue of manual runs.
```
lyondhill: The check for this is light weight. We may just want this to be `func hasQueue() bool { return len(meta.ManualRuns) > 0 }`. Im not sure it matters but in some later code update the hasQueue could get stale between setting and getting.

PR #699, opened 2018-08-24
docs(http): add disabled boolean for authorization tokens

goller:

Token permissions are immutable, but, it would be nice to enable/disable tokens rather than just DELETE them.

_What is the problem?_
Administrators should have the ability to temporarily disable a token (perhaps the writes using a token are hurting the system).

Once the issue has been resolved the token can be enabled.  This way the token doesn't need to be DELETED, but, just disabled.

_What was the solution?_

My first boolean!! I feel like it is legit in this case that a token only has 2 states.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -781,6 +781,37 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+    patch:
+      tags:
+        - Authorizations
+      summary: enable/disable authorization
+      requestBody:
+        description: authorization to update to apply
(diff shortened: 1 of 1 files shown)


goller:
@desa I've updated the PR test with
```
What is the problem?
Administrators should have the ability to temporarily disable a token (perhaps the writes using a token are hurting the system).

Once the issue has been resolved the token can be enabled. This way the token doesn't need to be DELETED, but, just disabled.```

PR #701, opened 2018-08-24
fix: Fix missing builtins in REPL

nathanielc:
The REPL's use of the interpreter was causing it to not get the builtins
defined in builtin scripts. For example the `top` function was missing.

This change fixes the issues by ensuring the builtins are only evaluated
once and that there is only one way to get the query Interpreter that is
guaranteed to have the proper builtins.


query/interpreter/interpreter.go
```diff
@@ -17,13 +17,27 @@ type Interpreter struct {
...
-	interpreter.options = optionScope
-	interpreter.globals = globalScope
+	interpreter := &Interpreter{
+		options: optionScope,
+		globals: globalScope.Nest(),
+	}
+	return interpreter
+}
+
+// NewInterpreter instantiates a new Flux Interpreter whose builtin values are mutable.
```
jlapacik: NewInterpreter -> NewMutableInterpreter

PR #712, opened 2018-08-27
 feat(auth): allow authorizations to be enabled/disabled 

### LINKED ISSUE #691: Add disabled: bool for authorization token

goller:
Closes #691 

Authorizations should be immutable, except we should allow them to be disabled without them being deleted.  See #699

_What was the problem?_
There are many cases when a token is to be temporarily disabled, like an administrative disable.

_What was the solution?_
Adding my first boolean

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] swagger.json updated (if modified Go structs or API) See #699


auth.go
```diff
@@ -9,6 +9,7 @@ import (
 type Authorization struct {
 	ID          ID           `json:"id"`
 	Token       string       `json:"token"`
+	Disabled    bool         `json:"disabled"`
```
nathanielc: This makes sense for the structure since its easy to store etc. My only concern is it will be easy for a code path that is validating authorizations  to forget to check the disabled flag. Should we add FindEnabled* variants to the Service that only return authorizations that are enabled? The content of a disabled auth is meaningless to the code since it in effect does not exist. Thoughts?

goller: @nathanielc I think all code paths that check properties of Authorization should be in this auth.go file.  That way there is a single file that we can inspect for permissions checking.

For example, in this file is an `Allowed` function.  It checks the `Disabled` bool.

nathanielc: That works for me.

goller:
@desa I'll make that change.  I think you are right that active/inactive is closer to what is really happening.

goller:
@desa I added active/inactive in 8b4cf48 and documented it in 552f20d

PR #724, opened 2018-08-28
ci: use Go cache and dep cache

### LINKED ISSUE #705: ci: run go test -race
There's already a step in the Makefile, `test-race` I believe, that runs `go test -race`; but we aren't currently running it in CI, and we have at least two tests that hit data races: #655 and #697.

Both of those issues have [`label: data race`](https://github.com/influxdata/platform/issues?q=is%3Aissue+is%3Aopen+label%3A%22data+race%22) applied. We probably need to fix those before we can enable the race detector.

mark-rushakoff:
- Update to Go 1.11 docker images
- Use the Go cache between Circle builds, and add `go test -race`
- Use the dep cache between Circle builds

So, we've added `go test -race` and some caching, and builds take as long as before, which did not have caching and did not run with the race detector.

Closes #705.


.circleci/config.yml
```diff
@@ -2,28 +2,59 @@ version: 2
...
       - checkout
       - run: make test-js
 
   gotest:
     docker:
-      - image: circleci/golang:1.10.3
+      - image: circleci/golang:1.11
+    environment:
+      GOCACHE: /tmp/go-cache
+      DEPCACHEDIR: /tmp/dep-cache
```
lukevmorris: These values are referred to explicitly below, do we still need them as environment variables?

mark-rushakoff: Yes, they're both special environment variables that `go` and `dep` respectively consider for overriding the default cache location.

PR #729, opened 2018-08-28
ci: restore cache for deploy job

mark-rushakoff:
The last few runs of deploy took 3m33s, 3m40s, and 4m27s. It should be
noticeably faster after we restore the cache.


mark-rushakoff:
In https://circleci.com/gh/influxdata/platform/1900, `make nightly` took 2m34s, plus 26s to restore the cache, so it looks like a net win.

PR #734, opened 2018-08-29
feat: use docker for pigeon and file generation

jsternberg:
This is the first attempt at using docker for generating files so that
we only need docker for generating files and we can build and pin the
tools used to generate files without using dep.

This is useful because it means we don't have to add the dependency to
dep and it works for non-go dependencies. This will make it trivial to
transition to using go modules so we can drop dep as a dependency.


tools/pigeon/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.10.3 as builder
```
mark-rushakoff: The rest of platform should be on Go 1.11 now, so this should probably match.

tools/pigeon/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ENV PIGEON_REVISION 2ad051b8b508f69e7dcf8febdea2856ec2db73ed
```
mark-rushakoff: I'd prefer to see this as a build arg, so we can record the revision in a single place. Probably would need to be named `TOOL_REVISION` so it can be applied generically from `build.sh`.

nathanielc:
As for non Go dependencies, like `protoc`. We can provide a single Dockerfile that may be used to build the code if a contributor does not want to spend the effort to the the right dependency installed on their host machine. 

Providing a Dockerfile is significantly different than requiring Docker in order to build the code.

tools/pigeon/Dockerfile
jsternberg: Has the docker image for 1.11 been pushed yet?

jsternberg:
In this scenario, docker wouldn't be needed for building the code. We commit all dependencies and `go generate` is meant to be a tool run by developers rather than contributors or consumers.

build-scripts/pigeon
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -e
+docker run --rm -v $PWD:/workdir -w /workdir quay.io/influxdb/pigeon:2ad051b "$@"
```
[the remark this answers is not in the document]
jsternberg: I think publishing is a better idea because it means that it is easier for developers to get off the ground. Only the person who modifies the dependency has to worry about the actual build or if you are on an airplane (in which case this probably wouldn't work anyway).

I do think we'll need to add something to CI, but I wanted to keep it out of this PR as you mentioned because it would be too large of a scope.

tools/pigeon/Dockerfile
jsternberg: One reason against build args is because I find that the scripts begin to depend on the build arg and the actual dockerfile becomes ridiculously out of date. I do understand the concern about specifying the revision in multiple places though.

Also, some of these tools are going to be specific versions. The pigeon dependency was weird and we're removing it in the future anyway.

jsternberg:
I'll see into how we can unify the directories. I really wanted to put the scripts in `bin`, but we already put built objects in there. Are you fine with having the scripts in the bin folder and explicitly including them like this?

```
# .gitignore
/bin
!/bin/pigeon
```

query/promql/gen.go
```diff
@@ -1,3 +1,3 @@
 package promql
 
-//go:generate pigeon -o promql.go promql.peg
+//go:generate ../../build-scripts/pigeon -o promql.go promql.peg
```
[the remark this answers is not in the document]
jsternberg: I did not run this using the makefile. I tested by manually invoking `go generate` on the directory.

If we rely on the path and have the Makefile set the path, then we have to rely on the Makefile and we cannot use `go generate` without manually setting the path in our local directories which adds another requirement.

mark-rushakoff:
The explicit includes make sense to me. Could be top-level like `bin/pigeon`, or maybe `bin/stubs/pigeon`. I think the stuff that `go generate` produces ends up in `bin/$GOOS/$GOARCH/<tool>`.

nathanielc:
> Are you fine with having the scripts in the bin folder and explicitly including them like this?

This is what we currently do.

jsternberg:
@nathanielc we don't commit anything in the bin folder. I was asking if it was ok to explicitly include them so that the scripts in the bin folder would be committed.

query/promql/gen.go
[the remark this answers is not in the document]
nathanielc: Let me rephrase, this means that the pigeon binary has to exist in a specific location. And since the Makefile places it in that location I connected the two.

The Makefile only sets the PATH for the commands it runs. Relying on PATH means that you can install pigeon anywhere you want, so long as its in your path you can run `go generate`. It won't require the developers to mess with their PATH anymore than they want.

jsternberg:
@nathanielc the module syntax doesn't work since pigeon itself isn't a module. I'll switch the build to use 1.11 anyway though.

nathanielc:
@jsternberg I just ran those commands locally and they worked to install the specific version of pigeon.

jsternberg:
@nathanielc I get the following:

```
jsternberg:[~] $ go get github.com/mna/pigeon@2ad051b8b508f69e7dcf8febdea2856ec2db73ed
go: cannot use path@version syntax in GOPATH mode
jsternberg:[~] $ env GO111MODULE=on go get github.com/mna/pigeon@2ad051b8b508f69e7dcf8febdea2856ec2db73ed
go: cannot find main module; see 'go help modules'
```

If I run that command from within the platform repository, it uses the `Gopkg.lock` to generate a go.mod file, but that's incorrect behavior for our purposes.

nathanielc:
The `go.mod` file has to exist if you want to use modules. So it looks like this change needs to be coupled with the change to use modules? This change was intended to help pave the way so maybe they just need to happen together?

> We commit all dependencies and go generate is meant to be a tool run by developers rather than contributors or consumers.

I consider `go generate` as part of the "build" process. Meaning that this change means that Docker is required to run `go generate`. I understand that most contributors will not need to run `go generate` as they are not touching those bits frequently. I still do not want Docker to be a hard dep of any step in the build process including `go generate`.  Like I said above I support using Docker to provide an env for easily running `go generate` and `go build`, but Docker should not be required.

jsternberg:
We won't be doing this at the moment.

PR #735, opened 2018-08-29
feat: switch to using go modules for building

jsternberg:
This updates the dependency manager to use go modules instead of dep so that we can remove dep as a dependency and we can begin using and giving feedback on modules within this project. It should simplify dependency management and make it faster to run the builds too.

The big change is updating the various Makefiles to stop relying on the vendor directory because it no longer exists. This change creates a `tools.go` file with the revisions of the tools we use that are currently in the `required` section of `Gopkg.toml`. We are currently in other discussions to modify how we handle tool dependency management, but this change does not change that.


nathanielc:
Looks pretty straightforward. A couple of questions:

* What is the purpose of the `go.sum` file?
* Are the vendored dirs ignored now? Or is that what `go.sum` is about?


The tools.go looks good and makes sense.

jsternberg:
The `go.sum` file is documented in [here](https://github.com/golang/go/wiki/Modules).

Vendored directories were never committed in this repository.

jsternberg:
One possible negative for go modules is it doesn't seem to strip dependencies for dependent packages you don't use. So it adds all of the dependencies for `influxdb` regardless of if they're used by any packages or not as indirect dependencies.

I don't think that affects any decisions, but it might be something we can send upstream as a potential bug. You end up seeing it when you run `go mod tidy` which I don't want to add to this PR.

Makefile
```diff
@@ -20,10 +20,10 @@ GO_ARGS=-tags '$(GO_TAGS)'
 
 # Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=go build $(GO_ARGS)
-export GO_TEST=go test $(GO_ARGS)
+export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
+export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
 export GO_GENERATE=go generate $(GO_ARGS)
```
mark-rushakoff: Maybe add a comment why `go generate` doesn't need `GO111MODULE=on`? I think I know why, but this may be confusing for other readers too.

tools.go
```diff
@@ -0,0 +1,10 @@
+//+build tools
+
```
mark-rushakoff: Maybe add a comment here explaining that this is a workaround to get tooling dependencies? Per https://github.com/golang/go/issues/25922, maybe this will eventually become common knowledge, but until then it seems worthy of a comment.

README.md
```diff
@@ -2,13 +2,10 @@
 
 This is the [monorepo](https://danluu.com/monorepo/) for InfluxData Platform, a.k.a. Influx 2.0 OSS.
 
-## Vendoring
+## Installation
 
-This project is using [`dep`](https://golang.github.io/dep/docs/introduction.html) for vendoring.
+This project requires Go 1.11 and Go module support. Set `GO111MODULE=on` or build the project outside of your `GOPATH` for it to succeed.
```
mark-rushakoff: IIUC, running `go build` will just automagically get the dependencies, i.e. there is no analog to `dep ensure -vendor-only`. 

Since we're all pretty new to Go modules, can you add a note here about how to update a dependency? Maybe even linking to the go modules docs would suffice.

.circleci/config.yml
```diff
@@ -72,17 +59,10 @@ jobs:
       - image: circleci/golang:1.11-node-browsers
     environment:
       GOCACHE: /tmp/go-cache
-      DEPCACHEDIR: /tmp/dep-cache
     working_directory: /go/src/github.com/influxdata/platform
     steps:
       - checkout
 
-      # Speed up `make nightly` by restoring caches from previous runs.
```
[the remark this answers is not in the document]
jsternberg: My apologies. I thought it referred to that specific entry, but it applies to the entire section.

Makefile
jsternberg: I just kind of assumed it didn't. I don't actually know.

README.md
jsternberg: I added a link to the wiki. Unfortunately, I need more experience myself before offering further advice for how to use go modules. I think I have a general idea and, to my knowledge, it all works automagically, but this is a learning process for everyone.

tools.go
jsternberg: Done.

README.md
mark-rushakoff: I think this part of `go help modules` is the answer:

```
	go get github.com/gorilla/mux@latest    # same (@latest is default for 'go get')
	go get github.com/gorilla/mux@v1.6.2    # records v1.6.2
	go get github.com/gorilla/mux@e3702bed2 # records v1.6.2
	go get github.com/gorilla/mux@c856192   # records v0.0.0-20180517173623-c85619274f5d
	go get github.com/gorilla/mux@master    # records current meaning of master
```

But let's leave it with the link to the wiki until we're all sure.

Makefile
```diff
@@ -20,10 +20,11 @@ GO_ARGS=-tags '$(GO_TAGS)'
 
 # Test vars can be used by all recursive Makefiles
 export GOOS=$(shell go env GOOS)
-export GO_BUILD=go build $(GO_ARGS)
-export GO_TEST=go test $(GO_ARGS)
+export GO_BUILD=env GO111MODULE=on go build $(GO_ARGS)
+export GO_TEST=env GO111MODULE=on go test $(GO_ARGS)
```
mark-rushakoff: For `go build` and `go test` I think we want to use the `-mod=readonly` flag.

```
If invoked with -mod=readonly, the go command is disallowed from the implicit
automatic updating of go.mod described above. Instead, it fails when any changes
to go.mod are needed. This setting is most useful to check that go.mod does
not need updates, such as in a continuous integration and testing system.
The "go get" command remains permitted to update go.mod even with -mod=readonly,
and the "go mod" commands do not take the -mod flag (or any other build flags).
```

jsternberg: I actually thought we would want to not use those since the automatic update behavior is correct for local development. I'm not sure about CI, but I would be inclined to think that it's fine to let it modify `go.mod` if it needs to. I'm of the opinion that it's better to let it do that and then fail the build if it actually updates `go.mod` rather than hide potential issues by using read only.

mark-rushakoff: We could set `GOFLAGS=-mod=readonly` in the Circle config, and then local use of the Makefile is free to modify go.mod, but CI would fail if a required change wasn't checked in. Of course, that's assuming that we want to enforce that go.mod is up to date. I think that's desirable, but I don't have enough experience with go modules yet to say so definitively.

GOFLAGS is new in 1.11.

```
GOFLAGS
	A space-separated list of -flag=value settings to apply
	to go commands by default, when the given flag is known by
	the current command. Flags listed on the command-line
	are applied after this list and therefore override it.
```

jsternberg: I'll set it and we'll see if it's appropriate or not later.

jsternberg:
@nathanielc I'll start working on one today to replace the dep docs.

Makefile
```diff
@@ -67,23 +68,20 @@ $(CMDS): $(SOURCES)
...
 
-bin/$(GOOS)/protoc-gen-gogofaster: vendor $(call go_deps,./vendor/github.com/gogo/protobuf/protoc-gen-gogofaster)
-	$(GO_BUILD) -i -o $@ ./vendor/github.com/gogo/protobuf/protoc-gen-gogofaster
+bin/$(GOOS)/protoc-gen-gogofaster: go.mod go.sum
+	$(GO_BUILD) -o $@ github.com/gogo/protobuf/protoc-gen-gogofaster
 
-bin/$(GOOS)/goreleaser: ./vendor/github.com/goreleaser/goreleaser/main.go
-	go build -i -o $@ ./vendor/github.com/goreleaser/goreleaser
+bin/$(GOOS)/goreleaser: go.mod go.sum
+	GO111MODULE=on go build -o $@ github.com/goreleaser/goreleaser
```
goller: seems like all these go builds could be $(GO_BUILD)

jsternberg: I agree, but they weren't previously which is why I left it as-is.

jsternberg: I updated this anyway.

e-dard:
@jsternberg some of us talked about this a little briefly yesterday. I think in general everyone is for `go modules`, but my main concern was (1) something in go modules might break and then slow us down if all projects get borked; (2) the developer load of learning something new across all projects; (3) documentation specifying how we will pin dependencies to specific releases when they're still pre-GA, and generally how we'll upgrade stuff. 

(3) I'm not too worried about as it's just writing more docs 😄 

Personally I would like to see us switch to go modules on smaller projects that are not dependent on other things, e.g., the new `flux` repo first and then work our way up to the bigger ones.

Having said all that, I think there is a majority consensus to move forward with go modules on most projects, so I'm happy to go with the flow. Just throwing in my two penneth. 

@nathanielc  @stuartcarnie  you haven't approved yet: what do you think?

nathanielc:
Yes, we are planning to use `go modules` for the flux repo as test bed. As for concern (2) about learning a new tool, if the flux team takes the plunge we can learn and share that knowledge. 

The `go modules` workflow is being documented [here](https://github.com/influxdata/idpe/pull/1554) Maybe we let that one be merged first and then updated as a result of experimenting in the flux repo. Then once we are not learning anything new about go modules as a flux team we then merge this PR.
Say after a week or so of using go modules in the flux repo without having to update the docs we call the experiment done and start migrating other projects?

jsternberg:
I think that likely works perfectly. I'm making a last modification to the initial PR for that document to show how to use vendor to debug dependencies. Once I'm done, we should get that merged.

We have an all hands in less than 2 weeks for the engineering team. Based on conversations I have had with @goller, overheard conversations in the Slack channel, and my own personal experience, the actual usage of modules should be very easy and relatively straightforward. They are still doing planning for what we will cover during the all hands, but in a conversation with @goller yesterday, both of us seemed to agree that having a short workshop to get people compiling using go modules in 1.11 during the all hands would be most effective. If we make the change with people remote, they might have questions, but if we just sit down for 15-30 minutes on one of the days, merge the pull request, and ask everyone to just try using it and ensuring their workflow works, we can work through any problems pretty quickly and answer any questions.

It seems like using flux as a test bed for that fits with the timing perfectly.

PR #736, opened 2018-08-29
Implement query.ResultIterator on influxql.Response

desa:

This PR implements `query.ResultIterator` on `influxql.Response`. The motivation behind this is so that chronograf can issue queries to v1.X InfluxDB instances and return 2.X style csv results.

  - [x] Rebased/mergeable
  - [x] Tests pass

TODO:

- [x] Fix comments

### DIFF
--- a/query/influxql/response_iterator.go
+++ b/query/influxql/response_iterator.go
@@ -1,0 +1,361 @@
+package influxql
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"time"
+
+	"github.com/influxdata/platform/query"
+	"github.com/influxdata/platform/query/execute"
+	"github.com/influxdata/platform/query/values"
+)
+
+// responseIterator implements query.ResultIterator for a Response.
+type responseIterator struct {
+	response  *Response
+	resultIdx int
+}
+
+// NewresponseIterator constructs a responseIterator from a query.ResultIterator.
+func NewResponseIterator(r *Response) query.ResultIterator {
+	return &responseIterator{
+		response: r,
+	}
+}
+
+// More returns true if there are results left to iterate through.
+// It is used to implement query.ResultIterator.
+func (r *responseIterator) More() bool {
+	return r.resultIdx < len(r.response.Results)
+}
+
+// Next retrieves the next query.Result.
+// It is used to implement query.ResultIterator.
+func (r *responseIterator) Next() query.Result {
+	res := r.response.Results[r.resultIdx]
+	r.resultIdx++
+	return newQueryResult(&res)
+}
+
+// Cancel is a noop.
+// It is used to implement query.ResultIterator.
+func (r *responseIterator) Cancel() {}
+
+// Err returns an error if the response contained an error.
+// It is used to implement query.ResultIterator.
+func (r *responseIterator) Err() error {
+	if r.response.Err != "" {
+		return fmt.Errorf(r.response.Err)
+	}
+
+	return nil
+}
+
+// seriesIterator is a simple wrapper for Result that implements query.Result and query.TableIterator.
+type seriesIterator struct {
+	result *Result
+}
+
+func newQueryResult(r *Result) *seriesIterator {
+	return &seriesIterator{
+		result: r,
+	}
+}
+
+// Name returns the results statement id.
+// It is used to implement query.Result.
+func (r *seriesIterator) Name() string {
+	return strconv.Itoa(r.result.StatementID)
+}
+
+// Tables returns the original as a query.TableIterator.
+// It is used to implement query.Result.
+func (r *seriesIterator) Tables() query.TableIterator {
+	return r
+}
+
+// Do iterates through the series of a Result.
+// It is used to implement query.TableIterator.
+func (r *seriesIterator) Do(f func(query.Table) error) error {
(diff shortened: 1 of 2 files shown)


query/influxql/response.go
```diff
@@ -42,3 +52,303 @@ type Row struct {
...
+	if len(r.Values) < 1 {
+		panic("must have at least one value")
+	}
+	data := r.Values[0]
+	for i := range r.Columns {
+		v := data[i]
+		if colMeta[i].Label == "_time" {
+			continue
+		}
+		switch v.(type) {
```
fntlnz: what if the type does not fall in one of the cases?

desa: it will remain type `query.TInvalid` https://github.com/influxdata/platform/pull/736/files/aef1d7eec089c81da23ff6449a714255762aad5c#diff-230d894fa0934ba9ae817f834db1c8faR193

query/influxql/response.go
```diff
@@ -42,3 +52,303 @@ type Row struct {
...
+// Len returns the length or r.Values
+// It is used to implement query.ColReader.
+func (r *Row) Len() int {
+	return len(r.Values)
+}
+
+// Bools returns the values in column index j as bools.
+// It will panic if any value in the column is not a bool.
+// It is used to implement query.ColReader.
+func (r *Row) Bools(j int) []bool {
```
jsternberg: It's expected that this method should be fast. This is far too much memory allocation. Since the JSON response format is row-oriented instead of column-oriented, you are going to have to convert it between those when you read a series.

An interesting thing though is that you should be able to reuse memory. Each `query.ColReader` instance is only valid for the scope of the method that gets passed to it so you can reuse the same buffers.

query/influxql/response.go
```diff
@@ -8,12 +18,12 @@ package influxql
 
 type Response struct {
 	Results []Result `json:"results,omitempty"`
-	Err     string   `json:"error,omitempty"`
+	Error   string   `json:"error,omitempty"`
```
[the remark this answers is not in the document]
desa: > Is there a reason for this change?

The only reason this changed was because I implemented `query.Resulterator` on `Response` and it has an `Err()` method. Happy to implement it on another struct instead. My reasoning for not doing this was to avoid an unnecessary copy of, but happy to change that if you think it's best.

query/influxql/response.go
```diff
@@ -42,3 +52,303 @@ type Row struct {
...
+		}
+	}
+
+	return nil
+}
+
+// Key constructs the query.GroupKey for a Row from the rows
+// tags and measurement.
+// It is used to implement query.Table and query.ColReader.
+func (r *Row) Key() query.GroupKey {
```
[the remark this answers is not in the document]
desa: I was unaware that this call needed to be memoized. Thought that it would only be called once. Good to know thought 👍

query/influxql/response.go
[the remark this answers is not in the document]
jsternberg: I don't think you need to make a copy. Keeping a pointer to the `Response` structure is fine.

desa:
@jsternberg this should be ready for another review.

query/influxql/response_iterator.go
```diff
@@ -0,0 +1,361 @@
...
+	"sort"
+	"time"
+
+	"github.com/influxdata/platform/query"
+	"github.com/influxdata/platform/query/execute"
+	"github.com/influxdata/platform/query/values"
+)
+
+// ResponseIterator implements query.ResultIterator for a Response.
+type ResponseIterator struct {
```
jsternberg: I like this a lot better. Can you rename this `resultIterator` and not expose it?

query/influxql/response_iterator.go
```diff
@@ -0,0 +1,361 @@
...
+	response  *Response
+	resultIdx int
+}
+
+// NewResponseIterator constructs a ResponseIterator from a Response.
+func NewResponseIterator(r *Response) *ResponseIterator {
```
jsternberg: This should return `query.ResultIterator`.

query/influxql/response_iterator.go
```diff
@@ -0,0 +1,361 @@
...
+// It is used to implement query.ResultIterator.
+func (r *ResponseIterator) Err() error {
+	if r.response.Err != "" {
+		return fmt.Errorf(r.response.Err)
+	}
+
+	return nil
+}
+
+// queryResult is a simple wrapper for Result that implements query.Result and query.TableIterator.
```
jsternberg: Can you rename this to `seriesIterator` since it iterates over the series? Everything else is perfect.

query/influxql/response_iterator.go
```diff
@@ -0,0 +1,361 @@
...
+func newQueryResult(r *Result) *queryResult {
+	return &queryResult{
+		result: r,
+	}
+}
+
+// Name returns the results statement id.
+// It is used to implement query.Result.
+func (r *queryResult) Name() string {
+	return fmt.Sprintf("%d", r.result.StatementID)
```
jsternberg: `strconv.Itoa` is probably better. The `Sprintf` requires the code to parse the string. Even if it isn't significant and doesn't matter, it would make me feel better.

Issue #249, opened 2018-07-03
Specify type inference rules


adamperlin:
From ifql created by [nathanielc](https://github.com/nathanielc) : influxdata/ifql#315  

A type defines the set of values and operations on those values.
Types are never explicitly declared as part of the syntax.
Types are always inferred from the usage of the value.

Type inference needs to enable polymorphic behavior of functions, and allow for describing such behavior in function signatures.

For example we would like to be able to call a function `alert` who's message argument is either a string or a function that returns a string and accepts a record as its only parameter.

```
m = "Message prefix"

alert(message: (r) =>  "{m}: {r.host} is {r._value}")
alert(message: m)

alert(message: 42) // this would still need to be invalid
```

nathanielc:
Adding some details here:

There are several different ways we want to make use of type inference. So far I have reduced it to three different cases:

1. polymorphism - We want functions to be able to operate on many different types
    For example the type conversion functions `v` parameter can take multiple types

       f = float(v:"42.56")
2. Parameterized polymorphism (generic) - We want Flux functions to be able to operate on any type given their function body allows it.
    For example

       add = (a,b) => a + b
    The `add` function should work for all types where the `+` operator is defined, (uint, int, float, string)
3. Specialization - We want functions to be able to consume different types and behave differently.

       m = "Message prefix"

       alert(message: (r) =>  "{m}: {r.host} is {r._value}")
       alert(message: m)

       alert(message: 42) // this would still need to be invalid


Maybe 1 and 3 are the same?

PR #756, opened 2018-08-30
define new map type

jlapacik:
Related to #212 

This PR adds a new _map type_ to the SPEC. A _map type_ represents an immutable collection of key value pairs which cannot be accessed once it is declared. 

_What was the problem?_

It would be useful to have a data type in Flux for holding a set of key value pairs, where a key can be any expression that evaluates to a string and a value can be an arbitrary expression. For instance, the `to` function requires such a data type

https://github.com/influxdata/platform/blob/master/query/docs/SPEC.md#to

However it is not possible to have both of the above requirements while having the language remain statically typed.

_What was the solution?_

The solution was to create a _map type_ that cannot be explicitly accessed, in the source text, after it is instantiated. This data type is only useful as an input to a built-in function, as built-in functions are not interpreted.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/docs/SPEC.md
+++ b/query/docs/SPEC.md
@@ -405,12 +405,25 @@ The regular expression type name is `regexp`.
 An _array type_ represents a sequence of values of any other type.
 All values in the array must be of the same type.
 The length of an array is the number of elements in the array.
+The values of an array may be retrieved using member expressions.
 
 #### Object types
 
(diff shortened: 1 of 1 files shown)


nathanielc:
We need to first understand type inference before we can tackle key expressions for objects. #249

PR #757, opened 2018-08-31
test(task): add platform adapter tests

mark-rushakoff:
Also fix a handful of segfaults caused by improperly retained byte slices
from bolt, and combine two view transactions into one to avoid a data
race when a delete happened between them.

### DIFF
--- a/task/backend/storetest/logstoretest.go
+++ b/task/backend/storetest/logstoretest.go
@@ -44,9 +44,10 @@ func updateRunState(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFun
 		ID:  platform.ID([]byte("ab01ab01ab01ab01")),
 		Org: platform.ID([]byte("ab01ab01ab01ab05")),
 	}
-	scheduledFor := time.Unix(1, 0)
+	scheduledFor := time.Unix(1, 0).UTC()
 	run := platform.Run{
 		ID:           platform.ID([]byte("run")),
+		TaskID:       task.ID,
 		Status:       "scheduled",
 		ScheduledFor: scheduledFor.Format(time.RFC3339),
 	}
@@ -65,7 +66,7 @@ func updateRunState(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFun
 		t.Fatalf("expected: %+v, got: %+v", run, returnedRun)
 	}
 
-	startAt := time.Unix(2, 0)
+	startAt := time.Unix(2, 0).UTC()
 	if err := writer.UpdateRunState(context.Background(), task, run.ID, startAt, backend.RunStarted); err != nil {
 		t.Fatal(err)
 	}
@@ -81,7 +82,7 @@ func updateRunState(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFun
 		t.Fatalf("expected: %+v, got: %+v", run, returnedRun)
 	}
 
-	endAt := time.Unix(3, 0)
+	endAt := time.Unix(3, 0).UTC()
 	if err := writer.UpdateRunState(context.Background(), task, run.ID, endAt, backend.RunSuccess); err != nil {
 		t.Fatal(err)
 	}
@@ -109,11 +110,12 @@ func runLogTest(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFunc) {
 
 	run := platform.Run{
 		ID:           platform.ID([]byte("run")),
+		TaskID:       task.ID,
 		Status:       "scheduled",
-		ScheduledFor: time.Now().Format(time.RFC3339),
+		ScheduledFor: time.Now().UTC().Format(time.RFC3339),
 	}
 
-	logTime := time.Now()
+	logTime := time.Now().UTC()
 
 	if err := writer.AddRunLog(context.Background(), task, run.ID, logTime, "bad"); err == nil {
 		t.Fatal("shouldn't be able to log against non existing run")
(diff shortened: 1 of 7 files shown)


task/backend/storetest/logstoretest.go
```diff
@@ -109,11 +110,12 @@ func runLogTest(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFunc) {
 
 	run := platform.Run{
 		ID:           platform.ID([]byte("run")),
+		TaskID:       task.ID,
 		Status:       "scheduled",
-		ScheduledFor: time.Now().Format(time.RFC3339),
+		ScheduledFor: time.Now().UTC().Format(time.RFC3339),
```
lyondhill: After taking a quick glance at RFC3339 `Because the daylight saving rules for local time zones are so
   convoluted and can change based on local law at unpredictable times,
   true interoperability is best achieved by using Coordinated Universal
   Time (UTC).  This specification does not cater to local time zone
   rules.`
I dont think it is worth it to use `UTC()` since the Format handles that.

task/backend/storetest/logstoretest.go
```diff
@@ -135,7 +137,7 @@ func runLogTest(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFunc) {
 		t.Fatal(err)
 	}
 
-	fmtLogTime := logTime.Format(time.RFC3339)
+	fmtLogTime := logTime.UTC().Format(time.RFC3339)
```
lyondhill: Here it's especially pointless since logTime is a UTC already.

task/platform_adapter.go
```diff
@@ -28,6 +28,12 @@ func (p pAdapter) FindTaskByID(ctx context.Context, id platform.ID) (*platform.T
 	if err != nil {
 		return nil, err
 	}
+
+	// The store interface specifies that a returned task is nil if the operation succeeded without a match.
+	if t == nil {
```
lyondhill: the comment on the store interface says `If no task matches the ID, the returned task is nil.` I agree with that but it should be necessary for the store to return a 'not found' (or some error) when it cant find a task.

docmerlin: ^^^

task/backend/storetest/logstoretest.go
```diff
@@ -218,7 +220,7 @@ func listRunsTest(t *testing.T, crf CreateRunStoreFunc, drf DestroyRunStoreFunc)
 	scheduledFor, _ := time.Parse(time.RFC3339, runs[34].ScheduledFor)
 	listRuns, err = reader.ListRuns(context.Background(), platform.RunFilter{
 		Task:      &task.ID,
-		AfterTime: scheduledFor.Add(-1 * time.Nanosecond).Format(time.RFC3339),
+		AfterTime: scheduledFor.Add(-time.Second).Format(time.RFC3339),
```
docmerlin: I'm not sure I understand why the switch to time.Second.

mark-rushakoff: Because the time.RFC3339 format doesn't print nanoseconds:

```
println(now.Format(time.RFC3339))                      // 2009-11-10T23:00:00Z
println(now.Add(time.Nanosecond).Format(time.RFC3339)) // 2009-11-10T23:00:00Z
println(now.Add(time.Second).Format(time.RFC3339))     // 2009-11-10T23:00:01Z
```

task/backend/storetest/logstoretest.go
mark-rushakoff: I don't understand what you're saying. `t.Format(time.RFC3339)` includes time zone information. https://play.golang.org/p/EsRZWvXl_9S

```go
package main

import (
	"time"
)

func main() {
	now := time.Now()
	println(now.Format(time.RFC3339))

	loc := time.FixedZone("UTC-8", -8*60*60)
	println(now.In(loc).Format(time.RFC3339))
}

// Output:
// 2009-11-10T23:00:00Z
// 2009-11-10T15:00:00-08:00
```

Adding the UTC call keeps the formatting uniform.

task/platform_adapter.go
mark-rushakoff: Are you saying we should change the store interface to return an error, or we should change PlatformAdapter and the task.Service interface to return an error in that case?

If it's the former, that should be a separate PR. If it's the latter, I can update this PR.

task/backend/storetest/logstoretest.go
lyondhill: OO your right. As I read the RFC it made me believe anything that complied with the RFC would strip time zone and assume UTC. I now understand that it doesn't strip timezone but instead just measures off of UTC. Because `2009-11-10T23:00:00Z` is identical to `2009-11-10T15:00:00-08:00` it really doesn't matter if we call the `UTC()` function or not.

Im fine with leaving it in

task/platform_adapter.go
lyondhill: the store is already returning an error https://github.com/influxdata/platform/pull/757/files/a8269bc81cd6d477da8ebcddae8f37cfae40d44f#diff-e6f25cdcb7e9e982dd87211b3950b6c3R27 
What I'm saying is if any store ever `return nil, nil` on a `FindTaskByID` it should be caught in the task store test not here.

mark-rushakoff: I hit this branch in at least one of the added tests in this PR. We need to check `t == nil` and return early, so we don't evaluate `toPlatformTask(*t)` below for a nil `t`.

Or we need to change `platform/task/backend.Store` from `return nil, nil` on no match, to `return nil, ErrTaskNotFound` on no match. I'm fine doing that in a separate PR. I believe the `nil, nil` decision earlier was completely arbitrary in the first place.

lyondhill: I agree. That should be a separate PR. As for catching it hear, it really doesn't hurt anything. I am fine with catching it here and leaving it even when we update the store tests to not allow nil, nil.

task/backend/storetest/logstoretest.go
docmerlin: thanks.

PR #771, opened 2018-09-04
feat(http): add flux endpoint /v2/query influx

goller:
Closes #

Add /v2/query to handle flux queries

_What was the problem?_
Queries are important to a database.

_What was the solution?_

We added a handler for flux queries.  Currently, the handler just prints out the queries. 
Adding the real flux engine probably requires storage.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/query.go
+++ b/http/query.go
@@ -91,7 +91,10 @@ func (r QueryRequest) Validate() error {
 }
 
 // ProxyRequest returns a request to proxy from the query.
-func (r QueryRequest) ProxyRequest() *query.ProxyRequest {
+func (r QueryRequest) ProxyRequest() (*query.ProxyRequest, error) {
+	if err := r.Validate(); err != nil {
+		return nil, err
+	}
 	// Query is preferred over spec
 	var compiler query.Compiler
 	if r.Query != "" {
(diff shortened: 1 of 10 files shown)


http/query_handler.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// PrometheusCollectors satisifies the prom.PrometheusCollector interface.
+func (h *FluxHandler) PrometheusCollectors() []prometheus.Collector {
+	// TODO: gather and return relevant metrics.
+	return nil
+}
+
+// FluxService connects to Influx via HTTP using tokens to run queries.
+type FluxService struct {
```
nathanielc: Why is this called the FluxService? It is implementing the ProxyQueryService interface.

nathanielc: This implementation is also identical to the `ProxyQueryService` that already exists in the http package.

fntlnz:
I'm not sure if the source implementation still works now that the decoding behavior is changed in `external_query_handler.go`

http/query_handler.go
goller: @nathanielc I should have mentioned this in the PR comment, but, I plan on renaming FluxService to ProxyQueryService once ProxyQueryService and others are moved into idpe.

I wanted to make this PR work in two parts: 1) add the implementation 2) move the other implementations

http/query.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf(`unknown dialect date time format: %s`, r.Dialect.DateTimeFormat)
+	}
+
+	return nil
+}
+
+// ProxyRequest returns a request to proxy from the query.
+func (r QueryRequest) ProxyRequest() *query.ProxyRequest {
+	// Query is preferred over spec
+	var compiler query.Compiler
```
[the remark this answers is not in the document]
goller: Fixed in 6307531

PR #773, opened 2018-09-04
build(chronograf): add conditional build tags for asset generation

### LINKED ISSUE #489: Decide on strategy for Chronograf assets
I had to merge a hotfix in #488 so that local builds would "just work", but that also involved disabling the call to `go generate` in the Makefile.

We probably don't want to check in the result of go-bindata because it's presumably minified javascript that we can regenerate when we need it.

Luckily, [go-bindata supports build tags](https://github.com/kevinburke/go-bindata#build-tags).

> With the optional -tags flag, you can specify any go build tags that must be fulfilled for the output file to be included in a build. This is useful when including binary data in multiple formats, where the desired format is specified at build time with the appropriate tags.

> The tags are appended to a // +build line in the beginning of the output file and must follow the build tags syntax specified by the go tool.

I would propose that we use a build tag like `// +build !chronograf` in the files I added in #488, and the use go-bindata with `-tags chronograf`. But I'm open to other ideas.

goller:
Closes #489 

Add tags for asset building

_What was the problem?_
Apparently, chronograf's assets were not being generated.

_What was the solution?_
Condition build flags are added if building from the Makefile.

  - [x] Rebased/mergeable
  - [x] Tests pass


Makefile
```diff
@@ -109,11 +109,10 @@ vendor/github.com/kevinburke/go-bindata/go-bindata/main.go: vendor
 fmt: $(SOURCES_NO_VENDOR)
 	goimports -w $^
 
-# generate:
-	# TODO: re-enable these after we decide on a strategy for building without running `go generate`.
-	# $(GO_GENERATE) ./chronograf/dist/...
-	# $(GO_GENERATE) ./chronograf/server/...
-	# $(GO_GENERATE) ./chronograf/canned/...
+generate:
```
mark-rushakoff: Maybe this should be renamed `chronograf_generate` so it isn't mistaken to `go generate ./...`?

goller: I'd rather have a generate target that will encompass all generation.  When we have more than one generate "type" we could split them into multiple targets.

This is for people to use... typing chronograf_generate is not so fun nor memorable.

PR #774, opened 2018-09-05
fix(taskd): Stop panic in options

lyondhill:
When type asserting a value we need to ensure it is the type we expect
before we assert the value.


task/options/options.go
```diff
@@ -182,3 +203,11 @@ func (o *Options) EffectiveCronString() string {
 	}
 	return ""
 }
+
+// checkKind panics if got != exp.
```
mark-rushakoff: Should be `returns an error`, not `panics`.

lyondhill: oo thanks

PR #781, opened 2018-09-06
feat(taskd): create a FindTaskByIDWithMeta function.

lyondhill:
Now we can retrieve both task data as well as meta in one call.


lyondhill:
Yep. Working on them now. Should have a PR ready soon.

PR #785, opened 2018-09-06
fix(query/influxql): join multiple aggregates with the new join function

jsternberg:


query/influxql/join.go
```diff
@@ -34,27 +29,12 @@ func Join(t *transpilerState, cursors []cursor, on, except []string) cursor {
 		tables[cur.ID()] = tableName
 
 		for _, k := range cur.Keys() {
-			// Generate a name for accessing this expression and generate the index entries for it.
-			name := fmt.Sprintf("val%d", len(exprs))
+			// Combine the table name with the name to access this attribute so we can know
+			// what it will be mapped to.
+			varName, _ := cur.Value(k)
+			name := fmt.Sprintf("%s_%s", tableName, varName)
```
jlapacik: What if `varName` is one of the join columns? If it is, then shouldn't `name = varName`?

jsternberg: The variables in the cursors should never be joined on to my knowledge.

jlapacik: Ok I didn't know that. Then this looks good.

PR #789, opened 2018-09-07
add types to bucket service tests

jademcgough:
This adds types to all buckets in our tests, as well as a new test to ensure that we set a default type on creation.


jademcgough:
@goller can I merge this?

PR #821, opened 2018-09-11
feat(query/influxql): normalize the time for all aggregate queries

jsternberg:
The transpiler will normalize the `_time` column by dropping any
existing time column and then duplicating `_start` when the query is an
aggregate type.

This works for the selectors because they did not normalize their
`_time` column at all and, while the aggregates did normalize their
`_time` column, we have made the decision to remove that functionality
and have aggregates not set a `_time` column at all.

Fixes influxdata/flux#14.


query/influxql/function.go
```diff
@@ -238,6 +237,16 @@ func createFunctionCursor(t *transpilerState, call *influxql.Call, in cursor) (c
 	default:
 		return nil, fmt.Errorf("unimplemented function: %q", call.Name)
 	}
+
+	// If we have been told to normalize the time, we do it here.
+	if normalize {
+		cur.id = t.op("duplicate", &functions.DuplicateOpSpec{
+			Col: execute.DefaultStartColLabel,
+			As:  execute.DefaultTimeColLabel,
+		}, t.op("drop", &functions.DropOpSpec{
```
aanthony1243: the new feature that drops `_time`  from aggregate calls is on its way.  I think this code might break at that point, I'm having a little trouble following when `normalize`  is true.

jsternberg: Does `drop` have an error if the column doesn't exist? I just assumed it didn't. Either way, I'm working on that aggregate behavior right now anyway.

PR #824, opened 2018-09-12
feat: add onboarding defaults

kelwang:
resolve https://github.com/influxdata/platform/issues/792

check if boltPath existed, before call client.Open()
if not, we assume, this is the first time start the single binary.
Then bolt client is opened, create default user, default bucket, default org, default auth.

print out the token


cmd/influxd/main.go
```diff
@@ -338,3 +349,54 @@ func Execute() {
...
+
+func isOnBoarding(boltPath string) bool {
+	if _, err := os.Stat(boltPath); os.IsNotExist(err) {
+		return true
+	}
+	return false
+}
+
+// create default org, bucket, user,
+func createDefaults(ctx context.Context, c *bolt.Client) error {
```
desa: What do you think about creating a special interface like 
https://github.com/influxdata/idpe/blob/7db8b8da1a071339e4877431785bebea5baa785d/gateway/backend.go#L14-L36
```go
type Backend interface {
	AuthorizationService() platform.AuthorizationService
	BucketService() platform.BucketService
	OrganizationService() platform.OrganizationService
	UserService() platform.UserService
}
```
that way we're not dependent on bolt.

cmd/influxd/main.go
```diff
@@ -338,3 +349,54 @@ func Execute() {
...
+				Action:   platform.WriteAction,
+			},
+			platform.WriteBucketPermission(bucket.ID),
+		},
+	}
+	err = c.CreateAuthorization(ctx, auth)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Token: %s \n\n", auth.Token)
```
desa: Hmm this seems a bit odd. What do you think about writing this to a file instead?

cmd/influxd/main.go
```diff
@@ -338,3 +349,54 @@ func Execute() {
...
+
+// create default org, bucket, user,
+func createDefaults(ctx context.Context, c *bolt.Client) error {
+	u := &platform.User{Name: "admin"}
+	err := c.CreateUser(ctx, u)
+	if err != nil {
+		return err
+	}
+	o := &platform.Organization{
+		Name: "default",
```
desa: What do you think about calling `default` something like `autogen`

cmd/influxd/main.go
```diff
@@ -338,3 +349,54 @@ func Execute() {
...
+func isOnBoarding(boltPath string) bool {
+	if _, err := os.Stat(boltPath); os.IsNotExist(err) {
+		return true
+	}
+	return false
+}
+
+// create default org, bucket, user,
+func createDefaults(ctx context.Context, c *bolt.Client) error {
+	u := &platform.User{Name: "admin"}
```
desa: What do you think about calling `admin` something like `autogen`?

cmd/influxd/main.go
```diff
@@ -338,3 +349,54 @@ func Execute() {
...
+	}
+	o := &platform.Organization{
+		Name: "default",
+	}
+	err = c.CreateOrganization(ctx, o)
+	if err != nil {
+		return err
+	}
+	bucket := &platform.Bucket{
+		Name:           "default",
```
desa: What do you think about calling `default` something like `autogen`?

cmd/influxd/main.go
kelwang: Both fine I think. 
It's part of production requirement,
https://github.com/influxdata/platform/issues/792#issuecomment-420336964
Maybe we can ask @russorat  about this.

desa: @russorat feelings on calling it default vs autogen?

bolt/onboarding.go
```diff
@@ -0,0 +1,67 @@
+package bolt
```
desa: Just had another thought.

What if we make the onboarding package generic. Something like

```go
package onboarding

type backend interface {
platform.BucketService
platform.OrganizationService
...
}

type Values struct {
Users []*platform.User
Organizations []*platform.Organization
Authorizations []*platform.Authorization
...
}


func Generate(b backend, vs Values) error {
...
}
```

This way we can describe the process generically

onboarding.go
```diff
@@ -0,0 +1,32 @@
...
+	Auth   Authorization `json:"auth"`
+}
+
+// OnboardingRequest is the request
+// to setup defaults.
+type OnboardingRequest struct {
+	UserName   string `json:"username"`
+	Password   string `json:"password"`
+	OrgName    string `json:"org"`
+	BucketName string `json:"bucket"`
```
desa: Why not make this `Org` and `Bucket`?

onboarding.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// OnboardingService represents a service for the first run.
+type OnboardingService interface {
+	// IsOnboarding determine if onboarding request is allowed.
+	IsOnboarding(ctx context.Context) bool
+	// OnboardingFinished will update the flag.
+	// Set it true, future onboarding request will be denied.
+	// Set it false, future onboarding request will allowed.
+	OnboardingFinished(ctx context.Context, v bool) error
```
desa: Does this need to be on the interface?

onboarding.go
```diff
@@ -0,0 +1,32 @@
...
+	// Generate OnboardingDefaults.
+	Generate(ctx context.Context, req OnboardingRequest) (*OnboardingDefaults, error)
```
desa: Why not have req be a pointer?

kelwang: Um, usually go pointer will be used in 3 scenarios. 1. the object could be nil. 2. the state of object could be changed. 3. easier serving a go interface. None of those applies to the req, req will be immutable for all the time, it can't nil either. So for the safety, I think we should use struct. I could be wrong, the only benefit I can think of using a pointer is saving a dereference action when performing json unmarshaling. But if you have other reason or feel strong about it, I can change it to a pointer. :)

onboarding.go
kelwang: Hi, @desa, maybe you can help me to figure out a better name, it's really just for testing. More like PUT method of the other services. What do you think?

onboarding.go
```diff
@@ -0,0 +1,32 @@
+package platform
+
+import "context"
+
+// OnboardingDefaults is a group of elements required for first run.
+type OnboardingDefaults struct {
+	User   User          `json:"user"`
+	Org    Organization  `json:"org"`
```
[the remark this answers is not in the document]
kelwang: I think they are all required field, and none of them should be nil or changed. What do you think?

http/onboarding.go
```diff
@@ -0,0 +1,81 @@
...
+	return h
+}
+
+// isOnboarding is the HTTP handler for the GET /setup route.
+// returns true/false
+func (h *SetupHandler) isOnboarding(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	result := h.OnboardingService.IsOnboarding(ctx)
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintln(w, result)
```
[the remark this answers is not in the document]
kelwang: Yeah, love it, thanks

onboarding.go
desa: I think it might be better to not have this on the interface at all. I'm not sure why we would ever need to set it back to false. Thoughts?

bolt/onboarding.go
```diff
@@ -0,0 +1,111 @@
...
+func (c *Client) initializeOnboarding(ctx context.Context, tx *bolt.Tx) error {
+	if _, err := tx.CreateBucketIfNotExists([]byte(onboardingBucket)); err != nil {
+		return err
+	}
+	return nil
+}
+
+// IsOnboarding checks onboardingBucket
+// to see if the onboarding key is true.
+func (c *Client) IsOnboarding(ctx context.Context) (isOnboarding bool) {
```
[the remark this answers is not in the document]
kelwang: Um, I think IsOnboarding  is less confusing, but I am not feeling strong. Are you feeling strong about Onboarding instead of IsOnboarding?

onboarding.go
kelwang: yeah, you are absolutely right

http/onboarding.go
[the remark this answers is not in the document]
kelwang: Changed, can you take a look again? thanks

bolt/onboarding.go
```diff
@@ -0,0 +1,111 @@
...
+		result := tx.Bucket(onboardingBucket).Get(onboardingKey)
+		isOnboarding = len(result) == 0
+		return nil
+	})
+	return isOnboarding
+}
+
+// OnboardingFinished will update the flag,
+// so future onboarding request will be denied.
+func (c *Client) OnboardingFinished(ctx context.Context, v bool) error {
```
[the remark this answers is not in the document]
kelwang: For testing, boltdb testing is under bolt_test package. Do you think we should create a PR to change bolt_test back to bolt?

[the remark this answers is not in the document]
desa: I think just keep it exported like you have it here. 👍

bolt/onboarding.go
```diff
@@ -0,0 +1,136 @@
...
+		result := tx.Bucket(onboardingBucket).Get(onboardingKey)
+		isOnboarding = len(result) == 0
+		return nil
+	})
+	return isOnboarding, err
+}
+
+// OnboardingFinished will update the flag,
+// so future onboarding request will be denied.
+func (c *Client) OnboardingFinished(ctx context.Context, v bool) error {
```
desa: *nit* maybe rename this function to `PutOnboardingStatus`

http/onboarding.go
```diff
@@ -0,0 +1,85 @@
...
+// returns true/false
+func (h *SetupHandler) isOnboarding(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	result, err := h.OnboardingService.IsOnboarding(ctx)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintln(w, `{"allowed": `+fmt.Sprintf("%v", result)+`}`)
```
desa: why not just do
```go
fmt.Fprintf(w, `{"allowed": %v}`,result)
```

onboarding.go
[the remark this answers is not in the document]
desa: I'd make them all pointers just so you don't have to dereference them elsewhere.

onboarding.go
```diff
@@ -0,0 +1,28 @@
...
+	Org      string `json:"org"`
+	Bucket   string `json:"bucket"`
+}
+
+// OnboardingService represents a service for the first run.
+type OnboardingService interface {
+	// IsOnboarding determine if onboarding request is allowed.
+	IsOnboarding(ctx context.Context) (bool, error)
+	// Generate OnboardingResults.
+	Generate(ctx context.Context, req OnboardingRequest) (*OnboardingResults, error)
```
desa: What do you think about making req a pointer?

PR #826, opened 2018-09-12
fix: Update REPL to use HTTP API

nathanielc:


cmd/influx/query.go
```diff
@@ -28,23 +27,10 @@ var queryCmd = &cobra.Command{
 }
 
 var queryFlags struct {
-	StorageHosts string
-	OrgID        string
-	Verbose      bool
+	OrgID string
```
goller: A nice alternative would be to have org name when one doesn't know the ID.

Perhaps as a follow-on issue?

PR #827, opened 2018-09-12
fix: Type alias config

nathanielc:
By using an alias we avoid the vendor within vendor issues when using this type downstream


nathanielc:
Turns out this was an artifact of a build issue and is not needed.

PR #829, opened 2018-09-13
fix(task): remove race condition when finishing a inmem run

lyondhill:


task/backend/inmem_store.go
```diff
@@ -286,10 +286,12 @@ func (s *inmem) CreateNextRun(ctx context.Context, taskID platform.ID, now int64
 
 // FinishRun removes runID from the list of running tasks and if its `now` is later then last completed update it.
 func (s *inmem) FinishRun(ctx context.Context, taskID, runID platform.ID) error {
+	s.mu.RLock()
 	stm, ok := s.runners[taskID.String()]
 	if !ok {
 		return errors.New("taskRunner not found")
```
mark-rushakoff: Need to RUnlock for the early return. Could just move the unlock before the `if !ok` line.

PR #834, opened 2018-09-14
dep: update xxhash

mark-rushakoff:
There's a more recent release that uses math/bits and has improved
benchmarks.


mark-rushakoff:
Closing this since go.mod now uses v1.1.0.

PR #841, opened 2018-09-14
Feat/http client tests

### LINKED ISSUE #825: json: cannot unmarshal object into Go value of type []*platform.Organization
to reproduce the issue. run the following command

```
$influx org find
```
will get
```
json: cannot unmarshal object into Go value of type []*platform.Organization
```

please also add tests for this

goller:
Closes #825

Add integration tests with HTTP client/server to test org/bucket/auth.

_What was the problem?_
Org and bucket HTTP client had incompatibility with server changes.

_What was the solution?_
This adds a test suite around client/server interactions to make refactoring auth/bucket/org easier.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/auth_test.go
+++ b/http/auth_test.go
@@ -478,9 +478,7 @@ func initAuthorizationService(f platformtesting.AuthorizationFields, t *testing.
 	client := AuthorizationService{
 		Addr: server.URL,
 	}
-	done := func() {
-		server.Close()
-	}
+	done := server.Close
 
 	return &client, done
 }
--- a/http/auth_test.go
+++ b/http/auth_test.go
@@ -234,7 +234,7 @@ func TestService_handleGetAuthorization(t *testing.T) {
 			r := httptest.NewRequest("GET", "http://any.url", nil)
 
 			r = r.WithContext(context.WithValue(
-				context.TODO(),
+				context.Background(),
 				httprouter.ParamsKey,
 				httprouter.Params{
 					{
@@ -416,7 +416,7 @@ func TestService_handleDeleteAuthorization(t *testing.T) {
 			r := httptest.NewRequest("GET", "http://any.url", nil)
 
 			r = r.WithContext(context.WithValue(
-				context.TODO(),
+				context.Background(),
 				httprouter.ParamsKey,
 				httprouter.Params{
 					{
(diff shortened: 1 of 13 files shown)


http/auth_test.go
```diff
@@ -443,3 +445,68 @@ func TestService_handleDeleteAuthorization(t *testing.T) {
...
+			 authorization by token string via headers or something
+		*/
+		t.Skip("TestAuthorizationService_FindAuthorizations/find_authorization_by_token skipped because user tokens cannot be queried")
+	}
+
+	svc := inmem.NewService()
+	svc.IDGenerator = f.IDGenerator
+	svc.TokenGenerator = f.TokenGenerator
+
+	ctx := context.TODO()
```
mark-rushakoff: Use `context.Background() `for tests:

https://golang.org/pkg/context/#Background

> Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

http/bucket_test.go
```diff
@@ -585,3 +586,55 @@ func TestService_handlePatchBucket(t *testing.T) {
...
+		}
+	}
+
+	handler := NewBucketHandler()
+	handler.BucketService = svc
+	server := httptest.NewServer(handler)
+	client := BucketService{
+		Addr: server.URL,
+	}
+	done := func() {
```
mark-rushakoff: Maybe could be `done := server.Close`. If the signature is `(*server).Close() error`, maybe would be more clear as `_ = server.Close()` so that this doesn't look like a missed simplification.

http/bucket_test.go
```diff
@@ -585,3 +586,55 @@ func TestService_handlePatchBucket(t *testing.T) {
 		})
 	}
 }
+
+func initBucketService(f platformtesting.BucketFields, t *testing.T) (platform.BucketService, func()) {
+	svc := inmem.NewService()
+	svc.IDGenerator = f.IDGenerator
+
+	ctx := context.TODO()
```
mark-rushakoff: `context.Background()`.

http/bucket_test.go
goller: Fixed in e0b1f89

goller:
@mark-rushakoff Added helper functions in 31a2ed1f to do entire service testing.

PR #864, opened 2018-09-25
feat(http): buckets now use influxql duration strings

### LINKED ISSUE #796: Update buckets retention policies to use duration strings

goller:
Note: I'll create a follow-on PR for changing the external format of the time duration from a string to an object.  I think this set of changes should be reviewed first before I plunge into an entirely different structure.

Closes #796

Change bucket retention periods from nanoseconds to influxql-style duration strings.

_What was the problem?_
Nanoseconds are difficult to use and a bit unpleasant to view.

_What was the solution?_
retention period is now a string in the influxql duration format.

  - [x] Rebased/mergeable
  - [x] Tests pass

/cc @gshif

### DIFF
--- a/http/duration.go
+++ b/http/duration.go
@@ -1,0 +1,133 @@
+package http
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/influxdata/platform/kit/errors"
+)
+
+// ErrInvalidDuration is returned when parsing a malformatted duration.
+var ErrInvalidDuration = errors.New("invalid duration", errors.MalformedData)
+
+// ParseDuration parses a time duration from a string.
+// This is needed instead of time.ParseDuration because this will support
+// the full syntax that InfluxQL supports for specifying durations
+// including weeks and days.
+func ParseDuration(s string) (time.Duration, error) {
+	// Return an error if the string is blank or one character
+	if len(s) < 2 {
+		return 0, ErrInvalidDuration
+	}
+
+	// Split string into individual runes.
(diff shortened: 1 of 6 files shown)


http/duration.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+// ErrInvalidDuration is returned when parsing a malformatted duration.
+var ErrInvalidDuration = errors.New("invalid duration", errors.MalformedData)
+
+// ParseDuration parses a time duration from a string.
+// This is needed instead of time.ParseDuration because this will support
+// the full syntax that InfluxQL supports for specifying durations
+// including weeks and days.
+func ParseDuration(s string) (time.Duration, error) {
```
desa: Does flux have a similar duration parser? If so, should we use that one?

http/duration.go
```diff
@@ -0,0 +1,157 @@
...
+			if i+1 < len(a) {
+				switch a[i+1] {
+				case 's': // ms == milliseconds
+					unit = string(a[i : i+2])
+					d += time.Duration(n) * time.Millisecond
+					i += 2
+					continue
+				case 'o': // mo == month
+					unit = string(a[i : i+2])
+					d += time.Duration(n) * 30 * 24 * time.Hour
```
nathanielc: This works for now, but let's add a TODO linking to this issue #657 to use the real duration type

http/duration.go
leodido: I've built a duration parser with ragel.

PR #865, opened 2018-09-25
fix: the master build no longer uses -i when building the influx binary

jsternberg:
The `-i` flag causes a problem when a cross compilation happens because
it attempts to build files in GOROOT which is located in a directory
that is not writable by the build user. This same problem exists for
when the `-tags` are changed like adding `-tags assets` in the `all`
build that runs on the master branch.

This removes `-i` because, in the latest go release, the `-i` flag
doesn't actually do anything useful since there is another build cache
that the results are placed in regardless of whether `-i` is used.

This fixes the build on master so that `make all` and `make nightly`
will function properly.


jsternberg:
As a note, a few other files changed since those files changed when I ran `make all`. We had not run `go generate` over promql and the output had changed and the `go.sum` file was missing a line that gets added when you call `go build` when that dependency is built.

PR #868, opened 2018-09-25
fix: goreleaser fork needed to have its version updated

jsternberg:


go.mod
```diff
@@ -35,7 +35,7 @@ require (
 	github.com/google/go-github v0.0.0-20180723152927-e1be32f26e66
 	github.com/google/go-querystring v0.0.0-20170111101155-53e6ce116135 // indirect
 	github.com/gopherjs/gopherjs v0.0.0-20180825215210-0210a2f0f73c // indirect
-	github.com/goreleaser/goreleaser v0.0.0-20180917235036-c23174192b3a
+	github.com/goreleaser/goreleaser v0.87.0
```
nathanielc: Why version 0.87 here but version 0.86 below?

jsternberg: I think it's a weird thing in modules. goreleaser recently released that version and so it shows up there when you ask for the latest one, but the replace below goes and changes it to our own version.

PR #873, opened 2018-09-25
chore: Do not use -i during build 

nathanielc:
Specifically the `-i` flag is now only needed when you actually want to install the dependencies, it is not used to speed up build times.

From the CL
```
Installing dependencies is no longer necessary for speed, it has confused
users in the past, and the more accurate rebuilds mean that it will confuse
users even more often in the future. This CL aims to end all that confusion
by not installing dependencies by default.
```
https://github.com/golang/go/commit/8f70e1f8a91db6c8a1bcc292766af95e52e8512b

If your build touches a standard lib package that needs to be installed it may try to install that to the system directories (depending on how Go was installed), which will likely cause a permission denied error.


nathanielc:
So this was already done here https://github.com/influxdata/platform/commit/2ee5308a27a4cfe8d2559427d9618de467e55efc#diff-b67911656ef5d18c4ae36cb6741b7965 and somehow the diff isn't showing it.

PR #879, opened 2018-09-26
Add kit/th

mark-rushakoff:
It's less of an issue in platform, but in enterprise, we've had to skip a lot of tests to satisfy CI, and they've gotten skipped with inconsistent messages. As a result, it's been difficult to grep for tests that are skipped due to flakiness.

This PR makes tests skipped due to flakiness more discoverable, and allows a way to opt out of those skips with an environment variable. It also adds a helper to retry a flaky test. Ideally the underlying cause of the flakiness would be addressed, but the majority of these skips are added by non-owners of the tests in question, whose PRs are blocked by CI failing as a result of their unrelated change. 

Now, non-owners can more easily handle flaky tests.


mark-rushakoff:
Per other discussions, we're not moving forward with special handling of flaky tests (other than prioritizing and deflaking them).

PR #891, opened 2018-09-26
Add routes that allow the user to exchange basic auth credentials for a session

### LINKED ISSUE #890: Add user session handler

desa:
Closes #890

This PR adds a session service that is intended to be used after a user has authenticated. A session should represent a users maximal privileges, although this has yet to be fully worked out. It contains an ID so that it will be auditable.

This PR adds `/signin` and `/signout` routes that allow the user to exchange basic auth credentials for a session.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/session_handler.go
+++ b/http/session_handler.go
@@ -1,0 +1,127 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/influxdata/platform"
+	"github.com/julienschmidt/httprouter"
+	"go.uber.org/zap"
+)
+
+// BasicAuthHandler represents an HTTP API handler for authorizations.
+type BasicAuthHandler struct {
+	*httprouter.Router
+	Logger *zap.Logger
+
+	BasicAuthService platform.BasicAuthService
+	SessionService   platform.SessionService
+}
+
+// NewBasicAuthHandler returns a new instance of BasicAuthHandler.
+func NewBasicAuthHandler() *BasicAuthHandler {
+	h := &BasicAuthHandler{
+		Router: httprouter.New(),
+	}
+
+	h.HandlerFunc("POST", "/signin", h.handleSignin)
+	h.HandlerFunc("POST", "/signout", h.handleSignout)
+	return h
+}
+
+// handleSignin is the HTTP handler for the POST /signin route.
+func (h *BasicAuthHandler) handleSignin(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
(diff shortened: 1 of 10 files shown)


kelwang:
I don't think store each session entry with expire time in boltdb is a good idea. Usually session should only end if no new activity. It also means we have to keep lock and unlock the boltdb for every users' activities.  I suggest we could have a sessionManager with 2 inmem sync.Map. One for time(each minute as key)->users, another one for user->time. The session Manager will try to expire the session every minute, same time user can update their own user->time index, then update the main index.  In a separated go routine, the boltdb will try to store the whole 2 indexes every minute, So when the application restart, user don't have to re-login.

desa:
@kelwang 

> Usually session should only end if no new activity. 

This depends on product requirements. There are plenty of services that maintain long lived sessions so that you don't have the re-authenticate. My thinking was that for the alpha it would be better for the user to not have to re-authenticate frequently. What do you think?

> It also means we have to keep lock and unlock the boltdb for every users' activities. I suggest we could have a sessionManager with 2 inmem sync.Map. One for time(each minute as key)->users, another one for user->time. The session Manager will try to expire the session every minute, same time user can update their own user->time index, then update the main index. 

I'm not too concerned with performance at the moment (authorizations have the same lookup path currently in platform), but when it become a concern we'll definitely need to do something like this. Maybe there's some way we can tweek the `SessionService` interface that will lend itself to this type of implementation?

kelwang:
Yeah, I agree it should be cool for now

http/session_handler.go
```diff
@@ -0,0 +1,127 @@
...
+	BasicAuthService platform.BasicAuthService
+	SessionService   platform.SessionService
+}
+
+// NewSessionHandler returns a new instance of SessionHandler.
+func NewSessionHandler() *SessionHandler {
+	h := &SessionHandler{
+		Router: httprouter.New(),
+	}
+
```
jademcgough: Should we document these endpoints in the swagger doc?

http/session_handler.go
```diff
@@ -0,0 +1,127 @@
...
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	if err := h.SessionService.ExpireSession(ctx, req.Key); err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	// TODO(desa): not sure what to do here maybe redirect?
```
jademcgough: IMO this is fine for now

http/session_handler.go
desa: yup

PR #894, opened 2018-09-27
add requestedAt to task log

### LINKED ISSUE #753: task: run log should store requestedAt correctly
Currently, `ScheduledFor` would only be set when on a call to UpdateRunState: 

https://github.com/influxdata/platform/blob/745376a5fe3f426aaeb7514b0cef1f634bb8afb5/task/backend/inmem_logreaderwriter.go#L32-L33

But in a separate discussion on the schema, I believe we concluded that we would treat `ScheduledFor` as a field on every point for a run's log entry. It seems like the API for task/backend.LogWriter needs to change to support this.

docmerlin:
This is the platform part of adding requestedAt to the task run log.
closes #753

### DIFF
--- a/task/backend/inmem_logreaderwriter.go
+++ b/task/backend/inmem_logreaderwriter.go
@@ -41,7 +41,15 @@ func (r *runReaderWriter) UpdateRunState(ctx context.Context, rlb RunLogBase, wh
 	if !ok {
 		tid := append([]byte(nil), rlb.Task.ID...)
 		sf := time.Unix(rlb.RunScheduledFor, 0).UTC()
-		run := &platform.Run{ID: rlb.RunID, TaskID: tid, Status: status.String(), ScheduledFor: sf.Format(time.RFC3339)}
+		run := &platform.Run{
+			ID:           rlb.RunID,
+			TaskID:       tid,
+			Status:       status.String(),
+			ScheduledFor: sf.Format(time.RFC3339),
+		}
+		if rlb.RequestedAt != 0 {
+			run.RequestedAt = time.Unix(rlb.RequestedAt, 0).UTC().Format(time.RFC3339)
+		}
 		timeSetter(run)
 		r.byRunID[ridStr] = run
 		tidStr := rlb.Task.ID.String()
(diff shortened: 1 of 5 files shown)


task/backend/inmem_logreaderwriter.go
```diff
@@ -34,14 +34,23 @@ func (r *runReaderWriter) UpdateRunState(ctx context.Context, rlb RunLogBase, wh
...
 		sf := time.Unix(rlb.RunScheduledFor, 0).UTC()
-		run := &platform.Run{ID: rlb.RunID, TaskID: tid, Status: status.String(), ScheduledFor: sf.Format(time.RFC3339)}
+		run := &platform.Run{
+			ID:           rlb.RunID,
+			TaskID:       tid,
+			Status:       status.String(),
+			ScheduledFor: sf.Format(time.RFC3339),
+		}
+		if rlb.RequestedAt != 0 {
+			run.RequestedAt = time.Unix(rlb.RunScheduledFor, 0).UTC().Format(time.RFC3339)
```
mark-rushakoff: I don't see where`rlb.RequestedAt` was set - it should have been set [from the scheduler's runner](https://github.com/influxdata/platform/blob/6a86f8ee1426653dd81dbc65e0652e015b347d83/task/backend/scheduler.go#L527), with the original value coming from the QueuedRun.

Then `run.RequestedAt` should be assigned the value of `rlb.RequestedAt`, not `rlb.RunScheduledFor`.

task/backend/inmem_logreaderwriter.go
```diff
@@ -34,14 +34,23 @@ func (r *runReaderWriter) UpdateRunState(ctx context.Context, rlb RunLogBase, wh
 		case RunFail, RunSuccess, RunCanceled:
 			r.FinishedAt = whenStr
 		}
+
```
lyondhill: not a necessary new line.

task/backend/inmem_logreaderwriter.go
docmerlin: I think I fixed it, care to look again?

docmerlin: thanks.

PR #908, opened 2018-09-28
Initial dump of code from OSS for storage engine

### LINKED ISSUE #883: Initial import of storage engine into platform repo
This issue tracks the initial import of storage related code from InfluxDB into the Platform repository. The main goals of this issue are to import the minimum amount of code from InfluxDB to create a basic storage engine, which will work at the shard level, and use the Series File, TSI index and the `tsm1` engine.

 - [x] `tsdb` package;
 - [x] `tsi1` package;
 - [x] `tsm1.Engine`;
 - [ ] Remove `tsm1.WAL` package from Engine;
 - [x] `models` package;
 - [ ] Remove all remnants of FGA / authorisation;
 - [x] `'pkg` package;

We should import test packages and fix tests where appropriate.

The aim of this issue is to ensure that the Platform repo is no longer directly importing the `tsdb` and child packages from  InfluxDB.

e-dard:
Resolves #883.

This PR is basically unreviewable.

It's an initial import of code from the OSS project. We didn't try and attempt to bring any git history in simply because it's not a case of pulling in an entire project. Instead we needed certain components that we will be using in the Platform project. Specifically we needed:

  - the `tsm1.Engine`  (the TSM engine for storing data in TSM files and an in-memory cache;
  - the `tsi1.Index` (the index we will be using in Platform;
  - the `tsdb.SeriesFile`
  - the `models` package. Initially we will still be supporting line-protocol, and much of the engine currently deals with these types.
  - Other packages that are imported by the engine, index etc. These are mainly contained in the `pkg` package for now.

In summary, we have only I believe added three top level packages: `models`, `pkg`, and `tsdb`. I certainly realise that the package layout we have added is not necessarily within keeping of how we want to lay code out in Platform. However, we need to get this code in to get a working storage engine. I am happy for the storage team to take ownership of refacting the storage engine package appropriately in the future.


e-dard:
@mark-rushakoff updated.

PR #921, opened 2018-10-01
task Scheduler->tickscheduler

docmerlin:


task/backend/scheduler.go
```diff
@@ -94,11 +93,6 @@ type Scheduler interface {
 	// Stop a scheduler from ticking.
 	Stop()
 
-	// Tick updates the time of the scheduler.
-	// Any owned tasks who are due to execute and who have a free concurrency slot,
-	// will begin a new execution.
```
mark-rushakoff: Since the Tick method is being removed from the interface, can you move the Godoc over to the implementation of Tick?

docmerlin:
accidentally closed.

PR #927, opened 2018-10-02
feat(tsdb/tsm1): Add TSM1 measurement stats.

benbjohnson:

This commit generates an additional `.tss` stats file alongside each TSM file when it is written that contains size stats for all measurements within the TSM file. These files can be combined to generate stats for all measurements across all TSM files.

_What was the problem?_

Previously we could not quickly compute the aggregate size of per-measurement data.

_What was the solution?_

We now calculate per-measurement size for each TSM file when it is compacted and have added a sidecar file to store the data.

  - [x] Rebased/mergeable
  - [x] Tests pass


tsdb/tsm1/reader.go
```diff
@@ -418,8 +436,9 @@ func (t *TSMReader) remove() error {
 	}
 
 	if path != "" {
-		err := os.RemoveAll(path)
-		if err != nil {
+		if err := os.RemoveAll(path); err != nil {
+			return err
+		} else if err := os.RemoveAll(strings.TrimSuffix(path, ".tsm") + ".tss"); err != nil && !os.IsNotExist(err) {
```
jacobmarble: Use `StatsFilename`

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// Sub subtracts the values of all measurements in other from s.
+func (s MeasurementStats) Sub(other MeasurementStats) {
+	for name, v := range other {
+		s[name] -= v
+	}
+}
+
+// ReadFrom reads stats from r in a binary format.
```
jacobmarble: Add to this comment: "Reader must also be an io.ByteReader."

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Read measurement name length.
+	nameLen, err := binary.ReadVarint(br)
+	if err != nil {
+		return n, fmt.Errorf("tsm1.MeasurementStats.readMeasurementFrom: cannot read stats measurement name length: %s", err)
+	}
+	n += int64(binaryutil.VarintSize(nameLen))
+
+	// Read measurement name. Use large maximum so it can usually be stack allocated.
```
jacobmarble: "can usually be heap allocated."

benbjohnson: It's usually stack allocated because the byte slice doesn't escape and it's less than 64KB. I updated the comment to explain further.

go.sum
```diff
@@ -129,6 +129,7 @@ github.com/influxdata/goreleaser v0.86.2-0.20180917235036-c23174192b3a h1:+nfKZu
 github.com/influxdata/goreleaser v0.86.2-0.20180917235036-c23174192b3a/go.mod h1:aVuBpDAT5VtjtUxzvBt8HOd0buzvvk7OX3H2iaviixg=
 github.com/influxdata/influxdb v0.0.0-20180904211643-ab81104697f6 h1:4lXKZjCh1+rd5MIauBsJ4FcwrqZFaErnkVWyrxoJpvQ=
 github.com/influxdata/influxdb v0.0.0-20180904211643-ab81104697f6/go.mod h1:qZna6X/4elxqT3yI9iZYdZrWWdeFOOprn86kgg4+IzY=
+github.com/influxdata/influxdb v1.6.3 h1:TioHM/BpNNH25J89jnL2tk45ww8e2CF+3Q/ih0CMw1I=
```
e-dard: This looks suspicous.

tsdb/tsm1/compact.go
```diff
@@ -1027,6 +1027,7 @@ func (c *Compactor) writeNewFiles(generation, sequence int, src []string, iter K
 
 		// New TSM files are written to a temp file and renamed when fully completed.
 		fileName := filepath.Join(c.Dir, c.formatFileName(generation, sequence)+"."+TSMFileExtension+"."+TmpTSMFileExtension)
+		statsFileName := filepath.Join(c.Dir, c.formatFileName(generation, sequence)+".tss")
```
e-dard: nit, `tss` could have a constant.

tsdb/tsm1/compact.go
```diff
@@ -1027,6 +1027,7 @@ func (c *Compactor) writeNewFiles(generation, sequence int, src []string, iter K
...
 
 		// Write as much as possible to this file
 		err := c.write(fileName, iter, throttle)
```
e-dard: Should `statsFileName` not be passed in here with `fileName`?

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Sub subtracts the values of all measurements in other from s.
+func (s MeasurementStats) Sub(other MeasurementStats) {
+	for name, v := range other {
+		s[name] -= v
+	}
+}
+
+// ReadFrom reads stats from r in a binary format. Reader must also be an io.ByteReader.
+func (s MeasurementStats) ReadFrom(r io.Reader) (n int64, err error) {
```
e-dard: As this is going to be used for important observations I wonder if some CRC checking would be good? We don't want a corrupted file to give us spurious results.

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	// Read measurement name length.
+	nameLen, err := binary.ReadVarint(br)
+	if err != nil {
+		return n, fmt.Errorf("tsm1.MeasurementStats.readMeasurementFrom: cannot read stats measurement name length: %s", err)
+	}
+	n += int64(binaryutil.VarintSize(nameLen))
+
+	// Read measurement name. Use large capacity so it can usually be stack allocated.
```
e-dard: I'm not sure it matters anyway does it? It's going to end up on the heap when it gets put in the map as a string right?

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,205 @@
...
+		return n, err
+	}
+
+	return n, err
+}
+
+// StatsFilename returns the path to the stats file for a given TSM file path.
+func StatsFilename(tsmPath string) string {
+	if strings.HasSuffix(tsmPath, ".tmp") {
+		tsmPath = strings.TrimSuffix(tsmPath, ".tmp")
```
e-dard: Use `tsm1.TmpTSMFileExtension` rather than`tmp`?

tsdb/tsm1/stats.go
```diff
@@ -0,0 +1,205 @@
+package tsm1
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	"github.com/influxdata/influxdb/pkg/binaryutil"
```
e-dard: `s/influxdb/platform`.

tsdb/tsm1/compact.go
benbjohnson: It's derived from `fileName` so it's not necessary.

tsdb/tsm1/stats.go
benbjohnson: Added `CRC32`.

tsdb/tsm1/stats.go
benbjohnson: The `string` would be the intermediate `[]byte` doesn't have to be. I removed the optimization and we can revisit if it becomes a performance issue.

PR #943, opened 2018-10-03
add inmem service for basic auth, http onboarding testing

kelwang:
This PR 
add inmem service for basic auth
add inmem service for onboarding
add common test case in testing package.


kelwang:
just for testing, I guess

PR #953, opened 2018-10-04
Add storage engine

### LINKED ISSUE #884: Create Initial Storage Engine
Platform needs an initial storage engine. The storage engin should use the following existing components:

  - A `tsm1.Engine` without a WAL;
  - A `tsi1.Index`;
  - A `tsdb.SeriesFile`.

The level of granularity to think about for a storage engine is that of a `tsdb.Shard`. Within `platform` for example, we will run a single storage engine. We will need to run multiple storage engines in other products, e.g., the new Enterprise and Cloud platforms.

#### Storage Engine Concerns

 
- A storage engine concerns itself with exposing an API that allows callers to provide series and field data that needs to be indexed and persisted to disk.
- Initially the data will be persisted to a modified `tsm1.Engine`.
- The storage engine will be concerned with writing data to the cache, snapshotting the cache to TSM files/data, and executing compactions. The compaction planner will be pluggable and a custom implementation might be used in other products.
 - Initially the API to write series/field data will expect a `[]models.Point`, however we will extend the API in the future to support alternative data formats that are more efficient for writing to the engine.   
 - The initial entry point to the storage engine could probably be `tsm1.Engine.WritePoints(points []models.Point) error`.
 - We need to continue to support the Delete APIs. Amongst other things, they're using by the retention service.
 - The engine should create a TSI index.
 - The engine should create a Series File.

e-dard:
Closes #884

This PR adds an initial storage engine, which comprises: a TSM engine, a SeriesFile and a TSI index.

_What was the problem?_

We didn't have a storage engine

_What was the solution?_

Write a storage engine.

There is still a huge amount of work I need to do, but this should give us a working storage engine that snapshots on every write. It should be enough for the Chronograf team to use as a backend for their development.

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -51,10 +51,10 @@ func init() {
 		}
 	}
 
-	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, opt tsdb.EngineOptions) tsdb.Index {
-		idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
-		return idx
-	})
+	// tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, ) tsdb.Index {
+	// 	idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
+	// 	return idx
+	// })
 }
 
 // DefaultPartitionN determines how many shards the index will be partitioned into.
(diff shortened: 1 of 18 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -51,10 +51,10 @@ func init() {
...
 	}
 
-	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, opt tsdb.EngineOptions) tsdb.Index {
-		idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
-		return idx
-	})
+	// tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, ) tsdb.Index {
+	// 	idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
+	// 	return idx
+	// })
```
zeebo: remove?

PR #968, opened 2018-10-04
feat(dashboard/views): Add log viewer view type

OfTheDelmer:
Adds `LogViewerView`  that was added in #937  to chronograf's `ViewProperties`
_What was the problem?_
View properties were updated to allow a `log-viewer` type
_What was the solution?_
Add the `LogViewerView` type with the notable divergence from chronograf org config being that  `LogViewerColumn` has a `settings` property name instead of `encodings`.

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
Is this a step towards log viewer cells??

PR #969, opened 2018-10-04
feat(logs/overlay): Add log options overlay

OfTheDelmer:
Adds the log options overlay components from chronograf in preparation for showing log
config for table columns.
_What was the problem?_
Log viewer config needs the options overlay to display the log config columns
_What was the solution?_
Port over options overlay components from chronograf 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

-----

- [x] Updated to use `clockface` imports instead of `reusable_ui`
- [x] Component's with stylesheets are in folders that contain their stylesheet
- [x] Component stylesheets are imported in chrongraf scss
- [x] Updated to use UI kit for all buttons, dropdowns, etc so that no instances of old UI components exist in Platform


chronograf/ui/src/logs/constants/index.ts
```diff
@@ -11,3 +16,97 @@ export const EMPTY_VIEW_PROPERTIES: LogViewerView = {
   type: ViewType.LogViewer,
   shape: ViewShape.ChronografV2,
 }
+
+export const DEFAULT_SEVERITY_LEVELS = {
+  [SeverityLevelOptions.emerg]: SeverityColorOptions.ruby,
```
chnn: (_nitpick_) can we camelcase enums? e.g. `SeverityLevelOptions.Emerg` rather than `SeverityLevelOptions.emerg`

PR #974, opened 2018-10-05
Filter dashboards by owner

### LINKED ISSUE #924: support owner filters for ownable resources
this includes dashboards, buckets, and organizations

jademcgough:
`GET /dashboards` now takes `ids` and `owner` params, so that a request can filter on dashboards that belong to an owner, or a slice of dashboard ids.

closes #924.

### DIFF
--- a/http/dashboard_service.go
+++ b/http/dashboard_service.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path"
+	"strings"
 
 	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/kit/errors"
@@ -161,6 +162,22 @@ func (h *DashboardHandler) handleGetDashboards(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	// if req.filter.OwnerID != nil {
+	// 	filter := platform.UserResourceMappingFilter{
+	// 		UserID:   req.filter.OwnerID,
+	// 		UserType: platform.Owner,
+	// 	}
+	//
+	// 	mappings, _, err := h.UserResourceMappingService.FindUserResourceMappings(ctx, filter)
+	// 	if err != nil {
+	// 		EncodeError(ctx, errors.InternalErrorf("Error loading dashboard owners: %v", err), w)
+	// 		return
+	// 	}
+	//
+	// 	// filter out dashboards by mapping; O(m*n)
+	// 	// can also make a hash to compare to
+	// }
+
 	if err := encodeResponse(ctx, w, http.StatusOK, newGetDashboardsResponse(dashboards)); err != nil {
 		EncodeError(ctx, err, w)
 		return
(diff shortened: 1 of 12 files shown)


http/dashboard_service.go
```diff
@@ -168,19 +187,30 @@ func (h *DashboardHandler) handleGetDashboards(w http.ResponseWriter, r *http.Re
...
 }
 
 func decodeGetDashboardsRequest(ctx context.Context, r *http.Request) (*getDashboardsRequest, error) {
 	qp := r.URL.Query()
 	req := &getDashboardsRequest{}
 
-	if id := qp.Get("id"); id != "" {
-		req.filter.ID = &platform.ID{}
-		if err := req.filter.ID.DecodeFromString(id); err != nil {
+	if idsStr := qp.Get("ids"); idsStr != "" {
```
goller: I prefer multiple id parameters rather than comma delimited ones.  Like this:

`/dashboards?id=1&id=2&id=3` 

To get the list of ids you do:

```go
if ids, ok := qp["id"]; ok {
    for _, id := range ids {
    }
}
```

http/dashboard_service.go
```diff
@@ -662,8 +692,12 @@ func (s *DashboardService) FindDashboards(ctx context.Context, filter platform.D
 	}
 
 	qp := url.Query()
-	if filter.ID != nil {
-		qp.Add("id", filter.ID.String())
+	if len(filter.IDs) > 0 {
```
goller: with the change to not use comma delimited:

```go
for _, id := range filter.IDs{
    qp.Add("id", id)
}
```

http/swagger.yml
```diff
@@ -723,11 +723,17 @@ paths:
...
           - in: query
-            name: org
-            description: specifies the organization of the resource
-            required: true
+            name: owner
+            description: specifies the owner id to return resources for
+            schema:
+              type: string
+          - in: query
+            name: ids
```
goller: This would be:

```yaml
    name: id
    in: query
    description: ID list of dashboards to return. If both this and owner are specified, only ids is used.
    required: false
    schema:
      type: array
      items:
        type: string
    style: form
    explode: true
```

(just looked it up here: https://swagger.io/specification/#parameterStyle)

inmem/dashboard.go
```diff
@@ -38,8 +44,8 @@ func filterDashboardFn(filter platform.DashboardFilter) func(d *platform.Dashboa
 
 // FindDashboards implements platform.DashboardService interface.
 func (s *Service) FindDashboards(ctx context.Context, filter platform.DashboardFilter) ([]*platform.Dashboard, int, error) {
-	if filter.ID != nil {
-		d, err := s.FindDashboardByID(ctx, *filter.ID)
+	if len(filter.IDs) != 1 {
```
goller: should this be `==` ?

mock/user_resource_mapping_service.go
```diff
@@ -0,0 +1,27 @@
...
+func (s *UserResourceMappingService) FindUserResourceMappings(ctx context.Context, filter platform.UserResourceMappingFilter, opt ...platform.FindOptions) ([]*platform.UserResourceMapping, int, error) {
+	return s.FindMappingsF(ctx, filter)
+}
+
+func (s *UserResourceMappingService) CreateUserResourceMapping(ctx context.Context, m *platform.UserResourceMapping) error {
+	return s.CreateMappingF(ctx, m)
+}
+
+func (s *UserResourceMappingService) DeleteUserResourceMapping(ctx context.Context, resourceID platform.ID, userID platform.ID) error {
+	return s.DeleteUserResourceMapping(ctx, resourceID, userID)
```
goller: This should be `DeleteMappingF`

http/dashboard_service.go
```diff
@@ -155,6 +156,24 @@ func (h *DashboardHandler) handleGetDashboards(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	if req.ownerID != nil {
+		filter := platform.UserResourceMappingFilter{
+			UserID:       *req.ownerID,
+			UserType:     platform.Owner,
```
goller: I really like how these mapping filters look

user_resource_mapping.go
```diff
@@ -42,12 +46,16 @@ func (m UserResourceMapping) Validate() error {
 	if m.UserType != Owner && m.UserType != Member {
 		return errors.New("A valid user type is required")
 	}
+	if m.ResourceType != DashboardResourceType && m.ResourceType != BucketResourceType {
```
goller: Mostly because I know we'll be adding more of these resources types, I usually write this sorta thing like:

```go
switch m.ResourceType{
case DashboardResourceType, BucketResourceType:
default:
    return fmt.Errorf("A valid resource type is required")
}
```

that way whenever the next resources are added, it's as easy as just appending to that list.

inmem/dashboard.go
jademcgough: lol, yes

user_resource_mapping.go
jademcgough: +1, was thinking something along those lines

e-dard: ^ Error messages shouldn't be capitalised.

http/swagger.yml
jademcgough: From https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md, `style: form` and `explode: true` are both defaults for `in: query`.

jademcgough:
@goller this is ready for a second review. I have a second PR coming up with etcd changes.

PR #975, opened 2018-10-05
Now you can cancel runs

docmerlin:
This should allow you to cancel runs.
It makes breaking changes to tasks, so will need some work in IDPE as well.

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -74,6 +74,7 @@ func NewTaskHandler(mappingService platform.UserResourceMappingService, logger *
 	h.HandlerFunc("GET", tasksIDRunsPath, h.handleGetRuns)
 	h.HandlerFunc("GET", tasksIDRunsIDPath, h.handleGetRun)
 	h.HandlerFunc("POST", tasksIDRunsIDRetryPath, h.handleRetryRun)
+	h.HandlerFunc("DELETE", tasksIDRunsIDPath, h.handleCancelRun)
 
 	return h
 }
@@ -585,6 +586,53 @@ func decodeGetRunRequest(ctx context.Context, r *http.Request, orgs platform.Org
 	}, nil
 }
 
+type cancelRunRequest struct {
+	RunID  platform.ID
+	TaskID platform.ID
+}
+
+func decodeCancelRunRequest(ctx context.Context, r *http.Request) (*cancelRunRequest, error) {
+	params := httprouter.ParamsFromContext(ctx)
+	rid := params.ByName("rid")
+	if rid == "" {
+		return nil, kerrors.InvalidDataf("you must provide a run ID")
+	}
+	tid := params.ByName("tid")
+	if tid == "" {
+		return nil, kerrors.InvalidDataf("you must provide a task ID")
+	}
+
+	var i platform.ID
+	if err := i.DecodeFromString(rid); err != nil {
+		return nil, err
+	}
+	var t platform.ID
+	if err := t.DecodeFromString(tid); err != nil {
+		return nil, err
+	}
+
+	return &cancelRunRequest{
+		RunID:  i,
+		TaskID: t,
+	}, nil
+}
+
+func (h *TaskHandler) handleCancelRun(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	req, err := decodeCancelRunRequest(ctx, r)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	err = h.TaskService.CancelRun(ctx, req.TaskID, req.RunID)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+}
+
 func (h *TaskHandler) handleRetryRun(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
@@ -879,6 +927,38 @@ func (t TaskService) FindRunByID(ctx context.Context, orgID, runID platform.ID) 
 
 func (t TaskService) RetryRun(ctx context.Context, id platform.ID) (*platform.Run, error) {
 	return nil, errors.New("not yet implemented")
+}
+
+func cancelPath(taskID, runID platform.ID) string {
+	return path.Join(taskID.String(), runID.String())
+}
+
+func (t TaskService) CancelRun(ctx context.Context, taskID, runID platform.ID) error {
+	u, err := newURL(t.Addr, cancelPath(taskID, runID))
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest("DELETE", u.String(), nil)
+	if err != nil {
+		return err
+	}
+
+	SetToken(t.Token, req)
+
+	hc := newClient(u.Scheme, t.InsecureSkipVerify)
+
+	resp, err := hc.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if err := CheckError(resp); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func taskIDPath(id platform.ID) string {
--- a/task.go
+++ b/task.go
@@ -56,7 +56,10 @@ type TaskService interface {
 	// Returns a single run
 	FindRunByID(ctx context.Context, orgID, runID ID) (*Run, error)
 
-	// Creates and returns a new run (which is a retry of another run)
+	// CancelRun cancels a currently running run.
+	CancelRun(ctx context.Context, taskID, runID ID) error
+
+	// Creates and returns a new run (which is a retry of another run).
 	RetryRun(ctx context.Context, id ID) (*Run, error)
 }
 
--- a/task/backend/scheduler.go
+++ b/task/backend/scheduler.go
@@ -15,6 +15,7 @@ import (
 )
 
 var (
+	// ErrRunCanceled is returned from the RunResult when a Run is Canceled.  It is used mostly internally.
 	ErrRunCanceled = errors.New("run canceled")
 
 	// ErrTaskNotClaimed is returned when attempting to operate against a task that must be claimed but is not.
@@ -110,6 +111,9 @@ type Scheduler interface {
 	// ReleaseTask immediately cancels any in-progress runs for the given task ID,
 	// and releases any resources related to management of that task.
 	ReleaseTask(taskID platform.ID) error
+
+	// Cancel stops an executing run.
+	CancelRun(ctx context.Context, taskID, runID platform.ID) error
 }
 
 // TickSchedulerOption is a option you can use to modify the schedulers behavior.
@@ -148,7 +152,7 @@ func NewScheduler(desiredState DesiredState, executor Executor, lw LogWriter, no
 		executor:       executor,
 		logWriter:      lw,
 		now:            now,
-		taskSchedulers: make(map[string]*taskScheduler),
+		taskSchedulers: make(map[platform.ID]*taskScheduler),
 		logger:         zap.NewNop(),
 		wg:             &sync.WaitGroup{},
 		metrics:        newSchedulerMetrics(),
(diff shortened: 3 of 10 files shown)


task/backend/scheduler.go
```diff
@@ -289,6 +305,11 @@ func (s *TickScheduler) PrometheusCollectors() []prometheus.Collector {
 	return s.metrics.PrometheusCollectors()
 }
 
+type runCtx struct {
```
lyondhill: I kinda like this. combining the ctx and its cancel function. It could actually be made a bit more generic. something like ctxWithCancel. Either way it looks good

task.go
```diff
@@ -53,10 +53,13 @@ type TaskService interface {
 	// Returns a list of runs that match a filter and the total count of returned runs.
 	FindRuns(ctx context.Context, filter RunFilter) ([]*Run, int, error)
 
-	// Returns a single run
+	// Returns a single run.
 	FindRunByID(ctx context.Context, id ID) (*Run, error)
 
-	// Creates and returns a new run (which is a retry of another run)
+	// CancelRun cancles a currently running run.
```
mark-rushakoff: `cancels` an in-progress run

task/backend/scheduler.go
```diff
@@ -500,29 +528,40 @@ func (r *runner) startFromWorking(now int64) {
...
 	}
-
-	rc, err := r.desiredState.CreateNextRun(r.ctx, r.task.ID, now)
+	r.ts.runningMu.Lock()
+	ctx, cancel := context.WithCancel(r.ctx)
+	rc, err := r.desiredState.CreateNextRun(ctx, r.task.ID, now)
 	if err != nil {
 		r.logger.Info("Failed to create run", zap.Error(err))
 		atomic.StoreUint32(r.state, runnerIdle)
+		cancel() // cancel to prevent context leak
```
mark-rushakoff: runningMu needs to be unlocked for this early return.

task/servicetest/servicetest.go
```diff
@@ -35,14 +35,23 @@ func init() {
 // If creating the System fails, the implementer should call t.Fatal.
 type BackendComponentFactory func(t *testing.T) (*System, context.CancelFunc)
 
+type naieveStoreRunController struct {
```
mark-rushakoff: The correct spelling is naive or naïve.

task/backend/scheduler.go
mark-rushakoff: I'm fine with including both of these in one struct, but I disagree about embedding them. In particular, accessing context.CancelFunc (which is defined to be just `func()`) is awkward. Do I call `runCtx()`? Or do I have to always do `runCtx.CancelFunc()`?

http/task_service.go
```diff
@@ -453,6 +453,54 @@ func decodeGetRunRequest(ctx context.Context, r *http.Request) (*getRunRequest,
...
+	var t platform.ID
+	if err := t.DecodeFromString(tid); err != nil {
+		return nil, err
+	}
+
+	return &cancelRunRequest{
+		RunID:  i,
+		TaskID: t,
+	}, nil
+
```
mark-rushakoff: Remove blank line.

mark-rushakoff:
Sorry, we didn't get to updating this branch this past week. There's going to be some merge conflicts.

http/task_service.go
```diff
@@ -469,16 +470,13 @@ func (h *TaskHandler) handleGetRun(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
 	tok, err := GetToken(r)
-	if err != nil {
```
mark-rushakoff: Seems like a mistaken delete, as this error is now unchecked.

task/backend/scheduler.go
```diff
@@ -172,6 +175,19 @@ type TickScheduler struct {
 	taskSchedulers map[string]*taskScheduler // Stringified task ID -> task scheduler.
 }
 
+func (s *TickScheduler) CancelRun(taskID, runID platform.ID) error {
+	s.schedulerMu.Lock()
+	defer s.schedulerMu.Unlock()
+	ts := s.taskSchedulers[string(taskID)]
```
mark-rushakoff: This needs to be `taskID.String()`. `string(<uint value>)` does something different.

task/backend/scheduler.go
```diff
@@ -511,12 +535,18 @@ func (r *runner) RestartRun(qr QueuedRun) bool {
...
-
 	r.wg.Add(1)
-	go r.executeAndWait(qr, runLogger)
+	r.ts.runningMu.Lock()
+	rCtx, ok := r.ts.running[qr.RunID.String()]
+	if !ok {
+		ctx, cancel := context.WithCancel(context.TODO())
+		rCtx = runCtx{Context: ctx, CancelFunc: cancel}
+		r.ts.running[qr.RunID.String()] = rCtx
+		r.ts.runningMu.Unlock()
```
mark-rushakoff: Move runningMu.Unlock outside the if?

task/servicetest/servicetest.go
mark-rushakoff: Still misspelled.

task/backend/scheduler.go
```diff
@@ -172,6 +175,19 @@ type TickScheduler struct {
 	taskSchedulers map[string]*taskScheduler // Stringified task ID -> task scheduler.
 }
 
+func (s *TickScheduler) CancelRun(taskID, runID platform.ID) error {
+	s.schedulerMu.Lock()
+	defer s.schedulerMu.Unlock()
+	ts := s.taskSchedulers[string(taskID)]
+	ts.runningMu.Lock()
+	cancel := ts.running[string(runID)].CancelFunc
```
mark-rushakoff: `runID.String()`.

task/mock/scheduler.go
```diff
@@ -287,14 +292,18 @@ func NewExecutor() *Executor {
...
 	rp := NewRunPromise(run)
 
 	id := run.TaskID.String() + run.RunID.String()
 	e.mu.Lock()
 	e.running[id] = rp
 	e.mu.Unlock()
 	go func() {
+		select {
+		case <-ctx.Done():
+		case <-time.After(e.hangingFor):
```
mark-rushakoff: Is a zero value here okay? I guess that matches previous behavior.

task/mock/scheduler.go
```diff
@@ -306,6 +315,10 @@ func (e *Executor) Execute(_ context.Context, run backend.QueuedRun) (backend.Ru
 
 func (e *Executor) WithLogger(l *zap.Logger) {}
 
+func (e *Executor) WithHanging(dt time.Duration) {
```
mark-rushakoff: I don't see where this new function is used.

task/mock/scheduler.go
docmerlin: yah, the zero value will just go through

task/backend/scheduler.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 var (
+	// ErrRunCanceled is returned when a Run is Canceled.
```
mark-rushakoff: This comment should be more specific. I don't think calling CancelRun returns this error, but that's what the comment reads like.

task/backend/scheduler.go
```diff
@@ -340,7 +369,9 @@ type taskScheduler struct {
 	wg     *sync.WaitGroup
 
 	// Fixed-length slice of runners.
-	runners []*runner
+	runners   []*runner
+	running   map[string]runCtx
```
mark-rushakoff: I'm on board with the running map, but I can't tell from this layout whether runningMu protects the map, the slice, or both.

The map keys probably should be `platform.ID` now that that's a uint64 and no longer a byte slice.

task/backend/scheduler.go
```diff
@@ -519,12 +551,18 @@ func (r *runner) RestartRun(qr QueuedRun) bool {
...
-
 	// create a QueuedRun because we cant stm.CreateNextRun
 	runLogger := r.logger.With(zap.String("run_id", qr.RunID.String()), zap.Int64("now", qr.Now))
-
 	r.wg.Add(1)
-	go r.executeAndWait(qr, runLogger)
+	r.ts.runningMu.Lock()
+	rCtx, ok := r.ts.running[qr.RunID.String()]
+	if !ok {
+		ctx, cancel := context.WithCancel(context.TODO())
```
mark-rushakoff: Maybe this context.TODO could be the parent TickScheduler's `ctx` field. It looks like you'd only get to the `taskScheduler` after calling TickScheduler.Start, but I'm not certain.

task/backend/scheduler.go
```diff
@@ -538,31 +576,43 @@ func (r *runner) startFromWorking(now int64) {
...
-	go r.executeAndWait(qr, runLogger)
+	go r.executeAndWait(ctx, qr, runLogger)
 
 	r.updateRunState(qr, RunStarted, runLogger)
 }
 
-func (r *runner) executeAndWait(qr QueuedRun, runLogger *zap.Logger) {
+func (r *runner) clearRunning(id platform.ID) {
+	r.ts.runningMu.Lock()
+	delete(r.ts.running, id.String())
```
mark-rushakoff: Should this call `Cancel` from the object in the map as a matter of resource cleanup?

task/backend/scheduler.go
```diff
@@ -179,6 +183,26 @@ type TickScheduler struct {
...
+		return ErrTaskNotFound
+	}
+	ts.runningMu.Lock()
+	c, ok := ts.running[runID.String()]
+	if !ok {
+		ts.runningMu.Unlock()
+		return ErrRunNotFound
+	}
+	ts.runningMu.Unlock()
+	if c.CancelFunc != nil {
```
mark-rushakoff: Not sure about this nil check. Are there any circumstances where CancelFunc is allowed to be nil?

task/platform_adapter.go
```diff
@@ -10,13 +10,22 @@ import (
...
+	RunController
+}
+type RunController interface {
+	CancelRun(ctx context.Context, taskID, runID platform.ID) error
+	//TODO: add retry run to this.
+}
+
 // PlatformAdapter wraps a task.Store into the platform.TaskService interface.
-func PlatformAdapter(s backend.Store, r backend.LogReader) platform.TaskService {
+func PlatformAdapter(s StoreRunController, r backend.LogReader) platform.TaskService {
```
mark-rushakoff: I think a few things will end up cleaner if the argument signature is instead `(backend.Store, RunController, backend.LogReader)`, and then the `RunController` can just be a new field on `pAdapter`.

- StoreRunController interface can be removed
- servicetest.System can refer to task.RunController and stop referring to backend.Scheduler
- servicetest can drop the NaiveStoreRunController type
- Possible simplifications in the enterprise changes, haven't gotten there yet

http/task_service.go
```diff
@@ -585,6 +586,53 @@ func decodeGetRunRequest(ctx context.Context, r *http.Request, orgs platform.Org
...
+}
+
+func decodeCancelRunRequest(ctx context.Context, r *http.Request) (*cancelRunRequest, error) {
+	params := httprouter.ParamsFromContext(ctx)
+	rid := params.ByName("rid")
+	if rid == "" {
+		return nil, kerrors.InvalidDataf("you must provide a run ID")
+	}
+	tid := params.ByName("tid")
+	if rid == "" {
```
mark-rushakoff: Should be checking `tid`.

task/backend/scheduler.go
docmerlin: That would cancel task running when a runner was canceled.  I don't think we want that behavior.  If a node is shut down, it shouldn't cancel the runs.

task/backend/scheduler.go
docmerlin: If it was just cleared, but something else is still holding the context in memory

task/platform_adapter.go
docmerlin: I went with (backend.Store, backend.LogReader, RunController)
I agree.  This also was helped by switching scheduler's CancelRun to take a context

PR #976, opened 2018-10-05
Add the tsm1.WAL to the storage engine

### LINKED ISSUE #954: Add the WAL
It's now been decided that writes need to be readable as soon as they're written. As such, we will add the WAL back to the storage engine. The WAL will default to off, and in the case of Platform, will be enabled for durability.

 - [ ] Add the WAL back.
 - [ ] Remove NATS.

e-dard:
Closes #954 

This PR adds the WAL to the storage engine, and removes the usage of NATS in the write path.

_What was the problem?_

The existing write path used NATS, which meant that writes were not immediately readable. It was decided from a product perspective that writes in Platform should be readable immediately (as they are in the 1.x product).

_What was the solution?_

To add the WAL back and not use an async log. 

Now, the handler will write directly into the storage engine. Writes will immediately be available to query from the cache, and will be made durable via the WAL. Back pressure will need to be implemented in a different way to prevent overloading of the cache.


storage/engine_test.go
```diff
@@ -0,0 +1,227 @@
...
+// MustOpen opens the engine or panicks.
+func (e *Engine) MustOpen() {
+	if err := e.Engine.Open(); err != nil {
+		panic(err)
+	}
+}
+
+// WritePoints converts old style points into the new 2.0 engine format.
+func (e *Engine) WritePoints(pts []models.Point) error {
+	points, err := tsdb.ExplodePoints([]byte("11111111"), []byte("22222222"), pts)
```
zeebo: fyi: it was very unclear that this was happening as a reviewer of this test lol.

tsdb/index/tsi1/index.go
```diff
@@ -638,9 +638,13 @@ func (i *Index) CreateSeriesListIfNotExists(collection *tsdb.SeriesCollection) e
...
 	for k := 0; k < n; k++ {
 		go func() {
+			i.mu.RLock()
+			partitionN := len(i.partitions)
+			i.mu.RUnlock()
+
 			for {
 				idx := int(atomic.AddUint32(&pidx, 1) - 1) // Get next partition to work on.
-				if idx >= len(i.partitions) {
+				if idx >= partitionN {
```
zeebo: since we index into i.partitions right under this, is it worthwhile to grab the value under the read lock?

tsdb/shard.go
```diff
@@ -186,13 +186,14 @@ func (s *Shard) close() error {
...
-	}
-	return err
+	return nil // TODO(edd)remove
+	// var err error
+	// if s._engine == nil {
+	// 	err = ErrEngineClosed
+	// } else if !s.enabled {
+	// 	err = ErrShardDisabled
+	// }
+	// return err
```
zeebo: was this a leftover todo or a future todo?

PR #989, opened 2018-10-06
Revise onboarding flow

ebb-tide:
This PR contains changes to component styles to improve presentation of onboarding flow. 
It also moves setup and signin state to the containers which control whether to display them rather than redux state.


chronograf/ui/src/Signin.tsx
```diff
@@ -31,27 +31,28 @@ export class Signin extends PureComponent<Props, State> {
 
     this.state = {
       loading: RemoteDataState.NotStarted,
-      isSourcesAllowed: false,
+      isUserSignedIn: false,
```
bthesorceror: Just from my context, isUserSignedIn feels like something that should be in redux.

chronograf/ui/src/clockface/components/wizard/ProgressBar.tsx
```diff
@@ -26,66 +26,69 @@ class ProgressBar extends PureComponent<Props, null> {
...
     const {stepStatuses, stepTitles, currentStepIndex} = this.props
 
     const lastIndex = stepStatuses.length - 1
     const lastEleIndex = stepStatuses.length - 2
 
-    const progressBar = stepStatuses.reduce((acc, stepStatus, i) => {
-      if (i === 0 || i === lastIndex) {
-        return [...acc]
-      }
+    const progressBar: JSX.Element[] = stepStatuses.reduce(
```
bthesorceror: Could progressBar be its own element that took stepStatuses as a prop?

ebb-tide:
Will address review comments and merge later today

chronograf/ui/src/Signin.tsx
ebb-tide: I decided that it was better to keep in state, since the consumer of the state is direct parent of this component.

PR #993, opened 2018-10-08
Added a find all buckets implementation according to updated interface in flux

aanthony1243:
Would be ideal if this implementation could be used in place of custom lookup implementations.


aanthony1243:
Will open a new PR that discusses new code.

PR #999, opened 2018-10-08
feat(v2/views): Add read views api util

OfTheDelmer:
Update dashboard views api to read views and accept params for type filters
_What was the problem?_
There was no util to fetch all views with a filter on type
_What was the solution?_
Add an api util to read all views and supply an optional type filter

  - [x] Rebased/mergeable
  - [x] Tests pass

Closes https://github.com/influxdata/applications-team-issues/issues/120


chronograf/ui/src/dashboards/apis/v2/view.ts
```diff
@@ -15,3 +18,21 @@ export const getView = async (url: string): Promise<View> => {
...
+    const response = await AJAX({
+      method: 'GET',
+      url,
+      params,
+    })
+
+    return getDeep<View[]>(response, 'data.views', [])
+  } catch (error) {
+    console.error(error)
+    throw error
```
bthesorceror: What do you think about logging the error and returning an empty array? It would make the error visible without potential breaking the app in a weird place.

PR #1006, opened 2018-10-08
add plugins

kelwang:
Closes #https://github.com/influxdata/platform/issues/1007

  - [x] Rebased/mergeable

### DIFF
--- a/telegraf/plugins/inputs/inputs_test.go
+++ b/telegraf/plugins/inputs/inputs_test.go
@@ -1,0 +1,215 @@
+package inputs
+
+import "testing"
+
+// local plugin
+type telegrafPluginConfig interface {
+	TOML() string
+}
+
+func TestTOML(t *testing.T) {
+	cases := []struct {
+		name    string
+		plugins map[telegrafPluginConfig]string
+	}{
+		{
+			name: "test empty plugins",
+			plugins: map[telegrafPluginConfig]string{
+				&CPUStats{}:  "[[inputs.cpu]]\n",
+				&DiskStats{}: "[[inputs.disk]]\n",
+				&DiskIO{}:    "[[inputs.diskio]]\n",
+				&File{}: `[[inputs.file]]	
+  ## Files to parse each interval.
+  ## These accept standard unix glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   /var/log/**.log     -> recursively find all .log files in /var/log
+  ##   /var/log/*/*.log    -> find all .log files with a parent dir in /var/log
+  ##   /var/log/apache.log -> only read the apache log file
+  files = []
+`,
+				&Kernel{}: "[[inputs.kernel]]\n",
+				&Kubernetes{}: `[[inputs.kubernetes]]
+  ## URL for the kubelet
+  ## exp: http://1.1.1.1:10255
+  url = ""	
+`,
+				&LogParserPlugin{}: `[[inputs.logparser]]	
+  ## Log files to parse.
+  ## These accept standard unix glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   /var/log/**.log     -> recursively find all .log files in /var/log
+  ##   /var/log/*/*.log    -> find all .log files with a parent dir in /var/log
+  ##   /var/log/apache.log -> only tail the apache log file
+  files = []
+`,
+				&MemStats{}:    "[[inputs.mem]]\n",
+				&NetIOStats{}:  "[[inputs.net]]\n",
+				&NetResponse{}: "[[inputs.net_response]]\n",
+				&Nginx{}: `[[inputs.nginx]]
+  # An array of Nginx stub_status URI to gather stats.
+  # exp http://localhost/server_status
+  urls = []
+`,
+				&Processes{}: "[[inputs.processes]]\n",
+				&Procstat{}: `[[inputs.procstat]]
+  ## executable name (ie, pgrep <exe>)
+  exe = ""
+`,
+				&Prometheus{}: `[[inputs.prometheus]]	
+  ## An array of urls to scrape metrics from.
+  urls = []
+`,
+				&Redis{}: `[[inputs.redis]]
+  ## specify servers via a url matching:
+  ##  [protocol://][:password]@address[:port]
+  ##  e.g.
+  ##    tcp://localhost:6379
+  ##    tcp://:password@192.168.99.100
+  ##    unix:///var/run/redis.sock
+  ##
+  ## If no servers are specified, then localhost is used as the host.
+  ## If no port is specified, 6379 is used
+  servers = []
+
+  ## specify server password
+  # password = ""
+`,
+				&SwapStats{}: "[[inputs.swap]]\n",
+				&Syslog{}: `[[inputs.syslog]]
+  ## Specify an ip or hostname with port - eg., tcp://localhost:6514, tcp://10.0.0.1:6514
+  ## Protocol, address and port to host the syslog receiver.
+  ## If no host is specified, then localhost is used.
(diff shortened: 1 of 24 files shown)


telegraf/plugins/input/ngnix.go
```diff
@@ -0,0 +1,19 @@
...
+type Nginx struct {
+	URLs []string `json:"urls,omitempty"`
+}
+
+// TOML encodes to toml string
+func (n *Nginx) TOML() string {
+	return fmt.Sprintf(`[[inputs.nginx]]	
+  # An array of Nginx stub_status URI to gather stats.
+  urls = ["%s"]
+`, strings.Join(n.URLs, ", "))
```
goller: these strings should be quoted.

telegraf/plugins/input/prometheus.go
```diff
@@ -0,0 +1,19 @@
...
+type Prometheus struct {
+	URLs []string `json:"urls,omitempty"`
+}
+
+// TOML encodes to toml string
+func (p *Prometheus) TOML() string {
+	return fmt.Sprintf(`[[inputs.prometheus]]	
+  ## An array of urls to scrape metrics from.
+  urls = ["%s"]
+`, strings.Join(p.URLs, ", "))
```
goller: these strings should be quoted

telegraf/plugins/input/redis.go
```diff
@@ -0,0 +1,31 @@
...
+##    tcp://:password@192.168.99.100
+##    unix:///var/run/redis.sock
+##
+## If no servers are specified, then localhost is used as the host.
+## If no port is specified, 6379 is used
+  servers = ["%s"]
+
+## specify server password
+# password = "%s"
+`, strings.Join(r.Servers, ", "), r.Password)
```
goller: this should be quoted

telegraf/plugins/input/syslog.go
```diff
@@ -0,0 +1,10 @@
+package input
+
+// Syslog is based on telegraf Syslog plugin.
+type Syslog struct{}
+
+// TOML encodes to toml string
+func (s *Syslog) TOML() string {
+	return `[[inputs.syslog]]
```
goller: syslog needs `server = "tcp://:6514"`

https://github.com/influxdata/telegraf/blob/master/plugins/inputs/syslog/syslog.go#L57

telegraf/plugins/input/tail.go
```diff
@@ -0,0 +1,27 @@
...
+  ## These accept standard unix glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   "/var/log/**.log"  -> recursively find all .log files in /var/log
+  ##   "/var/log/*/*.log" -> find all .log files with a parent dir in /var/log
+  ##   "/var/log/apache.log" -> just tail the apache log file
+  ##
+  ## See https://github.com/gobwas/glob for more examples
+  ##
+  files = ["%s"]
+`, strings.Join(t.Files, ", "))
```
goller: all files need quoting

telegraf/plugins/output/file.go
```diff
@@ -0,0 +1,33 @@
...
+		if v.Typ == "stdout" {
+			s[k] = v.Typ
+			continue
+		}
+		s[k] = v.Path
+	}
+	return fmt.Sprintf(`[[outputs.file]]
+  ## Files to write to, "stdout" is a specially handled file.
+  files = [%s]
+`, strings.Join(s, ", "))
```
goller: this needs to be quoted

telegraf/plugins/output/influxdb_v2.go
```diff
@@ -0,0 +1,36 @@
...
+type InfluxDBV2 struct {
+	URLs         []string `toml:"urls"`
+	Token        string   `toml:"token"`
+	Organization string   `toml:"organization"`
+	Bucket       string   `toml:"bucket"`
+}
+
+// TOML encodes to toml string.
+func (i *InfluxDBV2) TOML() string {
+	s := make([]string, len(i.URLs))
```
goller: Use https://golang.org/pkg/strconv/#Quote instead

telegraf_config.go
```diff
@@ -0,0 +1,215 @@
+package platform
```
goller: probably just named it telegraf.go

telegraf_config_test.go
```diff
@@ -0,0 +1,104 @@
+package platform
```
goller: This would go into PR #3

telegraf_config.go
goller: Available plugin code should be moved into the telegraf package PR #2.  It will be used for validation.
Also, move all the JSON tags, marshaling/unmarshaling into PR #3

telegraf/plugins/input/file.go
```diff
@@ -0,0 +1,29 @@
...
+	"strings"
+)
+
+// File is based on telegraf input File plugin.
+type File struct {
+	Files []string `json:"files"`
+}
+
+// TOML encodes to toml string
+func (f *File) TOML() string {
```
goller: I think that if len(files) ==0 it should return empty string.
Check for all plugins that have configuration options that are empty would return an empty string.

telegraf/plugins/input/cpu.go
```diff
@@ -0,0 +1,10 @@
+package input
```
goller: I think each plugin needs to record what version of telegraf it was added.

telegraf/plugins/output/file.go
```diff
@@ -0,0 +1,34 @@
+package output
```
goller: should be `outputs` rather than `output`

telegraf/plugins/input/tail.go
```diff
@@ -0,0 +1,32 @@
+package input
```
goller: should be package `inputs` rather than input

telegraf/plugins/input/file.go
```diff
@@ -0,0 +1,29 @@
...
+	s := make([]string, len(f.Files))
+	for k, v := range s {
+		s[k] = strconv.Quote(v)
```
goller: Add tests for things like quotes in the filename... need to check they are escaped.

telegraf/plugins/input/nvidia_smi.go
```diff
@@ -0,0 +1,10 @@
+package input
```
goller: Let's remove this for now.

telegraf/plugins/input/file.go
kelwang: If we print an empty string, telegraf will ignore the plugin. Shall we let telegraf report a config error instead?

telegraf/plugins/input/cpu.go
kelwang: Hey, @danielnelson, which version do you think we should start with, 1.8.0?

danielnelson: Yes, this is the first version with InfluxDB 2.0 support.

danielnelson: Oh, except only 1.9.0 will have support for direct downloading, so that might factor in.

telegraf/plugins/inputs/inputs_test.go
```diff
@@ -0,0 +1,215 @@
...
+  ##    tcp://localhost:6379
+  ##    tcp://:password@192.168.99.100
+  ##    unix:///var/run/redis.sock
+  ##
+  ## If no servers are specified, then localhost is used as the host.
+  ## If no port is specified, 6379 is used
+  servers = ["tcp://localhost:6379", "unix:///var/run/redis.sock"]
+
+  ## specify server password
+  # password = "somepassword123"
```
goller: Looks like this cannot be a simple replace.

PR #1017, opened 2018-10-09
Add owner/member endpoints for views

jademcgough:


http/swagger.yml
```diff
@@ -2515,6 +2687,178 @@ paths:
...
+        - Users
+        - Views
+      summary: List all view members
+      parameters:
+        - in: path
+          name: viewID
+          schema:
+            type: string
+          required: true
+          description: ID of the dashboard
```
goller: type: ID of the view

PR #1023, opened 2018-10-09
test(inmem): re-enable skipped CreateBucket test

mark-rushakoff:
I don't know why it was skipped originally, but I ran it under the race
detector for tens of thousands of iterations and couldn't get it to
fail.


goller:
@mark-rushakoff looks like a test that should have had it's skip removed when the fix to the inmem sync map was updated.

PR #1028, opened 2018-10-10
feat: checktidy will check the go.mod and go.sum files in circleci

jsternberg:


Makefile
```diff
@@ -107,7 +107,12 @@ chronograf/ui/build:
 fmt: $(SOURCES_NO_VENDOR)
 	gofmt -w -s $^
 
-checkfmt: $(SOURCES_NO_VENDOR)
+checkfmt:
+	GO111MODULE=on go mod tidy
```
goller: Rather than add shell scripts to the makefile, make a shell script in etc and a new make target, like `tidy`.

CI should call tidy.

jsternberg: I'll add two new targets: `tidy` and `checktidy`.

jsternberg:
I've updated this PR.

PR #1031, opened 2018-10-10
 Replace Flux Builder with Data Explorer

chnn:
- Implements a Redux reducer for identical state between VEO and DE
- Replaces the “Flux Builder” page with the Data Explorer


chronograf/ui/src/dataExplorer/components/DataExplorer.tsx
```diff
@@ -0,0 +1,49 @@
...
+// Components
+import TimeMachine from 'src/shared/components/TimeMachine'
+
+// Actions
+import {setActiveTimeMachineID} from 'src/shared/actions/v2/timeMachines'
+
+// Utils
+import {DE_TIME_MACHINE_ID} from 'src/shared/constants/timeMachine'
+
+interface StateProps {}
```
AlirieGray: what is this for?

chnn: We started using this pattern the other day to get some better type safety for Redux-connected components. The idea is the specify which props are connected from state, which are dispatch props, and which are passed in.

There's an example of how we use these props [here](https://github.com/influxdata/platform/blob/4a9f0959b8c167633b264e1bd154d23b945aed9c/chronograf/ui/src/dashboards/components/VEO.tsx#L89-L99).

PR #1042, opened 2018-10-10
Replace outdated TimeRangeDropdown

chnn:
Replaces the `TimeRangeDropdown` component with a simple clockface based implementation. This is temporary until we decide how we want to modernize the time range selection.


chronograf/ui/src/shared/components/TimeRangeDropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!selectedTimeRange) {
+      throw new Error('TimeRangeDropdown passed unknown TimeRange')
+    }
+
+    return (
+      <Dropdown
+        selectedID={selectedTimeRange.label}
+        onChange={this.handleChange}
+      widthPixels={100}
```
AlirieGray: spacing?

chronograf/ui/src/shared/constants/timeRanges.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+  {
+    seconds: 900,
+    lower: 'now() - 15m',
+    upper: null,
+    label: 'Past 15m',
+  },
+  {
+    seconds: 3600,
+    lower: 'now() - 1h',
```
AlirieGray: why lower - time instead of using lower and upper and subtracting them?

chronograf/ui/src/shared/components/TimeRangeDropdown.tsx
chnn: thanks

chronograf/ui/src/shared/constants/timeRanges.ts
chnn: purely for historic reasons... these `TimeRange` interfaces are used all over

PR #1059, opened 2018-10-11
docs(http): tasks fields can now be updated

goller:
Tasks should be able to update fields that correspond to flux options.  This would imply that the tasks would rewrite the flux AST and format it back into flux.

_What was the problem?_
It is nicer from a UI point of view to be able to update certain fields and have the corresponding flux updated.

_What was the solution?_


http/swagger.yml
```diff
@@ -3281,7 +3281,6 @@ components:
           description: The ID of the organization that owns this Task.
           type: string
         name:
-          readOnly: true
           description: A read-only description of the task.
```
mark-rushakoff: I guess this isn't going to be read-only anymore.

PR #1063, opened 2018-10-11
Logs/transform log response util function

AlirieGray:
Closes #https://github.com/influxdata/applications-team-issues/issues/117

_What was the problem?_
The response from the server for a logs query returns data with unnecessary properties, such as empty string. 

_What was the solution?_
Added a util function,`transformFluxLogsResponse` to turn the server response into `TableData` shape so that it can be used by the Logs Viewer.

  - [x] Rebased/mergeable
  - [x] Tests pass


chronograf/ui/src/logs/utils/index.ts
```diff
@@ -1,7 +1,37 @@
...
+): TableData => {
+  const {tables} = response
+
+  const values: TimeSeriesValue[][] = []
+  const rows = getDeep<TimeSeriesValue[][]>(tables, '0.data', [])
+  const columnNamesRow = getDeep<string[]>(tables, '0.data.0', [])
+  const columns: string[] = columnNamesRow.slice(3)
+
+  for (let i = 1; i < rows.length; i++) {
+    values.push(rows[i].slice(3))
```
chnn: This might be a bit brittle, if the order of columns in the Flux response ever changes.

Would it be possible to have a whitelist of column names that we want to extract? And then extract those by first looking up what index they are at, and then extracting only those indices from each row. Or alternatively, if we don't know what columns to expect in advance, perhaps we could a blacklist of columns that we for sure do not want to consume (`'result'`, `''`, `'table'`, etc.)

I'm OK with merging this, but it's something we might want to consider for the future (cc log master @OfTheDelmer)

AlirieGray: Good point--I like the idea of extracting column names

PR #1065, opened 2018-10-11
chore(bolt): log path of bolt file

mark-rushakoff:
Now at startup, there will be a log line like:

    Opened bolt database {"log_id": "0B5Oks9W000", "path": "influxd.bolt"}

Which gives a somewhat better idea of what's going on.


mark-rushakoff:
That is where I expected it in the first place and part of the motivation for logging its location. But let's keep that as a separate PR if anyone else needs to weigh in on it.

PR #1068, opened 2018-10-12
enter "n" to cancel setup

kelwang:
```
Russ [7:32 PM]
looks good! not urgent for tomorrow, but if a user enters ‘n’ in the confirm part, 
we should exit the cli
```


cmd/influx/setup.go
```diff
@@ -66,16 +66,16 @@ func getOnboardingRequest() (or *platform.OnboardingRequest) {
...
+	or = new(platform.OnboardingRequest)
+	fmt.Println(promptWithColor(`Welcome to InfluxDB 2.0!`, colorYellow))
+	or.User = getInput(ui, "Please type your primary username", "")
+	or.Password = getPassword(ui)
+	or.Org = getInput(ui, "Please type your primary organization name.\r\nOr ENTER to use \"default\"", "default")
+	or.Bucket = getInput(ui, "Please type your primary bucket name.\r\nOr ENTER to use \"default\"", "default")
+
+	if confirmed := getConfirm(ui, or); !confirmed {
+		fmt.Println("Setup is canceled.")
+		os.Exit(0)
```
goller: Because this was not a "normal" exit of setup this should be os.Exit(1)

PR #1077, opened 2018-10-13
Allow setting task status during creation

### LINKED ISSUE #1066: Task Creation does not persist status
When I hit the /api/v2/tasks endpoint with status in my payload this status is not persisted and therefore not returned on requests for that task.

mark-rushakoff:
Fixes #1066.


task/backend/bolt/bolt.go
```diff
@@ -93,8 +93,8 @@ func New(db *bolt.DB, rootBucket string) (*Store, error) {
 }
 
 // CreateTask creates a task in the boltdb task store.
-func (s *Store) CreateTask(ctx context.Context, org, user platform.ID, script string, scheduleAfter int64) (platform.ID, error) {
-	o, err := backend.StoreValidator.CreateArgs(org, user, script)
+func (s *Store) CreateTask(ctx context.Context, req backend.CreateTaskRequest) (platform.ID, error) {
```
lyondhill: Exactly the right decision. Once the arg list is over 3 we should do this for sure.

PR #1093, opened 2018-10-15
feat(logs/page): render logs tail

OfTheDelmer:
Closes influxdata/application-team-issues#131

Render tail of logs loop of logs. This is a small step toward rendering logs in the logs table.
_What was the problem?_
Logs were not viewable in the logs page
_What was the solution?_
Begin by starting a fetch interval to buffer in logs using ported actions from chronograf.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


alexpaxton:
LOOKS GREAT

PR #1099, opened 2018-10-16
Add tsi1 measurement cardinality stats.

benbjohnson:

Adds per-measurement cardinality stats to the TSI1 index.

_What was the problem?_

We need to quickly be able to fetch the number of series for every measurement in the system.

_What was the solution?_

This commit adds `MeasurementCardinalityStats` to `tsi1.Index`. The stats are maintained in each partition and all stats not in the non-active log file are written to disk on every compaction. At query time, every partition merges the non-active log file stats and the active log file stats and then the index merges every partition stats.

_Note: I need to rework the `LogFile` stats computation because there is a race condition where a series inserted and deleted in the same log file is double decremented.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/tsdb/tsi1/partition.go
+++ b/tsdb/tsi1/partition.go
@@ -1,6 +1,7 @@
 package tsi1
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -33,8 +34,13 @@ const (
 	CompactingExt = ".compacting"
 )
 
-// ManifestFileName is the name of the index manifest file.
-const ManifestFileName = "MANIFEST"
+const (
+	// ManifestFileName is the name of the index manifest file.
+	ManifestFileName = "MANIFEST"
+
+	// StatsFileName is the name of the file containing cardinality stats.
+	StatsFileName = "STATS"
+)
 
 // Partition represents a collection of layered index files and WAL.
 type Partition struct {
@@ -45,6 +51,9 @@ type Partition struct {
 	activeLogFile *LogFile         // current log file
 	fileSet       *FileSet         // current file set
 	seq           int              // file id sequence
+
+	// Measurement stats
+	stats MeasurementCardinalityStats
 
 	// Fast series lookup of series IDs in the series file that have been present
 	// in this partition. This set tracks both insertions and deletions of a series.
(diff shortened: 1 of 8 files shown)


tsdb/tsi1/stats.go
```diff
@@ -0,0 +1,233 @@
...
+	s[string(name)]++
+}
+
+// Dec decrements a measurement count by 1. Deleted if zero.
+func (s MeasurementCardinalityStats) Dec(name []byte) {
+	v := s[string(name)]
+	if v == 1 {
+		delete(s, string(name))
+	} else {
+		s[string(name)] = v - 1
```
jacobmarble: Does `s[string(name)]--` not work? Just curious because I've never thought to do `s[string(name)]++` as above.

jacobmarble:
Just had a think about my last comment. Cardinality isn't count of points, it's count of series keys, so of course we count each key individually.

tsdb/tsi1/stats.go
benbjohnson: Yeah, `s[string(name)]--` works but we have to check the value of `s[string(name)]` to see if it's `1` first. If we checked the value first and then did the unary `--` then it would perform the look up of the value twice so I store the value in `v`.

benbjohnson:
> _There are some println debug statements in there still._

Removed.

> _ Cardinality isn't count of points, it's count of series keys, so of course we count each key individually._

Yeah, we keep series cardinality as a whole in the series sets but tracking with a simple counter per measurement works without needing to allocate a bunch of series sets.

tsdb/tsi1/partition.go
```diff
@@ -669,11 +699,17 @@ func (p *Partition) createSeriesListIfNotExists(collection *tsdb.SeriesCollectio
 }
 
 func (p *Partition) DropSeries(seriesID tsdb.SeriesID) error {
+	// Ignore if the series is already deleted.
+	if !p.seriesIDSet.Contains(seriesID) {
```
e-dard: Is this a fix for an existing bug? Not sure how it relates to the stats count.

benbjohnson: Previously TSI wouldn't care if we double deleted a series, however, with stats counting double deletes would cause the stats to decrement twice.

benbjohnson:
@e-dard @jacobmarble I updated `tsi1.LogFile` to maintain a `stats` object instead of deriving it from the series sets. That fixes a bug where a series created & deleted in the same log file would mess up the counts.

PR #1100, opened 2018-10-16
Add CreatedAt and UpdatedAt fields to dashboards

### LINKED ISSUE #1098: Add `createdAt` and  `updatedAt` attributes to dashboards
And Allow sorting by those attributes.

desa:
Closes #1098

This PR introduces the meta fields `CreatedAt` and `UpdatedAt` to dashboards. Additionally it allows the user to specify if they would like to sort the results by those fields.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/dashboard_service.go
+++ b/http/dashboard_service.go
@@ -73,13 +73,6 @@ type dashboardResponse struct {
 }
 
 func (d dashboardResponse) toPlatform() *platform.Dashboard {
-	if len(d.Cells) == 0 {
-		return &platform.Dashboard{
-			ID:   d.ID,
-			Name: d.Name,
-		}
-	}
-
 	cells := make([]*platform.Cell, 0, len(d.Cells))
 	for i := range d.Cells {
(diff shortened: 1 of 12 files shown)


http/dashboard_service.go
```diff
@@ -210,6 +205,13 @@ func decodeGetDashboardsRequest(ctx context.Context, r *http.Request) (*getDashb
 		}
 	}
 
+	sortBy := qp.Get("sortBy")
+	if sortBy == "" {
+		sortBy = "ID"
```
fntlnz: Doing this here would make this harder to break default options if we change them in the future.
```
req.opts = platform.DefaultDashboardFindOptions
```

kelwang:
Another thing I remembered, not sure if you seen it yet. usually, in test case, if you call time.Now(), it usually includes monotonic values, sometimes it could cause the test fail. To get around with it. you might want to do time.Now().Round(0).

Issue #945, opened 2018-10-03
Telegraf Config service: storing the structures in bolt and inmem


kelwang:
creating the `testing` code as well

PR #1107, opened 2018-10-17
add telegraf store service

kelwang:
Closes #https://github.com/influxdata/platform/issues/945

- mostly common crud
- fixed a small issue for inmem/user_resource_mapping filter func, which is inconsistent with bolt. 
when the filter has multiple requirements.
- add new user resource mapping for telegraf
- add some helper method in user resource mapping

### DIFF
--- a/bolt/telegraf.go
+++ b/bolt/telegraf.go
@@ -1,0 +1,240 @@
+package bolt
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	bolt "github.com/coreos/bbolt"
+	"github.com/influxdata/platform"
+)
+
+var (
+	telegrafBucket = []byte("telegrafv1")
+)
+
+var _ platform.TelegrafConfigStore = new(Client)
+
+func (c *Client) initializeTelegraf(ctx context.Context, tx *bolt.Tx) error {
+	if _, err := tx.CreateBucketIfNotExists(telegrafBucket); err != nil {
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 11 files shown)


bolt/telegraf.go
```diff
@@ -0,0 +1,240 @@
...
+			return &platform.Error{
+				Code: platform.EEmptyValue,
+				Err:  err,
+			}
+		}
+		err = tx.Bucket(telegrafBucket).Delete(encodedID)
+		if err != nil {
+			return err
+		}
+		return c.deleteUserResourceMappings(ctx, tx, platform.UserResourceMappingFilter{
```
jademcgough: 👍 good to see this is being cleaned up on deletion. I'm going to add an issue to add this to our other resources.

bolt/user_resource_mapping.go
```diff
@@ -175,3 +179,20 @@ func (c *Client) deleteUserResourceMapping(ctx context.Context, tx *bolt.Tx, res
 
 	return tx.Bucket(userResourceMappingBucket).Delete(key)
 }
+
+func (c *Client) deleteUserResourceMappings(ctx context.Context, tx *bolt.Tx, filter platform.UserResourceMappingFilter) error {
+	ms, err := c.findUserResourceMappings(ctx, tx, filter)
```
jademcgough: doesn't necessarily need to happen in this PR, but I'd love to see `deleteUserResourceMapping` refactored to take a mapping, so that `deleteUserResourceMappings` could just call that for every mapping.

PR #1113, opened 2018-10-17
add toml for telegraf

kelwang:
Closes #https://github.com/influxdata/platform/issues/948

### DIFF
--- a/telegraf.go
+++ b/telegraf.go
@@ -3,6 +3,7 @@ package platform
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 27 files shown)


telegraf.go
```diff
@@ -46,6 +46,20 @@ type TelegrafConfig struct {
...
+	plugins := ""
+	for _, p := range tc.Plugins {
+		plugins += p.Config.TOML()
+	}
+	interval := time.Duration(tc.Agent.Interval * 1000000)
+	return fmt.Sprintf(`# Configuration for telegraf agent
+[agent]
+  ## Default data collection interval for all inputs
+  interval = "%s"
+%s`, interval.String(), plugins)
```
danielnelson: We should set all fields that are not commented in the example config.  Omitting any unset fields could be problematic since it is not defined what the defaults are or even if it works without them being set.

I feel like we will want more configurable settings here, but check with @russorat.

telegraf_test.go
```diff
@@ -144,3 +144,55 @@ func TestTelegrafConfigJSON(t *testing.T) {
...
+  ##   /var/log/**.log     -> recursively find all .log files in /var/log
+  ##   /var/log/*/*.log    -> find all .log files with a parent dir in /var/log
+  ##   /var/log/apache.log -> only read the apache log file
+  files = ["f1", "f2"]
+[[inputs.cpu]]
+[[outputs.file]]
+  ## Files to write to, "stdout" is a specially handled file.
+  files = ["stdout"]
+`
+	if result := tc.TOML(); result != want {
```
danielnelson: The way I'd probably do these tests is reparse the output and verify the values are set after round tripping.  This would tell you that the output is valid TOML, and that the settings are set.  The downside is it doesn't check comments but I think that may be an advantage because there will be less testcase maintenance required.

telegraf.go
kelwang: Hey @russorat, what do you think about this? I think we can have all other fields to be optional, so we can set those configs

russorat: if we are just returning a toml string, I don't see the downside to including the entire section of `[agent]` that comes in the default telegraf? including the commented out ones. otherwise, this will be "worse" than generating a default config from the command line because a user won't have the docs for all the fields. It's very possible that someone might download this config file, open it up, and customize as needed.

telegraf_test.go
kelwang: added parsing funcs

PR #1120, opened 2018-10-18
Fix/onboarding change

kelwang:
Close https://github.com/influxdata/platform/issues/1119
- removed default org, and bucket
- add retention period
- updated API, swagger, tests

screen shot:
https://gyazo.com/621365ddf318dd0c585e53ccf29b63ec

https://gyazo.com/ef2255f0b6a8f9a5882ab2fde1b343eb

### DIFF
--- a/cmd/influx/setup.go
+++ b/cmd/influx/setup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/influxdata/platform"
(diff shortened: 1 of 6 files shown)


cmd/influx/setup.go
```diff
@@ -70,12 +71,21 @@ func getOnboardingRequest() (or *platform.OnboardingRequest) {
 	fmt.Println(promptWithColor(`Welcome to InfluxDB 2.0!`, colorYellow))
 	or.User = getInput(ui, "Please type your primary username", "")
 	or.Password = getPassword(ui)
-	or.Org = getInput(ui, "Please type your primary organization name.\r\nOr ENTER to use \"default\"", "default")
-	or.Bucket = getInput(ui, "Please type your primary bucket name.\r\nOr ENTER to use \"default\"", "default")
+	or.Org = getInput(ui, "Please type your primary organization name.", "")
+	or.Bucket = getInput(ui, "Please type your primary bucket name.", "")
+	for {
+		rpStr := getInput(ui, "Please type your retention period in hours (exp 168 for 1 week).\r\nOr press ENTER for infinite.", "0")
```
goller: Would you put a constant  `const InfiniteRetention = 0` into platform/bucket and use it here?  I'm likely going to change the meaning of InfiniteRetention at some point and would like all places to be using that.

PR #1121, opened 2018-10-18
feat(cmd/influx): add org name option to repl

goller:
Require org or org-id for repl and add better error responses.

_What was the problem?_
influx repl required org id but would return a mysterious error.

_What was the solution?_
Add better error language, but also, allow org to be specified by name.

  - [x] Rebased/mergeable
  - [x] Tests pass


cmd/influx/repl.go
```diff
@@ -30,15 +32,39 @@ func init() {
 	if h := viper.GetString("ORG_ID"); h != "" {
 		replFlags.OrgID = h
 	}
+
+	replCmd.PersistentFlags().StringVarP(&replFlags.Org, "org", "o", "", "name of the organization")
```
kelwang: If we do either or, maybe we can add some text to explain, if both entered, only org-id will be used

Issue #962, opened 2018-10-04
add the ability to filter the tasks list by task name


russorat:
as a task builder
with viewer permissions or higher
when i view the tasks page
then i have the ability to filter the tasks based on the name of the task

PR #1122, opened 2018-10-18
Handle empty state for tasks with search term

AlirieGray:
Closes: https://github.com/influxdata/platform/issues/962

_What was the problem?_
If no tasks can be found, the empty state would tell the user that they had no tasks.

_What was the solution?_
Update the empty state if no tasks can be found with the current search term. 

  - [x] Rebased/mergeable
  - [x] Tests pass


chronograf/ui/src/tasks/components/TasksList.tsx
```diff
@@ -111,19 +112,33 @@ export default class TasksList extends PureComponent<Props> {
     onDelete(task)
   }
 
-  private get emptyState(): JSX.Element {
-    const {onCreate} = this.props
+  private get emptyStateDisplay(): JSX.Element {
```
bthesorceror: Could we move any getters that return a JSX.Element to their own components?

AlirieGray:
also: https://github.com/influxdata/platform/commit/f1f645ddad736fd4b1b6a00c2872ca79e1f4abad

PR #1144, opened 2018-10-19
feat(logs/scroll):  Add logs scroll handlers

OfTheDelmer:
Closes https://github.com/influxdata/applications-team-issues/issues/175

Adds simple scroll handlers to logs page with 
_What was the problem?_
Scroll handling needed to be added for rendering scroll position for live search
_What was the solution?_
Add hasScrolled and tableScrollToRow. Also included the extra `and !hasScrolled` guard to prevent scroll getting stuck in Firefox

  - [x] Rebased/mergeable
  - [x] Tests pass


chronograf/ui/src/logs/containers/logs_page/LogsPage.tsx
```diff
@@ -451,12 +455,22 @@ class LogsPage extends Component<Props, State> {
   /**
    * Handle scrolling to the top and resuming logs tail
    */
-  private handleScrollToTop = () => {}
+  private handleScrollToTop = () => {
+    if (!this.isLiveUpdating && this.shouldLiveUpdate) {
```
OfTheDelmer: Note in firefox this and tableScrollToRow race because rendering of scroll position isn't immediate. Causing the scroll to stick

OfTheDelmer: But, that's prevented in these changes by adding `hasScrolled` to `tableScrollToRow`

chronograf/ui/src/logs/containers/logs_page/LogsPage.tsx
```diff
@@ -92,6 +92,7 @@ type Props = StateProps & DispatchProps & WithRouterProps
 interface State {
   liveUpdating: boolean
   isOverlayVisible: boolean
+  hasScrolled: boolean
```
alexpaxton: does having more state cause more re-renders and affect performance?

OfTheDelmer: @alexpaxton it's a good question. The other use case this plays back in chronograf is to guard from setting `tableScrollToRow` when tableTime custom/relative is set to some particular point and we want to avoid sticking the scrollbar to the middle of the logs. However, given it's just two booleans we could probably just encode the four states into more readable enum.

OfTheDelmer:
- [ x ] Rebased
- [ x ] Make test passes

PR #1163, opened 2018-10-22
Add tasks to the influx cli

lyondhill:
TO facilitate this we created a http client implementation of TaskService. 

Three methods are "not yet implemented". I will be creating separate issues for those methods because the task service testing framework doesn't yet test those functions.


cmd/influx/task.go
```diff
@@ -0,0 +1,599 @@
...
+var taskCreateFlags TaskCreateFlags
+
+func init() {
+	taskCreateCmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create task",
+		Run:   taskCreateF,
+	}
+
+	taskCreateCmd.Flags().StringVarP(&taskCreateFlags.orgID, "org-id", "", "", "id of the organization that owns the task")
```
goller: Would you add org as a name as well?

http/task_service.go
```diff
@@ -130,12 +135,27 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request) (*getTasksReque
 func (h *TaskHandler) handlePostTask(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
+	tok, err := GetToken(r)
```
goller: authentication_middleware handles this now; use GetAuthorizer instead.

http/task_service.go
```diff
@@ -575,3 +601,260 @@ func decodeRetryRunRequest(ctx context.Context, r *http.Request) (*retryRunReque
 		RunID: i,
 	}, nil
 }
+
+type TaskService struct {
```
goller: Would you be able to add tests for the TaskService as well?  

The testing package has examples about how to use the service and http together to test client and server together.

lyondhill: The http.TaskService Implements the platform.TaskService interface (https://github.com/influxdata/platform/blob/master/task.go#L33). This make testing any task service very convenient weather its http, rpc, or the platform adaptor. We created a generic testing framework for all of the things that implement this interface (https://github.com/influxdata/platform/blob/master/task/servicetest/servicetest.go).

task/servicetest/servicetest.go
```diff
@@ -426,6 +431,20 @@ func testTaskConcurrency(t *testing.T, sys *System) {
 	extraWg.Wait()
 }
 
+func creds(t *testing.T, s *System) (orgID, userID platform.ID, token string) {
+	t.Helper()
+
+	if s.CredsFunc == nil {
+		return idGen.ID(), idGen.ID(), idGen.ID().String()
```
docmerlin: Token is just a stringified snowflake ID? something about that bothers me

http/task_service.go
goller: @lyondhill we did the same in `testing`  Do you have tests for the client in place?

lyondhill: Yes. They are in http/task_test.go

PR #1166, opened 2018-10-22
feat(http): buckets now have multiple retention rules

goller:
Closes #

Bucket retentions are now rules.  Currently, only one rule is allowed, but, this gives us flexibility in the future.  No rules indicate infinite retention.

_What was the problem?_
Retention policies were time durations and were not flexible enough to express more than one rule

_What was the solution?_
Many rules and rule types

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -3388,11 +3388,9 @@ components:
           $ref: "#/components/schemas/Owners"
         name:
           type: string
-        retentionPeriods:
+        retentionRules:
           type: array
-          default:
-            type: expire
-            duration: "infinite"
+          description: rules to expire or retain data.  No rules means data never expires.
           items:
             type: object
             properties:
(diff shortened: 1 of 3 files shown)


http/swagger.yml
```diff
@@ -3401,12 +3399,13 @@ components:
...
-                type: string
-                description: durations are an extension of the golang duration syntax
-                example: 1d
-                default: "infinite"
-      required: [organizationID, name, retentionPeriod]
+              everySeconds:
+                type: integer
+                description: duration in seconds for how long data will be kept in the database.
+                example: 86400
+                minimum: 1
```
fntlnz: I see that when there's nothing in this array the retention policy is just not applied and so it has the same behavior as if it was infinite.

There is no way to describe an infinite retention policy here?

goller: Right, retention rules are rules for when data is changed (expired, moved to s3, etc).

No rule means nothing is changed. The big problem having a rule that says "infinite" is when you have more than one rule.  

For example, imagine a rule that is infinite and one that expires data in 1 day.  Which one takes precedence?  Now there are dependencies between items in the array.

goller: @fntlnz here is describing infinite:
https://github.com/influxdata/platform/pull/1166/files/0527e515bf39da0beef952314f5b11a276486c84#diff-33aa613ab26f77a960b47c92852150afR3393

PR #1175, opened 2018-10-23
moved to_http and to_kafka to flux and update to latest flux

aanthony1243:
latest flux had some refactoring of the values package, this catches us up and moves some output functions to `flux` since they don't depend on platform.


aanthony1243:
the test failure really wasn't.  It failed to post a status update to github and marked the build unstable.

PR #1185, opened 2018-10-25
fix User mapping response

jademcgough:
* fixes DELETE endpoint handler
* responses are now proper k/v objects


http/user_resource_mapping_service.go
```diff
@@ -20,6 +20,34 @@ type UserResourceMappingService struct {
...
 }
 
+type userResourceResponse struct {
+	Links map[string]string `json:"links"`
+	platform.UserResourceMapping
+}
+
+func newUserResourceResponse(u *platform.UserResourceMapping) *userResourceResponse {
+	return &userResourceResponse{
+		Links:               map[string]string{},
```
desa: What do you think about including links to the user and the resource here?

http/user_resource_mapping_service.go
```diff
@@ -20,6 +20,34 @@ type UserResourceMappingService struct {
...
+	UserResourceMappings []*userResourceResponse `json:"userResourceMappings"`
+}
+
+func newUserResourcesResponse(opt platform.FindOptions, f platform.UserResourceMappingFilter, ms []*platform.UserResourceMapping) *userResourcesResponse {
+	rs := make([]*userResourceResponse, 0, len(ms))
+	for _, m := range ms {
+		rs = append(rs, newUserResourceResponse(m))
+	}
+	return &userResourcesResponse{
+		Links:                map[string]string{},
```
desa: Does the list response need links?

jademcgough: No, it doesn't.

http/user_resource_mapping_service.go
jademcgough: Yep, we have an issue for that from Tuesday, I can include it in this PR though :)

jademcgough:
I kind of want to restructure my response to return more than just user/resource IDs - might be nice to return more user info (like their name) to avoid having to send multiple requests for that info. The mapping resource seems like more of an internal implementation detail, and doesn't need to be fully exposed to the frontend.

jademcgough:
@desa added links

PR #1187, opened 2018-10-25
Allow coordinator to pick up existing tasks on startup

### LINKED ISSUE #1154: task: bolt store should reclaim tasks at startup
Currently, created tasks become orphaned after a restart of influxd. 

To fix this, all active tasks need to be claimed through the coordinator at startup. Should be straightforward to add after #1106 is addressed.

lyondhill:
Closes #1154


task/backend/coordinator/coordinator_test.go
```diff
@@ -136,7 +137,7 @@ func TestCoordinator_DeleteUnclaimedTask(t *testing.T) {
 	st := backend.NewInMemStore()
 	sched := mock.NewScheduler()
 
-	coord := coordinator.New(sched, st)
+	coord := coordinator.New(zap.NewNop(), sched, st)
```
mark-rushakoff: Prefer `zaptest.NewLogger(t)` in test.

task/backend/coordinator/coordinator.go
```diff
@@ -24,20 +26,45 @@ func WithLimit(i int) Option {
...
+			}
+		}
+		tasks, err = c.Store.ListTasks(context.Background(), backend.TaskSearchParams{
+			After: tasks[len(tasks)-1].Task.ID,
+		})
+		if err != nil {
+			c.logger.Error("failed list additional tasks", zap.Error(err))
+		}
+	}
+	return
```
mark-rushakoff: Unnecessary return.

task/backend/coordinator/coordinator_test.go
```diff
@@ -153,3 +154,26 @@ func TestCoordinator_DeleteUnclaimedTask(t *testing.T) {
...
 	}
 }
+
+func TestCoordinator_ClaimExistingTasks(t *testing.T) {
+	st := backend.NewInMemStore()
+	sched := mock.NewScheduler()
+
+	createChan := sched.TaskCreateChan()
+
+	_, err := st.CreateTask(context.Background(), backend.CreateTaskRequest{Org: 1, User: 2, Script: script})
```
mark-rushakoff: It would be better if this test had more than one "page" of tasks created.

lyondhill: I thought about that. Do you think it would be worth the time to create 1000+ records?

lyondhill: OO nevermind. page default is 100 :) I can do that.

PR #1195, opened 2018-10-25
Add local flag to influx command

desa:
This PR adds a local flag to the `influx` command that works with the data directly in the filesystem.
_What was the problem?_
People have been regularly losing their tokens and do not have a way to retrieve them.
_What was the solution?_
Add a local flag that allows users to interact directly with a local influxdb file.

  - [x] Rebased/mergeable
  - [x] Tests pass

Add support for:
- [x] organization
- [x] bucket
- [x] authorization
- [x] setup
- [x] query
- [x] task
- [x] user

### DIFF
--- a/cmd/influx/authorization.go
+++ b/cmd/influx/authorization.go
@@ -6,8 +6,10 @@ import (
 	"os"
 
 	"github.com/influxdata/platform"
+	"github.com/influxdata/platform/bolt"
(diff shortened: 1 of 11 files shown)


desa:
@leodido opened an issue for the flag cleanup. Going to merge this now. Will follow up with the fallback in an other PR.

Issue #1081, opened 2018-10-15
API user /authenticate endpoint change


kelwang:
Discussed with @goller, we would like to standardize our user signin, signout, reset password endpoint

We will have a /api/v2/authenticate
POST for sign in
PUT for reset password
DELETE for sign out

- [x] update swagger doc

- [x] update http endpoints

PR #1205, opened 2018-10-26
Fix/authenticate

kelwang:
Closes #https://github.com/influxdata/platform/issues/1081

- [x] fixed logger panic for signin, signout
- [x] updated the swagger doc for new endpoint
- [x] updated the endpoint
- [x] add put for password reset

### DIFF
--- a/http/session_handler.go
+++ b/http/session_handler.go
@@ -36,7 +36,7 @@ func (h *SessionHandler) handleSignin(w http.ResponseWriter, r *http.Request) {
 
 	req, err := decodeSigninRequest(ctx, r)
 	if err != nil {
-		h.Logger.Info("failed to decode request", zap.String("handler", "basicAuth"), zap.Error(err))
+		h.Logger.Info("failed to decode request", zap.Error(err))
 		EncodeError(ctx, err, w)
 		return
 	}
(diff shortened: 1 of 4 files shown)


http/session_handler.go
```diff
@@ -25,18 +26,37 @@ func NewSessionHandler() *SessionHandler {
 		Router: httprouter.New(),
 	}
 
-	h.HandlerFunc("POST", "/api/v2/signin", h.handleSignin)
-	h.HandlerFunc("POST", "/api/v2/signout", h.handleSignout)
+	h.HandlerFunc("POST", "/api/v2/authenticate", h.handleSignin)
+	h.HandlerFunc("PUT", "/api/v2/authenticate", h.handleResetPassword)
```
desa: What about making this `PUT /api/v2/users/:id/password`?

http/session_handler.go
```diff
@@ -62,6 +82,24 @@ type signinRequest struct {
 	Password string
 }
 
+type passwordResetRequest struct {
```
desa: This struct behaves differently from the others in this file. Namely it has the username and password on the request structure rather than passed in as headers of basic auth.

http/user_service.go
```diff
@@ -27,14 +28,74 @@ func NewUserHandler() *UserHandler {
...
+func (h *UserHandler) handlePutPassword(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	req, err := decodePasswordResetRequest(ctx, r)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	if err := h.BasicAuthService.ComparePassword(ctx, req.Username, req.PasswordOld); err != nil {
```
desa: What do you need this for?

kelwang: For the reset password? I need to add basic auth service as part of the onboarding service for unit testing, since onboarding will need to create password. The idea is the to use the same http testing frame work with inmem service for every services we can test.  Currently Onboarding service is lacking of http level unit testing.

kelwang: Good catch

PR #1216, opened 2018-10-29
fix(cmd/influx): require query command's org-id flag

mark-rushakoff:
Before, if you ran `influx query` and didn't have --org-id, you would
get an unhelpful runtime error like `id must have a length of 16 bytes`.
After this, you get a more helpful `required flag(s) "org-id" not set`
error.


goller:
@mark-rushakoff this code will need to be refactored after #1217 but I think it is ok for now.

mark-rushakoff:
Sure, this will reduce friction until #1217 is addressed.

Issue #947, opened 2018-10-03
Telegraf Config service: write the http server / client code


kelwang:
based on swagger doc

goller:
Closed by #1229

PR #1229, opened 2018-10-30
add http for telegraf

kelwang:
Closes #https://github.com/influxdata/platform/issues/947

change http json api to camel cases
add http handlers

### DIFF
--- a/telegraf.go
+++ b/telegraf.go
@@ -120,8 +120,8 @@ type telegrafConfigEncode struct {
 	ID        ID        `json:"id"`
 	Name      string    `json:"name"`
 	Created   time.Time `json:"created"`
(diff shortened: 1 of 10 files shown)


telegraf.go
```diff
@@ -205,7 +207,7 @@ func (tc *TelegrafConfig) MarshalJSON() ([]byte, error) {
 	}
 	for k, p := range tc.Plugins {
 		tce.Plugins[k] = telegrafPluginEncode{
-			Name:    p.Config.PluginName(),
+			Name:    convertPluginNameToCamel(p.Config.PluginName()),
```
goller: Hey @kelwang, I think we should stay as close as possible to telegraf for telegraf configuration. Would you keep these telegraf snake-case?

Issue #1236, opened 2018-10-30
telegrafs: api of telegraf should return multiple configurations and not sources


goller:
+type telegrafResponses struct {
+       Sources []*platform.TelegrafConfig `json:"sources"`
+       Links   link                       `json:"links"`
+}

goller:
Closed by #1240

PR #1240, opened 2018-10-30
fix telegraf sources

kelwang:
Closes #https://github.com/influxdata/platform/issues/1236

### DIFF
--- a/http/telegraf.go
+++ b/http/telegraf.go
@@ -69,8 +69,7 @@ type telegrafResponse struct {
 }
 
 type telegrafResponses struct {
-	Sources []*platform.TelegrafConfig `json:"sources"`
-	Links   link                       `json:"links"`
+	TelegrafConfigs []telegrafResponse `json:"configurations"`
 }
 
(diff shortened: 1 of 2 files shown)


http/telegraf.go
```diff
@@ -69,8 +69,8 @@ type telegrafResponse struct {
 }
 
 type telegrafResponses struct {
-	Sources []*platform.TelegrafConfig `json:"sources"`
-	Links   link                       `json:"links"`
+	Telegrafs []*platform.TelegrafConfig `json:"telegrafs"`
```
goller: I'd like this to be `configurations`

http/telegraf.go
```diff
@@ -84,7 +84,7 @@ func newTelegrafResponse(tc *platform.TelegrafConfig) telegrafResponse {
 
 func newTelegrafResponses(tcs []*platform.TelegrafConfig) telegrafResponses {
 	return telegrafResponses{
-		Sources: tcs,
+		Telegrafs: tcs,
 		Links: link{
```
goller: I don't think there is any need for a link here back to the plural endpoint. Probably remove `Links` for the multiple response.

PR #1242, opened 2018-10-30
Use snowflake for bolt task ids

### LINKED ISSUE #779: task: use snowflake IDs instead of sequential IDs
Tasks and runs are using sequential IDs, but for consistency with the rest of the system, they should be using snowflake IDs.

This may be a bit complicated change to do in one pass, so it's fine if this needs to be broken into converting run IDs in one PR and task IDs in a following PR.

docmerlin:
Closes #779


task/backend/storetest/storetest.go
```diff
@@ -769,9 +769,10 @@ from(bucket:"test") |> range(start:-1h)`
 		if err != nil {
 			t.Fatal(err)
 		}
+		fmt.Println(rc)
```
mark-rushakoff: Remove print.

Issue #1206, opened 2018-10-26
Me handler for /me


kelwang:
cc @desa 

We currently have the /me in users service, which shows the user name and id.
Ideally, we could make a me handler, to show all the user's resources and its link, including buckets, org, dashboards, telegrafs ....

kelwang:
dup with https://github.com/influxdata/platform/issues/1200

PR #1247, opened 2018-10-31
Resources endpoint

jademcgough:
closes #1200 #1206 

Add `/api/v2/resources` endpoint that returns resources that belong to the user.

I'll add a PR for `idpe` before this is merged, changing `ID` filters to `IDs` is going to break some etcd stuff.


http/resources_handler.go
```diff
@@ -0,0 +1,110 @@
...
+	ctx := r.Context()
+
+	a, err := platcontext.GetAuthorizer(ctx)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	var userID plat.ID
+	switch s := a.(type) {
```
goller: @kelwang just added a `GetUserID()` to the interface yesterday for this sorta thing:

https://github.com/influxdata/platform/blob/master/authz.go#L14

http/resources_handler.go
```diff
@@ -0,0 +1,110 @@
...
+func (h *ResourcesHandler) handleGetResources(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	a, err := platcontext.GetAuthorizer(ctx)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	var userID plat.ID
```
desa: What do you think about requiring userID to be supplied explicitly in the request? Possibly as a query parameter?

http/resources_handler.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"net/http"
+
+	plat "github.com/influxdata/platform"
+	platcontext "github.com/influxdata/platform/context"
+	"github.com/julienschmidt/httprouter"
+)
+
+const (
+	resourcesPath = "/api/v2/resources"
```
desa: What do you think about making the routes for retrieving the resources a user owns

```
/api/v2/user/:userID/buckets
/api/v2/user/:userID/dashboards
/api/v2/user/:userID/views
```
and then including links to each of them on the `usersResponse`?

inmem/view.go
```diff
@@ -25,20 +26,25 @@ func (s *Service) FindViewByID(ctx context.Context, id platform.ID) (*platform.V
 	return s.loadView(ctx, id)
 }
 
-func filterViewFn(filter platform.ViewFilter) func(d *platform.View) bool {
-	if filter.ID != nil {
-		return func(d *platform.View) bool {
-			return d.ID == *filter.ID
+func filterViewFn(filter platform.ViewFilter) func(v *platform.View) bool {
+	if len(filter.IDs) > 0 {
```
desa: What do you think about just retrieving all of the views by IDs explicitly like on like 46? This way we don't have to scan across all of the views.

http/resources_handler.go
jademcgough: @desa I like it. Would it make sense to change this route to `/api/v2/users/resources`?

http/resources_handler.go
jademcgough: Great! I figured we'd want something like that, since it's going to be used in a lot of places.

http/resources_handler.go
desa: Yeah I like that.

http/resources_handler.go
```diff
@@ -0,0 +1,110 @@
...
+		case plat.DashboardResourceType:
+			df.IDs = append(df.IDs, &m.ResourceID)
+		// case plat.TaskResourceType:
+		// 	tf.IDs = append(tf.IDs, &m.ResourceID)
+		case plat.ViewResourceType:
+			vf.IDs = append(vf.IDs, &m.ResourceID)
+		}
+	}
+
+	response := &resourcesResponse{}
```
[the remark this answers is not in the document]
jademcgough: That's a really good question. Would it be sufficient to request the nth page of each resource and stitch them together (or even take a fraction of each page, since that'd still be pretty big).

inmem/view.go
jademcgough: Is `FindViewByID` constant time? I thought it was doing a full scan, which would be `O(mn)` if that were true. I need to add some test cases here to make sure that the `IDs` filtering is actually working as intended, and I'll reexamine this when I do.

http/swagger.yml
```diff
@@ -3003,6 +3017,46 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  '/users/{userID}/resources':
```
goller: Would you add a link to this path in the user object?

http/resources_handler.go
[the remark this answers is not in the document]
desa: I'm not sure @goller do you have any thoughts?

inmem/view.go
desa: I think `FindViewByID` is constant time.

http/resources_handler.go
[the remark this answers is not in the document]
goller: @jademcgough that's fine to me.  taking the nth page of each is very straight-forward to think about.

mark-rushakoff:
Closing since this is fairly old and has a lot of conflicts.

PR #1251, opened 2018-10-31
fix(http): check Accept header for mime-type

### LINKED ISSUE #1238: telegrafs: check Accept header for mime-type; default response should be TOML
This should be a switch over the `Accept` header and not `Content-Type`

Additionally, the default case should be TOML

```
+       contentType := r.Header.Get("Content-Type")
+       switch contentType {
+       case "application/octet-stream":
+               w.Header().Set("Content-Type", "application/octet-stream")
+               w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s.toml\"", tc.Name))
+               w.WriteHeader(http.StatusOK)
+               w.Write([]byte(tc.TOML()))
+       case "application/toml":
+               w.Header().Set("Content-Type", "application/toml; charset=utf-8")
+               w.WriteHeader(http.StatusOK)
+               w.Write([]byte(tc.TOML()))
+       case "application/json":
+               fallthrough
+       default:
+               if err := encodeResponse(ctx, w, http.StatusOK, newTelegrafResponse(tc))
```

zhulongcheng:
Closes #1238

  - [x] Rebased/mergeable
  - [x] Tests pass


http/telegraf.go
```diff
@@ -141,24 +141,19 @@ func (h *TelegrafHandler) handleGetTelegraf(w http.ResponseWriter, r *http.Reque
...
 		w.Header().Set("Content-Type", "application/octet-stream")
 		w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s.toml\"", tc.Name))
 		w.WriteHeader(http.StatusOK)
 		w.Write([]byte(tc.TOML()))
-	case "application/toml":
-		w.Header().Set("Content-Type", "application/toml; charset=utf-8")
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte(tc.TOML()))
 	case "application/json":
 		fallthrough
```
goller: This code will need to be in the "application/json" case

```
if err := encodeResponse(ctx, w, http.StatusOK, newTelegrafResponse(tc)); err != nil {
		w.Header().Set("Content-Type", "application/toml; charset=utf-8")
			EncodeError(ctx, err, w)
			return
		}
```

kelwang:
I like the idea to default TOML output. 
The only question I have is the api endpoint consistency.  Since the GET /telegrafs endpoint's output is JSON. I wonder how did everyone else deal with this situation.

danielnelson:
Either way works for me, we will always explicitly ask for TOML from Telegraf, and I'm sure if someone clicked the link to the page they would get JSON since their browser is unlikely to accept TOML by default.  The only case where you might run into issues would be if you use curl since it accepts */*

http/telegraf.go
zhulongcheng: Updated

PR #1254, opened 2018-10-31
fix the logger, and get resturns nothing

kelwang:
Closes https://github.com/influxdata/platform/issues/1253

- fixed panic in the demo
- fixed get /telegrafs endpoint for user resource mapping


http/telegraf.go
```diff
@@ -162,9 +171,26 @@ func (h *TelegrafHandler) handleGetTelegraf(w http.ResponseWriter, r *http.Reque
...
+		}
+		f.ResourceID = *id
+	}
+
+	if idStr := q.Get("userId"); idStr != "" {
+		id, err := platform.IDFromString(idStr)
+		if err != nil {
+			return nil, err
+		}
+		f.ResourceID = *id
```
goller: Should this be user id?

http/telegraf.go
```diff
@@ -196,6 +222,7 @@ func (h *TelegrafHandler) handlePostTelegraf(w http.ResponseWriter, r *http.Requ
 	ctx := r.Context()
 	tc, err := decodePostTelegrafRequest(ctx, r)
 	if err != nil {
+		h.Logger.Info("failed to decode request", zap.Error(err))
```
goller: Probably debug here

http/telegraf.go
```diff
@@ -222,6 +249,7 @@ func (h *TelegrafHandler) handlePutTelegraf(w http.ResponseWriter, r *http.Reque
 	ctx := r.Context()
 	tc, err := decodePutTelegrafRequest(ctx, r)
 	if err != nil {
+		h.Logger.Info("failed to decode request", zap.Error(err))
```
goller: Probably debug here

http/telegraf.go
```diff
@@ -31,10 +33,16 @@ const (
 )
 
 // NewTelegrafHandler returns a new instance of TelegrafHandler.
-func NewTelegrafHandler(mappingService platform.UserResourceMappingService) *TelegrafHandler {
+func NewTelegrafHandler(
+	mappingService platform.UserResourceMappingService,
+	telegrafSvc platform.TelegrafConfigStore,
+	logger *zap.Logger,
```
goller: I forget where, but, @mark-rushakoff made a nice point where logger should be the first parameter right after context.Context if context is include.

I'd recommend changing this to:

```
func NewTelegrafHandler(
	logger *zap.Logger,
	mappingService platform.UserResourceMappingService,
	telegrafSvc platform.TelegrafConfigStore,
) *TelegrafHandler {```

PR #1258, opened 2018-10-31
WIP: add operation log for resources

### LINKED ISSUE #1233: Design Audit/Op log service

desa:
Closes #1233 

**Briefly describe your proposed changes:**
I'd like to get some initial feedback

**What was the problem?**
We want a way to view the history of operation against a resource. It is not entirely clear what specific values should be saved for each operation. I imagine we'll want some sort of plaintext description of what happened, but I'm not sure.

One concern that I have is that the log will likely consume more disk than the resource does and I think simply having a plaintext record of what took place may not be as valuable as having a way to undo the operation that took place.
 
**What was the solution?**
We use the key value log to store a each event that takes place for a resource.

**TODO**
- [x] Add operation log for dashboards
- [x] Add route to retrieve the operation log for dashboards
- [x] Add operation log for buckets
- [x] Add route to retrieve the operation log for bucket
- [x] Add operation log for users
- [x] Add route to retrieve the operation log for users
- [x] Add operation log for orgs
- [x] Add route to retrieve the operation log for orgs 
- [x] Add tests
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/bolt/dashboard.go
+++ b/bolt/dashboard.go
@@ -5,16 +5,30 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
 
 	bolt "github.com/coreos/bbolt"
 	"github.com/influxdata/platform"
+	platformcontext "github.com/influxdata/platform/context"
 )
 
 var (
 	dashboardBucket = []byte("dashboardsv2")
 )
 
+// TODO(desa): what do we want these to be?
+const (
+	dashboardCreatedEvent = "Dashboard Created"
+	dashboardUpdatedEvent = "Dashboard Updated"
+
+	dashboardCellsReplacedEvent = "Dashboard Cells Replaced"
+	dashboardCellAddedEvent     = "Dashboard Cell Added"
+	dashboardCellRemovedEvent   = "Dashboard Cell Removed"
+	dashboardCellUpdatedEvent   = "Dashboard Cell Updated"
+)
+
 var _ platform.DashboardService = (*Client)(nil)
+var _ platform.DashboardOperationLogService = (*Client)(nil)
 
 func (c *Client) initializeDashboards(ctx context.Context, tx *bolt.Tx) error {
 	if _, err := tx.CreateBucketIfNotExists([]byte(dashboardBucket)); err != nil {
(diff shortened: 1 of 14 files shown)


operation_log.go
```diff
@@ -0,0 +1,19 @@
+package platform
+
+import (
+	"context"
+	"time"
+)
+
+// OperationLogEntry is a record in an operation log.
+type OperationLogEntry struct {
+	Description string    `json:"description"`
```
kelwang: shall we make some enum types for this

operation_log.go
```diff
@@ -0,0 +1,19 @@
...
+	UserID      ID        `json:"userID,omitempty"`
+	Time        time.Time `json:"time,omitempty"`
+}
+
+// DashboardOperationLogService is an interface for retrieving the operation log for a dashboard.
+type DashboardOperationLogService interface {
```
kelwang: Not necessary in this pr. What do you think about embed DashboardOperationLogService as part of DashboardService,

like 
```go
type   DashboardService interface{
    DashboardOperationLogService
    
    // FindDashboardByID returns a single dashboard by ID.
   FindDashboardByID(ctx context.Context, id ID) (*Dashboard, error)
    ....
}
```

bolt/dashboard.go
```diff
@@ -5,16 +5,30 @@ import (
...
 	"github.com/influxdata/platform"
+	platformcontext "github.com/influxdata/platform/context"
 )
 
 var (
 	dashboardBucket = []byte("dashboardsv2")
 )
 
+// TODO(desa): what do we want these to be?
+const (
```
kelwang: maybe have a prefix, like "bolt/create dashboard"

operation_log.go
kelwang: not necessary in this pr by the way

desa: I think this is just a placeholder until we know what we really want from the frontend.

operation_log.go
desa: I thought about that, but my concern with doing that is that it makes the DashboardService even larger, but I think you're right it should eventually be on there.

bolt/dashboard.go
desa: I think thats I good idea eventually, but for now I'm not even sure we'll use these. They're really just a placeholder.

PR #1265, opened 2018-11-01
Add scanned value & bytes statistics.

benbjohnson:

This pull request adds `CursorStats` to the cursors, cursor iterator, and bubbles up to the `ResultSet`.

_What was the problem?_

We need to know how many values and how many uncompressed bytes are scanned by query.

_What was the solution?_

Maintain counts of values/bytes at the cursor level and aggregate up to the `ResultSet`.

Requires https://github.com/influxdata/flux/pull/230

  - [x] Rebased/mergeable
  - [x] Tests pass


query/bridges.go
```diff
@@ -32,8 +34,29 @@ func (b ProxyQueryServiceBridge) Query(ctx context.Context, w io.Writer, req *Pr
...
+
+	if w, ok := w.(http.ResponseWriter); ok {
+		if hasStats {
+			data, _ := json.Marshal(stats.Statistics())
+			w.Header().Set("Influx-Query-Statistics", string(data))
+		}
+	}
+
+	return n, nil
 }
```
benbjohnson: @nathanielc @aanthony1243 This function is pretty ugly because I have to type assert `w.(http.ResponseWriter)`. I would like to set the statistics trailer in caller, `http.FluxHandler` (https://github.com/influxdata/platform/blob/master/http/query_handler.go#L93) but I don't have access to the `flux.ResultIterator`.

Is the `ProxyQueryServiceBridge` necessary? The current implementation is only 4 lines (https://github.com/influxdata/platform/blob/master/query/bridges.go#L24-L37) and it seems like we could move those lines directly into the `http.FluxHandler`.

nathanielc: The problem is that the FluxHandler.ProxyQueryService  is not always a ProxyQueryServiceBridge implementation. 

I agree that these changes are ugly ;), we need to figure out a better way of dealing with this. 

Where are the stats getting lost? Its not clear to me where you are trying to propogate the stats to. Are you trying to get them to be returned to the end user client? I am missing some of the bigger picture here.

benbjohnson:
Per discussion with @nathanielc, keeping this PR open until `ProxyQueryServiceBridge` refactor is done to remove it from the `FluxHandler`.

PR #1266, opened 2018-11-01
open tracing context handling for tasks and for HTTP endpoints for tasks and query

docmerlin:
This adds open tracing for http endpoints for tasks and query.  Originally it was just my intent to add it for tasks, but then I realized it would be useful for query too, as both will often be behind backend software for a lot of people.

WIP

### DIFF
--- a/cmd/influxd/main.go
+++ b/cmd/influxd/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/influxdata/platform/snowflake"
+	"github.com/opentracing/opentracing-go"
 	nethttp "net/http"
 	_ "net/http/pprof"
 	"os"
@@ -33,6 +35,7 @@ import (
 	taskexecutor "github.com/influxdata/platform/task/backend/executor"
 	_ "github.com/influxdata/platform/tsdb/tsi1"
 	_ "github.com/influxdata/platform/tsdb/tsm1"
+	pzap "github.com/influxdata/platform/zap"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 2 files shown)


cmd/influxd/main.go
```diff
@@ -152,7 +155,10 @@ func run() error {
 		return err
 	}
 	defer logger.Sync()
-
+	tracer := new(pzap.Tracer)
+	tracer.Logger = logger
+	tracer.IDGenerator = snowflake.NewIDGenerator()
+	opentracing.SetGlobalTracer(tracer)
```
goller: what do you think of a newline here to let tracing and prom look logically separate?

http/query_handler.go
```diff
@@ -113,7 +113,7 @@ type postFluxASTResponse struct {
 // postFluxAST returns a flux AST for provided flux string
 func (h *FluxHandler) postFluxAST(w http.ResponseWriter, r *http.Request) {
 	var request langRequest
-	ctx := r.Context()
+	ctx := pcontext.OpenTracingCtx(r, "post_query_flux_ast")
```
goller: Now that these are added to the API would you update swagger.yaml for each endpoint?

This means we support the header (`Zap-Trace-Span`) here:
https://github.com/influxdata/platform/blob/master/zap/tracer.go#L19
coming in.

goller:
```yaml
            - in: header
              name: Zap-Trace-Span
              description: HTTP open tracing trace id.
              schema:
                type: string
```

Something like this for all endpoints

http/query_handler.go
```diff
@@ -184,7 +184,7 @@ type suggestionsResponse struct {
 
 // getFluxSuggestions returns a list of available Flux functions for the Flux Builder
 func (h *FluxHandler) getFluxSuggestions(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
+	ctx := pcontext.OpenTracingCtx(r, "get_query_flux_suggestions")
```
goller: feels like rather than encoding the http verb into the span name perhaps the span can be marked up with the open tracing convention `http.method` and `http.url`

https://opentracing.io/specification/conventions/

If you follow the conventions there would be no need to put these in every handler function, but, rather, it could be done as a middleware.

cmd/influxd/main.go
docmerlin: Sounds good

PR #1268, opened 2018-11-01
generate initial influxQL test cases

jlapacik:
This PR adds influxql test cases to platform. These tests are similar to the csv tests. Test cases are defined in `query/functions/random`. Each test case consists of 3 separate files. They are:

1. [name].influxql - file containing an influxql query
2. [name].in.json - InfluxQL json file specifying the input data (only a single result allowed)
3. [name].out.json - InfluxQL json file specifying the output results

  - [x] Rebased/mergeable
  - [x] Tests pass


jlapacik:
The following test queries that are added in this PR are failing:

* `SELECT count("n") FROM "ctr" WHERE time >= 0m AND time <= 210m`
* `SELECT count("n") FROM "ctr" WHERE time >= 210m AND time <= 630m`

due to this issue #1278 .

PR #1275, opened 2018-11-02
feat(http): add plan endpoint to analyze queries

goller:
Add /api/v2/query/plan endpoint to analyze queries and return spec, logical and physical plan.

_What was the solution?_
Update the docs for other analyzers.
Generate logical and physical plans for a query.

  - [x] Rebased/mergeable
  - [x] Tests pass


goller:
@nathanielc Here is the latest in exposing analysis functions of queries.  The logical and physical planners are constructed without any extra options... should there be?

http/query_handler.go
```diff
@@ -168,6 +170,99 @@ func (h *FluxHandler) postFluxSpec(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func decodePostPlanRequest(ctx context.Context, r *http.Request) (*postPlanRequest, error) {
+	req := &postPlanRequest{}
+	err := json.NewDecoder(r.Body).Decode(req)
+	if err != nil {
+		return nil, errors.MalformedDataf("invalid json: %v", err)
+	}
+
+	if req.Query == "" && req.Spec == nil {
```
jademcgough: One pattern I've seen used elsewhere in the codebase is to put this validation logic in a function, like `func (request *postPlanRequest) Valid() error`

http/query_handler.go
```diff
@@ -168,6 +170,103 @@ func (h *FluxHandler) postFluxSpec(w http.ResponseWriter, r *http.Request) {
...
+// postFluxPlan returns a flux plan for provided flux string
+func (h *FluxHandler) postFluxPlan(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	req, err := decodePostPlanRequest(ctx, r)
+	if err != nil {
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	var spec *flux.Spec = req.Spec
```
mark-rushakoff: Looks like this doesn't get assigned req.Spec.

http/query_handler.go
```diff
@@ -168,6 +170,103 @@ func (h *FluxHandler) postFluxSpec(w http.ResponseWriter, r *http.Request) {
...
 
+type fluxPlan struct {
+	Roots     []plan.PlanNode         `json:"roots"`
+	Resources flux.ResourceManagement `json:"resources"`
+	Now       time.Time               `json:"now"`
+}
+
+func newFluxPlan(p *plan.PlanSpec) *fluxPlan {
+	res := &fluxPlan{
+		Roots:     []plan.PlanNode{},
```
mark-rushakoff: This could be `make([]plan.PlanNode, len(p.Roots))` to ease up the GC ever so slightly.

http/query_handler.go
goller: @mark-rushakoff this is assigned here:
https://github.com/influxdata/platform/pull/1275/files#diff-83d50194a0d56fc3123d19ec0d0ec577R251

http/query_handler.go
goller: Fixed in #1280

nathanielc:
In general I like this idea, we need some kind of "explain" equivalent. This current implementation is to restrictive of our planner implementation. Since the planners internal structures are being directly exposed this means we will not be able to easy make updates to the planner.

We should make an explicit design around how we want to expose an "explain" like feature.

PR #1276, opened 2018-11-02
added (unused) stats struct, and piped written data to output table

aanthony1243:
Closes #

_What was the problem?_
_What was the solution?_

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/functions/outputs/to.go
+++ b/query/functions/outputs/to.go
@@ -398,6 +398,33 @@ func (d ToDependencies) Validate() error {
 	return nil
 }
 
+type Stats struct {
+	NRows   int
+	MaxTime time.Time
+	MinTime time.Time
+	NFields int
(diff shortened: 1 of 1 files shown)


query/functions/outputs/to.go
```diff
@@ -513,11 +548,28 @@ func writeTable(t *ToTransformation, tbl flux.Table) error {
 				}
 			})
 
+			mstats := Stats{
```
wolffcm: At the top of this file you call `RegisterProcedureWithSideEffect`, which is good.  

I think that the *function* should also be registered with a side effect, i.e., call `RegisterFunctionWithSideEffect` instead of `RegisterFunction` in the `init` method.  I'm not sure what this does, but it would be good to be consistent.

PR #1312, opened 2018-11-06
Implement ResultSet and GroupResultSet for protobuf streams

stuartcarnie:
This PR provides new implementations of a number of components required for merging data streams from storage.

The `reads.ResponseWriter` is an existing type that is responsible for streaming data from local TSM data. It transforms a local `reads.ResultSet` or `reads.GroupResultSet` into a stream of Protocol buffer messages that can be serialized to a stream, such as a network.

* `ResultSetStreamReader` implements a protobuf reader for these frames and produces an implementation of a `reads.ResultSet` on the client.
`GroupResultSetStreamReader` implements a protobuf reader for these frames that produces an implementation of a `reads.GroupResultSet` on the client.

By restoring these network streams to an implementation of `reads.ResultSet` and `reads.GroupResultSet`, we can then merge these instances with local TSM data, which is also surfaced as `reads.ResultSet` and `reads.GroupResultSet`.

There are three APIs which implement merging.

* `reads.NewMergedResultSet` takes a slice of `reads.ResultSet` and produces a single, merged `reads.ResultSet`

* `reads.NewGroupNoneMergedGroupResultSet` takes a slice of `reads.GroupResultSet` instances that use the `datatypes.GroupNone` group strategy to produce a single group

* `reads.NewGroupByMergedGroupResultSet` takes a slice of `reads.GroupResultSet` instances that use the `datatypes.GroupBy` group strategy to produce a group per partition key.


pkg/bytesutil/bytesutil.go
```diff
@@ -130,6 +130,31 @@ func Intersect(a, b [][]byte) [][]byte {
 	return other
 }
 
+// CompareSlice returns an integer comparing two sorted slices of byte slices
+// lexicographically.
+// The result will be 0 if a==b, -1 if a < b, and +1 if a > b.
+func CompareSlice(a, b [][]byte) int {
```
e-dard: What do you think about moving this into the `slices` package? `slices.CompareBytes` or something?

pkg/bytesutil/bytesutil.go
```diff
@@ -130,6 +130,31 @@ func Intersect(a, b [][]byte) [][]byte {
 	return other
 }
 
+// CompareSlice returns an integer comparing two sorted slices of byte slices
+// lexicographically.
+// The result will be 0 if a==b, -1 if a < b, and +1 if a > b.
```
e-dard: This implementation isn't what I would expect. If I understand it correctly it's saying:

```
a [0, 1, 2]
b [0, 1, 1, 3]
```

That `b` comes before `a` and the function will return `1`.

In my mind, a lexical comparitor would always return the shorter of two arguments given arguments of differing lengths, no matter what the contents.

I would expect a comparison for two `[][]byte` to have something like this at the beginning:

```go
if len(a) < len(b) {
    return -1
} else if len(b) > len(a) {
    return 1
}
```

If this behaviour is needed specifically for your use-case I wonder whether the function should be defined somewhere in the package of the caller?

pkg/testing/gen/tags_sequence.go
```diff
@@ -0,0 +1,93 @@
+package gen
+
+import (
+	"fmt"
+	"math"
+	"sort"
+
+	"github.com/influxdata/influxdb/models"
```
e-dard: Unless there is a good reason, we shouldn't be importing `models` from `influxdb`, but should be using the `platform` package.

storage/reads/mergegroupresultset.go
```diff
@@ -0,0 +1,251 @@
...
+func (r *groupByMergedGroupCursor) PartitionKeyVals() [][]byte {
+	return r.vals
+}
+
+type groupCursorItem struct {
+	grs GroupResultSet
+	gc  GroupCursor
+}
+
+func ComparePartitionKey(a, b [][]byte, nilVal []byte) int {
```
e-dard: This should probably be documented to explain how the `nilVal` fits in, or just unexported if it's not relevant outside of `reads`.

storage/reads/mergegroupresultset.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		if v := bytes.Compare(av, bv); v == 0 {
+			i++
+			continue
+		} else {
+			return v
+		}
+	}
+
+	if i < len(b) {
```
e-dard: Same comments apply as before with respect to a longer `b` where the first elements are `<` the first elements of an `a`.

e-dard:
@zeebo can you also take a look at this today please?

pkg/bytesutil/bytesutil.go
stuartcarnie: Interesting, as I look at it this way:

```
a = ["a", "b", "d"]
b = ["a", "b", "b", "c"]
```

`a` and `b` could be rewritten:

`a = abd`
`b = abbc` 

Sorting those two strings should result in `a` sequenced before `b`, which is the behavior of the function. It is only if `a` is a prefix of `b` do we care about the length. I could add additional documentation to indicate this is the behavior.

storage/reads/mergegroupresultset.go
stuartcarnie: Agree – I'll demote it

pkg/testing/gen/tags_sequence.go
stuartcarnie: Fixed

storage/reads/mergegroupresultset.go
stuartcarnie: Ordinarily, partition keys are the same number of values (byte slices), as they represent the values of the selected tags we are grouping by, i.e. the `GroupBy` keys. 

We reach this point if all the slices in `a` and `b` are equal. `i` may have exceeded the length of either `a` or `b`, given the earlier `for` loop:

```go
for i < len(a) && i < len(b) {
```

Therefore, this final test of `if i < len(b)` means `b` is longer than `a`, therefore `a` is smaller.

pkg/bytesutil/bytesutil.go
stuartcarnie: Yes – good call. I didn't notice that package!

pkg/bytesutil/bytesutil.go
stuartcarnie: Also, I'll adjust the comment:

```go
// CompareSlice returns an integer comparing two sorted slices of byte slices
// lexicographically.
```

as **sorted**" is irrelevant. That leaked from my desire to compare two sets of tag keys, which must be sorted.

e-dard: Yeah sounds good. Just wasn't sure if the behaviour was as intended 👍

pkg/slices/bytes_test.go
```diff
@@ -76,3 +81,66 @@ func TestCopyChunkedByteSlices_multipleChunks(t *testing.T) {
...
+			a:   ss("aaa", "ccc"),
+			b:   ss("aaa", "bbb", "ccc"),
+			exp: 1,
+		},
+
+		{
+			a:   ss("aaa", "aaa"),
+			b:   ss("aaa", "bbb", "ccc"),
+			exp: -1,
+		},
```
zeebo: The implementation is correct, but empty slices seems like a good missed test case.

storage/reads/group_resultset.go
```diff
@@ -76,7 +91,16 @@ func NewGroupResultSet(ctx context.Context, req *datatypes.ReadRequest, newCurso
...
 }
 
-var nilKey = [...]byte{0xff}
+// nilSort values determine the lexicographical order of nil values in the
+// partition key
+var (
+	// nil sorts lowest
+	nilSortLo = []byte{0x00}
+	// nil sorts highest
+	nilSortHi = []byte{0xff} // sort nil values
```
zeebo: This isn't actually higher than any byte slice. `[]byte{0xff, 0xff}`, for example. Is this because we know the space of things it's comparing against can't contain a `0xff` byte?

stuartcarnie: Correct, each value of a partition key is expected to be valid UTF-8, therefore it could not start with `0xff`. Additionally, `0x00` being a single byte would sort equally or lower than any other value starting with `0x00`.

pkg/slices/bytes_test.go
stuartcarnie: Good call – I'll add additional test cases.

stuartcarnie:
@e-dard / @zeebo I'm going to add additional sanity checks and unit tests to 

* `ResultSetStreamReader` to verify series keys are correctly ordered
* `GroupResultSetStreamReader` to verify partition keys are correctly ordered

pkg/slices/bytes_test.go
stuartcarnie: Done

PR #1329, opened 2018-11-08
fix(tsdb): eliminate data race from *SeriesIDSet.Clone

mark-rushakoff:
And add a test to cover that.

The data race would look roughly like:

```
WARNING: DATA RACE
Write at 0x00c000024e18 by goroutine 8:
  github.com/RoaringBitmap/roaring.(*roaringArray).markAllAsNeedingCopyOnWrite()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaringarray.go:881 +0x6b
  github.com/RoaringBitmap/roaring.(*roaringArray).clone()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaringarray.go:266 +0x808
  github.com/RoaringBitmap/roaring.(*Bitmap).Clone()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaring.go:385 +0x58
  github.com/influxdata/platform/tsdb.(*SeriesIDSet).CloneNoLock()
      /Users/mr/go/src/github.com/influxdata/platform/tsdb/series_set.go:229 +0x73
  github.com/influxdata/platform/tsdb.(*SeriesIDSet).Clone()

Previous write at 0x00c000024e18 by goroutine 7:
  github.com/RoaringBitmap/roaring.(*roaringArray).markAllAsNeedingCopyOnWrite()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaringarray.go:881 +0x6b
  github.com/RoaringBitmap/roaring.(*roaringArray).clone()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaringarray.go:266 +0x808
  github.com/RoaringBitmap/roaring.(*Bitmap).Clone()
      /Users/mr/go/pkg/mod/github.com/!roaring!bitmap/roaring@v0.4.16/roaring.go:385 +0x58
  github.com/influxdata/platform/tsdb.(*SeriesIDSet).CloneNoLock()
      /Users/mr/go/src/github.com/influxdata/platform/tsdb/series_set.go:229 +0x73
  github.com/influxdata/platform/tsdb.(*SeriesIDSet).Clone()
      /Users/mr/go/src/github.com/influxdata/platform/tsdb/series_set.go:223 +0x7b
```


mark-rushakoff:
The other approach I could imagine here is to decide at runtime whether we need to Lock or RLock, based on whether the bitmap is COW. But making that decision seems inherently racy. 

Or maybe there's a way to patch roaring bitmaps so that cloning a COW bitmap doesn't require a write lock. The fact that cloning involved a write is what's most surprising here in the first place.

e-dard:
@mark-rushakoff in general, this doesn't seem to have impacted the benchmark I run much. However, I need to check more carefully on Monday `CloneNoLock()`. In the clone without a lock case, it would be fine to have concurrent access if you could guarantee there were no writers. However it seems like concurrent clones of the same bitmap might not be safe. Not sure though, and will take another look next week.

This looks like it might be all we need.

PR #1333, opened 2018-11-09
Add a Flux editor to the VEO

chnn:
editor

Adds a basic Flux editor to the VEO. Queries can be run and saved to a cell. Editing a cell will populate the Flux editor with the cell's query. All cells are assumed to be Flux cells for now. It is also assumed that a Flux cell only has a single query.


ui/src/shared/components/TimeMachine.tsx
```diff
@@ -92,21 +94,21 @@ class TimeMachine extends PureComponent<Props> {
     )
   }
 
-  private get customizationPanels(): JSX.Element {
+  private get bottom(): JSX.Element {
```
bthesorceror: Can we move this getter to a separate component? I think we should move away from getters that return `JSX.Element` in most cases.

chnn: sounds like a good guideline to follow, I'll update this

Issue #1336, opened 2018-11-09
bug: transpiler group by time 


jlapacik:
Given the following input data:

[...]

PR #1337, opened 2018-11-09
transpiler test case for basic group by time

jlapacik:
This is the test case that produces #1336 . It is skipped for right now.


jlapacik:
Incorrect input data was used for this test case. `basic_random_group_by_time_sum.in.json` has been updated with the correct input data and I have un-skipped the test case. The test passes and so I am closing #1336 .

PR #1339, opened 2018-11-09
transpiler test case for grouping by time interval (windowing)

jlapacik:
This is the test case that produced #1338 . It is skipped because currently failing.


jlapacik:
I updated this test case to use correct input data. Currently the test still fails due to slight differences in the serialized floating point values.

PR #1342, opened 2018-11-09
ci: re-enable race builds in circle config

mark-rushakoff:
I looked through the past few days of Circle Go failures, and they all
appeared to be real test errors, indicating that the -p=8 flag most
likely fixed the linker OOMs we were seeing.

If the race-enabled tests continue to OOM, we can try setting -p=2 to
match the number of CPUs actually available on our Circle container.


mark-rushakoff:
I've run this branch ten times and it's passed every time: https://circleci.com/gh/influxdata/workflows/platform/tree/ci%2Fre-enable-race-tests

I don't know what frequency we were seeing OOMs on race tests before, but this is at least encouraging.

PR #1345, opened 2018-11-12
reduce parsing and copying of tags

zhulongcheng:
Closes #

This PR is to reduce parsing and copying of tags in `ExplodePoints`.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/tsdb/explode.go
+++ b/tsdb/explode.go
@@ -45,19 +45,19 @@ func ExplodePoints(org, bucket platform.ID, points []models.Point) ([]models.Poi
 	ob := EncodeName(org, bucket)
 	name := string(ob[:])
 
-	var tags models.Tags
+	tags := make(models.Tags, 2)
 	for _, pt := range points {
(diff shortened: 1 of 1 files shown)


tsdb/explode.go
```diff
@@ -45,19 +45,19 @@ func ExplodePoints(org, bucket platform.ID, points []models.Point) ([]models.Poi
 	ob := EncodeName(org, bucket)
 	name := string(ob[:])
 
-	var tags models.Tags
+	tags := make(models.Tags, 2)
 	for _, pt := range points {
-		t := pt.Time()
+		tags = tags[:1]
```
e-dard: I feel like this would read in a better way if you did this:

```go
tags  = tags[:2]
tags[1] = models.NewTag(MeasurementTagKeyBytes, pt.Name())
pt.ForEachTag(func(k, v []byte) bool {
...
...
```

PR #1349, opened 2018-11-12
Add end-to-end write/query integration testing.

benbjohnson:

Adds integration testing and an end-to-end write+query test.

_What was the problem?_

We don't have any integration tests.

_What was the solution?_

Add an integration test.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/cmd/influxd/main.go
+++ b/cmd/influxd/main.go
@@ -3,15 +3,14 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
+	"net"
 	nethttp "net/http"
 	_ "net/http/pprof"
 	"os"
 	"path/filepath"
 	"sync"
 	"time"
-
-	"github.com/influxdata/platform/snowflake"
-	"github.com/opentracing/opentracing-go"
 
 	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/bolt"
@@ -26,6 +25,7 @@ import (
 	"github.com/influxdata/platform/nats"
 	"github.com/influxdata/platform/query"
 	_ "github.com/influxdata/platform/query/builtin"
+	"github.com/influxdata/platform/snowflake"
 	"github.com/influxdata/platform/source"
 	"github.com/influxdata/platform/storage"
 	"github.com/influxdata/platform/storage/readservice"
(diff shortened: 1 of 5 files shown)


cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+// MustShutdown stops the program and cleans up temporary paths. Fail on error.
+func (m *Main) MustShutdown(tb testing.TB, ctx context.Context) {
+	tb.Helper()
+	if err := m.Shutdown(ctx); err != nil {
+		tb.Fatal(err)
+	}
+}
+
+// MustSetup creates a new user, bucket, org, and auth token. Fail on error.
+func (m *Main) MustSetup(tb testing.TB) {
```
goller: What do you think about using `http.SetupService` ?

The nice thing about using it, is that we would have end-to-end testing with the HTTP client as well.

code would be something like:

```go
svc := &http.SetupService{
    Addr: m.URL(),
}

setupResponse := svc.Generate(ctx, &platform.OnboardingRequest{
    User: "USER",
    Password: "PASSWORD",
    Org: "ORG",
    Bucket: "BUCKET",
})
```

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+func TestMain_Write(t *testing.T) {
+	m := MustRunMain(t, ctx)
+	m.MustSetup(t)
+	defer m.MustShutdown(t, ctx)
+
+	// Execute single write against the server.
+	if resp, err := http.DefaultClient.Do(m.MustNewHTTPRequest(t, "POST", fmt.Sprintf("/api/v2/write?org=%s&bucket=%s", m.OrgID, m.BucketID), `m,k=v f=0i 946684800000000000`)); err != nil {
```
goller: What do you think about using the `http.WriteService`? (I'm writing this comment after the one below about SetupService.  The nice thing about using the service is that we can get a more "full" integration using our HTTP client used in our CLI

would look something like this:

```go
svc := &http.WriteService{
    Addr: m.URL,
    Token: m.Token,
}
r := strings.NewReader(`m,k=v f=0i 946684800000000000`)
if err := svc.Write(ctx, m.OrgID, m.BucketID, r); err != nil {
   t.Fatal(err)
}


```

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+		t.Fatal(err)
+	} else if err := resp.Body.Close(); err != nil {
+		t.Fatal(err)
+	} else if resp.StatusCode != http.StatusNoContent {
+		t.Fatalf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	// Query server to ensure write persists.
+	q := `{"query":"from(bucket:\"BUCKET\") |> range(start:2000-01-01T00:00:00Z,stop:2000-01-02T00:00:00Z)"}`
```
goller: This one has more steps trickier.

If you just want to get raw bytes back:

```go
svc := &http.FluxService{
    Addr: m.URL,
    Token: m.Token,
}

req := http.QueryRequest{
    Query: `from(bucket:"BUCKET") |> range(start:2000-01-01T00:00:00Z,stop:2000-01-02T00:00:00Z)`
}.WithDefaults()

q, err := req.ProxyRequest()
if err != nil {
   t.Fatal(err)
}

w := new(bytes.Buffer)
n, err := svc.Query(ctx, w, q) // n is the # of bytes

```

I don't exactly remember the reason it was called a ProxyRequest... do you remember @nathanielc ?

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+	AuthToken string
+
+	// Standard in/out/err buffers.
+	Stdin  bytes.Buffer
+	Stdout bytes.Buffer
+	Stderr bytes.Buffer
+}
+
+// NewMain returns a new instance of Main.
+func NewMain() *Main {
```
mark-rushakoff: What do you think about just passing in the testing.TB to NewMain and holding that in the Main struct, so that all the panic calls can be tb.Fatal and the Must calls don't need to take tb as an argument?

The only obvious downside would be that you couldn't easily create one Main to use in multiple subtests.

cmd/influxd/main_test.go
mark-rushakoff: +1 to using the the http services, to keep the tests more readable and to further exercise our HTTP wrappers.

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+	m.AuthToken = setupResponse.Auth.Token
+}
+
+// MustNewHTTPRequest returns a new http.Request with base URL and auth attached. Fail on error.
+func (m *Main) MustNewHTTPRequest(tb testing.TB, method, rawurl, body string) *http.Request {
+	req, err := http.NewRequest(method, m.URL()+rawurl, strings.NewReader(body))
+	if err != nil {
+		tb.Fatal(err)
+	}
+	req.Header.Set("Authorization", "Token "+m.AuthToken)
```
mark-rushakoff: If bringing in platform/http, this would be `phttp.SetToken`.

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return m
+}
+
+// Run executes the program with additional arguments to set paths and ports.
+func (m *Main) Run(ctx context.Context, args ...string) error {
+	args = append(args, "--bolt-path", filepath.Join(m.Path, "influxd.bolt"))
+	args = append(args, "--engine-path", filepath.Join(m.Path, "engine"))
+	args = append(args, "--nats-path", filepath.Join(m.Path, "nats"))
+	args = append(args, "--http-bind-address", ":0")
```
mark-rushakoff: This should be `127.0.0.1:0` so that it doesn't bind to `0.0.0.0`.

cmd/influxd/main.go
```diff
@@ -37,54 +37,143 @@ import (
...
+	engine     *storage.Engine
+
+	httpPort   int
+	httpServer *nethttp.Server
+
+	natsServer *nats.Server
+
+	scheduler *taskbackend.TickScheduler
+
+	logger *zap.Logger
```
mark-rushakoff: I kind of want `logger` to be exported so it could be directly reassigned in test... but it's probably fine unexported.

cmd/influxd/main_test.go
goller: @mark-rushakoff the http services have set this for ya if you add them to the service like:

```go
svc := &http.WriteService{
    Addr: m.URL,
    Token: m.Token,
}
```

benbjohnson:
@goller @mark-rushakoff I updated the PR to use the `http` services to get better coverage. This PR is ready for re-review.

cmd/influxd/main_test.go
benbjohnson: I like explicitly passing `t` so that it's obvious where code can fail. Also, it's a pretty short variable so it's not too bad. :)

I did change the method names from `MustXXX` to `XXXorFail()` since `Must` is usually used for panicking. I couldn't think of a better name.

cmd/influxd/main.go
benbjohnson: I started looking at this but `logger` depends on command line arguments so it'd take a bit of a refactor to separate it and make it exported.

cmd/influxd/main_test.go
```diff
@@ -0,0 +1,218 @@
...
+	m.Bucket = results.Bucket
+	m.Auth = results.Auth
+}
+
+func (m *Main) FluxService() *http.FluxService {
+	return &http.FluxService{Addr: m.URL(), Token: m.Auth.Token}
+}
+
+// MustNewHTTPRequest returns a new nethttp.Request with base URL and auth attached. Fail on error.
+func (m *Main) MustNewHTTPRequest(method, rawurl, body string) *nethttp.Request {
```
goller: Now that you are using services, I think these Must functions and jsonSetupResponses are no longer needed.

PR #1351, opened 2018-11-12
feat(task): integrate task logs with system buckets

mark-rushakoff:
- Brought over enterprise's QueryLogReader, with small adjustments
  - Time filters are for the run's ScheduledFor field, per spec
- Adjusted run log timestamps for consistent formatting:
  - ScheduledFor is RFC3339 because it's a whole-second timestamp
  - StartedAt, FinishedAt use RFC3339Nano for high precision
- Several test adjustments to use relative time, for easier integration
  with storage retention

Race tests are expected to fail right now until #1329 is merged, but this PR should be otherwise reviewable.


task/backend/query_logreader.go
```diff
@@ -0,0 +1,342 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/semantic"
+	"github.com/influxdata/platform"
+	pctx "github.com/influxdata/platform/context"
+	"github.com/influxdata/platform/query"
+)
+
+var errAuthorizerNotSupported = errors.New("your authorizer is not supported, please use *platform.Authorization as authorizer")
```
docmerlin: This should be a public variable so people can check against it should they need to.

mark-rushakoff: I don't really see what having that error exported would achieve. Wouldn't it be better for the caller to check the input type against `*platform.Authorization`, rather than looking for a specific error that just reports the input type wasn't a `*platform.Authorization`?

This code was copied from a section of enterprise that I'm not really familiar with, so maybe I'm missing something.

PR #1357, opened 2018-11-12
opentracing

docmerlin:
Pass open tracing spans through the context from tasks to the query engine.

### DIFF
--- a/task/backend/scheduler.go
+++ b/task/backend/scheduler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


gianarb:
Probably what I am saying has no sense and it brings the codebase too close to opentracing but what I am doing in cloud 1 is to pass the `Span` itself using a function `WithSpan` (by default it has a noop span). In this way, I can add more context to a span (adding tags for example) from inside the function.

In my case I have a scheduler that executes steps. The executor creates the span and it injects the span in every step (a different one). So in this case from inside the step, I can add more context to the span related to what's going on inside it.

PR #1361, opened 2018-11-13
fix(ui/logs): time point selection dropdown

OfTheDelmer:
Closes #https://github.com/influxdata/applications-team-issues/issues/177

These changes just update the point in time dropdown to use clockface and additional changes will follow to actually add selected time ranges to the redux state. 
_What was the problem?_
The styling for the custom time selection needed to be ported and the dropdown needed to be updated to use a clockface dropdown. However, clicking on the open calender was also considered outside the clockface dropdown causing it to close when a time was being selected. 
_What was the solution?_
- port over the styles for the custom time point selector 
- portal the custom time overlay as a child of the the clockface dropdown so clicks would appear to be internal to the clockface dropdown
 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/logs/components/PointInTimeDropDown.tsx
```diff
@@ -139,22 +107,17 @@ class TimeRangeDropdown extends Component<Props, State> {
...
-    this.setState({isOpen: !this.state.isOpen})
+    return 'custom'
   }
 
   private handleCloseCustomTime = () => {
     this.setState({isTimeSelectorOpen: false})
   }
 
-  private handleOpenCustomTime = () => {
+  private handleOpenCustomTime = e => {
```
OfTheDelmer: Prevent clicking date picker from closing the dropdown with no custom time actually applied.

PR #1366, opened 2018-11-13
Add secret service interface and boltdb implementation

### LINKED ISSUE #1364: Add SecretService Interface and implement in boltdb

desa:
Closes #1364

This PR introduces a service for storing user secrets. The current implementation stores secrets as base64 encoded strings. This is marginally better than plaintext. A follow up to this work will be to offer the ability to store these secrets in vault instead.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/bolt/secret.go
+++ b/bolt/secret.go
@@ -1,0 +1,191 @@
+package bolt
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+
+	bolt "github.com/coreos/bbolt"
+	"github.com/influxdata/platform"
+)
+
+var (
+	secretBucket = []byte("secretsv1")
+)
+
+var _ platform.SecretService = (*Client)(nil)
+
+func (c *Client) initializeSecretService(ctx context.Context, tx *bolt.Tx) error {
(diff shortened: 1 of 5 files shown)


bolt/secret.go
```diff
@@ -0,0 +1,191 @@
...
+	// store the secret value base64 encoded so that it's marginally better than plaintext
+	v := make([]byte, base64.StdEncoding.DecodedLen(len(val)))
+	if _, err := base64.StdEncoding.Decode(v, val); err != nil {
+		return "", err
+	}
+
+	return string(v), nil
+}
+
+func encodeSecretValue(v string) []byte {
```
jademcgough: Should we be encrypting values by default, or should users just use vault if they want security?

desa: The only issue with encrypting things here is that whatever encryption key used will have to reside on the same instance as where this data is stored. So anyone that got access to the instance would have access to the data that is stored here.

So I think our recommendations will be, if you're ok with things being stored in plaintext, use the bolt store. If you need more security than that, use vault.

PR #1383, opened 2018-11-13
Add enable toggle feature to tasks page

### LINKED ISSUE #1327: Add Ability to Enable/Disable Tasks

Palakp41:
Co-authored-by: Palak Bhojani <patel.palak41@gmail.com>

Closes #1327 

Added a toggle button on the Tasks Page and created an API call function for updating the task status on toggle change

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/tasks/actions/v2/index.ts
+++ b/ui/src/tasks/actions/v2/index.ts
@@ -27,7 +27,7 @@ export type Action =
   | SetCurrentScript
   | SetCurrentTask
 
-type GetStateFunc = () => Promise<AppState>
+type GetStateFunc = () => AppState
 
 export enum ActionTypes {
   SetNewScript = 'SET_NEW_SCRIPT',
@@ -102,7 +102,7 @@ export const updateTaskStatus = (task: Task) => async (
   try {
     const {
       links: {tasks: url},
-    } = await getState()
+    } = getState()
     await updateTaskStatusAPI(url, task.id, task.status)
 
     dispatch(populateTasks())
@@ -119,7 +119,7 @@ export const deleteTask = (task: Task) => async (
   try {
     const {
       links: {tasks: url},
-    } = await getState()
+    } = getState()
 
     await deleteTaskAPI(url, task.id)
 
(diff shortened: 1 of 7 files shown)


ui/src/tasks/containers/TasksPage.tsx
```diff
@@ -102,6 +108,7 @@ const mstp = ({tasks: {tasks, searchTerm}}): ConnectedStateProps => {
 
 const mdtp: ConnectedDispatchProps = {
   populateTasks,
+  updateTaskStatus,
```
chnn: it might be easier to connect this directly to the components that need it, rather than plumbing it through

ui/src/tasks/api/v2/index.ts
```diff
@@ -30,6 +30,21 @@ export const updateTaskFlux = async (url, id, flux: string): Promise<Task> => {
   return data
 }
 
+export const updateTaskStatus = async (
+  url: string,
+  id: string,
+  status: string
+): Promise<Task> => {
+  const completeUrl = `${url}/${id}`
```
chnn: I think this should take the URL for the task directly, rather than hardcoding an assumption about how the task ID relates to the task index URL. The URL for a particular task is the `self` property of its `links`.

ui/src/tasks/api/v2/index.ts
```diff
@@ -30,6 +30,21 @@ export const updateTaskFlux = async (url, id, flux: string): Promise<Task> => {
   return data
 }
 
+export const updateTaskStatus = async (
```
chnn: What do you think about making this function more generic by taking a `taskUpdate: Partial<Task>` argument rather than a `status: string` argument? Then it would obviate the need for the above `updateTaskFlux` function too (which is also just a partial task update).

ui/src/tasks/actions/v2/index.ts
```diff
@@ -94,6 +95,23 @@ export const setSearchTerm = (searchTerm: string) => ({
...
 })
 
+export const updateTaskStatus = (task: Task) => async (
+  dispatch,
+  getState: GetStateFunc
+) => {
+  try {
+    const {
+      links: {tasks: url},
+    } = await getState()
```
chnn: no need to `await getState()`, it returns the state synchronously

ui/src/tasks/api/v2/index.ts
Palakp41: the response from the `/tasks` endpoint does not contain links for each task at the moment. We have communicated with goller for fix. will revisit at later date.

Issue #1390, opened 2018-11-13
/telegrafs GET should return an empty array, if there is no configs.


kelwang:
currently it returns 404.

PR #1394, opened 2018-11-14
fix 404 to an empty array

kelwang:
Closes https://github.com/influxdata/platform/issues/1390

Currently if no telegraf configs was in db, it will return 404 not found error.
As chronograf team points out, we should not return an empty array without error.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/bolt/telegraf.go
+++ b/bolt/telegraf.go
@@ -88,9 +88,7 @@ func (c *Client) findTelegrafConfigs(ctx context.Context, tx *bolt.Tx, filter pl
 		}
 	}
 	if len(m) == 0 {
-		return nil, 0, &platform.Error{
-			Code: platform.ENotFound,
-		}
+		return tcs, 0, nil
 	}
 	for _, item := range m {
 		tc, err := c.findTelegrafConfigByID(ctx, tx, item.ResourceID)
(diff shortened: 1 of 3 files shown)


kelwang:
cc @ischolten

bolt/telegraf.go
```diff
@@ -88,9 +88,7 @@ func (c *Client) findTelegrafConfigs(ctx context.Context, tx *bolt.Tx, filter pl
 		}
 	}
 	if len(m) == 0 {
-		return nil, 0, &platform.Error{
-			Code: platform.ENotFound,
-		}
+		return nil, 0, nil
```
desa: Why is this not an error?

inmem/telegraf.go
```diff
@@ -66,9 +66,7 @@ func (s *Service) findTelegrafConfigs(ctx context.Context, filter platform.UserR
 		}
 	}
 	if len(m) == 0 {
-		return nil, 0, &platform.Error{
-			Code: platform.ENotFound,
-		}
+		return nil, 0, nil
```
desa: Same?

bolt/telegraf.go
desa: Hmm. I think my issue is that if a user checks

```go
cfgs, n, err := findTelegrafConfigs(...)
if err != nil {
    ...
}

// use cfg
```

I expect to be able to use the array is if it is non-nil.

What do you think about just returning `tcs` instead of `nil`?

kelwang: 👍 like it, thanks

PR #1399, opened 2018-11-14
fix(logs/ui): add loading spinner

OfTheDelmer:
- Port over page spinner while fetching sources
- Increase logs table to reach bottom of page
- Import styles for logs message and only show copy button on hover
_What was the problem?_
There was no page spinner while the logs fetched source info
_What was the solution?_
- Port over the `PageSpinner` component because the `Spinner` component from `clockface` seems to interfere with the styling of the logs page.
- Add height to the logs table that was previously being utilized by the logs histogram

  - [ ] Rebased/mergeable
  - [ ] Tests pass


OfTheDelmer:
@alexpaxton with sources being loaded outside the logs page I no longer needed the `PageSpinner` and removed it. So this basically just fixes the table height.

PR #1402, opened 2018-11-14
Update service to include methods needed by querytest tool.

wolffcm:


wolffcm:
These are all changes needed to get the test generation to work for this issue:
https://github.com/influxdata/idpe/issues/1714

PR #1403, opened 2018-11-14
Added toggle on Task Header for option to show only active

### LINKED ISSUE #1369: Add the option to toggle only active tasks
as a task user
when i view the list of tasks
then i would like the option to only display the active tasks

Palakp41:

Closes #1369 

_What was the solution?_ Added a toggle on task header for option to show only active tasks.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/tasks/containers/TasksPage.tsx
+++ b/ui/src/tasks/containers/TasksPage.tsx
@@ -46,21 +46,28 @@ type Props = ConnectedDispatchProps & PassedInProps & ConnectedStateProps
 
 @ErrorHandling
 class TasksPage extends PureComponent<Props> {
-  constructor(props) {
+  constructor(props: Props) {
     super(props)
 
-    this.props.setSearchTerm('')
-    this.props.setShowInactive(true)
+    props.setSearchTerm('')
+    if (!props.showInactive) {
+      props.setShowInactive()
+    }
   }
 
   public render(): JSX.Element {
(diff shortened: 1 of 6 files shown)


ui/src/tasks/containers/TasksPage.tsx
```diff
@@ -92,18 +97,33 @@ class TasksPage extends PureComponent<Props> {
...
+
+      return searchTermFilter && activeFilter
+    })
 
     return matchingTasks
   }
+
+  private handleInactiveToggle = () => {
+    const {showInactive, setShowInactive} = this.props
+    setShowInactive(!showInactive)
```
ischolten: just a suggestion: since you're using redux to manage state here, how would you feel about having a `toggleShowInactive` action/reducer rather than a set. that way this checking would be done in the reducer where state is being managed and the component would only be worried about calling the action.

ui/src/tasks/reducers/v2/index.ts
```diff
@@ -34,6 +36,9 @@ export default (state: State = defaultState, action: Action): State => {
     case ActionTypes.SetSearchTerm:
       const {searchTerm} = action.payload
       return {...state, searchTerm}
+    case ActionTypes.SetShowInactive:
+      const {showInactive} = action.payload
+      return {...state, showInactive}
```
ischolten: for the suggestion above, this line would just be `{...state, showInactive: !state.showInactive}`

PR #1406, opened 2018-11-14
Add initial source management UI

chnn:
- Adds a simple UI for listing sources, adding a new source, deleting a source, and switching the active source
- Normalizes the schema used to store sources in Redux
- Makes Redux the state of truth for the currently active source, rather than the query param (the query param will stay synced via the `SetActiveSource` component)
- Fixes the inability to navigate backwards after switching to certain pages

screen shot 2018-11-14 at 2 37 07 pm

screen shot 2018-11-14 at 2 37 04 pm


ui/src/logs/containers/logs_page/LogsPage.tsx
```diff
@@ -120,7 +119,6 @@ class LogsPage extends Component<Props, State> {
 
   public async componentDidMount() {
     try {
-      await this.props.getSources()
```
chnn: This was removed since it already happens when the app first boots up (every route in the app is a child of the `GetSources` component, which fetches sources before rendering its children).

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -59,89 +62,76 @@ class SideNav extends PureComponent<Props> {
...
   }
+}
 
-  private get sourceParam(): string {
-    const {location, sources = []} = this.props
-
-    const {query} = location
-    const defaultSource = sources.find(s => s.default)
-    const id = query.sourceID || _.get(defaultSource, 'id', 0)
+const mapStateToProps = (state: AppState) => {
```
jaredscheib: 1. `mstp`?
1. explicit return type

ui/src/shared/containers/SetActiveSource.tsx
```diff
@@ -0,0 +1,82 @@
+import React, {PureComponent} from 'react'
```
jaredscheib: y'all still organizing these under comment categories?

ui/src/shared/containers/SetActiveSource.tsx
```diff
@@ -0,0 +1,82 @@
...
+  onSetActiveSource: typeof setActiveSource
+}
+
+type Props = ConnectStateProps & ConnectDispatchProps & PassedInProps
+
+class SetActiveSource extends PureComponent<Props> {
+  constructor(props: Props) {
+    super(props)
+
+    this.resolveActiveSource()
```
jaredscheib: could side effects from within the constructor be problematic?

ui/src/sources/actions/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  type: 'REMOVE_SOURCE',
+  payload: {sourceID},
+})
+
+export const readSources = () => async (
+  dispatch: Dispatch<Action>,
+  getState: GetState
+) => {
+  const sourcesLink = getState().links.sources
+  const sources = await readSourcesAJAX(sourcesLink)
```
jaredscheib: anything catching this and all the following?

ui/src/sources/actions/index.ts
```diff
@@ -0,0 +1,112 @@
...
+export const updateSource = (source: Source) => async (
+  dispatch: Dispatch<Action>
+) => {
+  const updatedSource = await updateSourceAJAX(source)
+
+  dispatch(setSource(updatedSource))
+}
+
+export const deleteSource = (sourceID: string) => async (
+  dispatch: any,
```
jaredscheib: `any` desired?

ui/src/sources/components/CreateSourceOverlay.tsx
```diff
@@ -0,0 +1,165 @@
...
+  private handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const draftSource = {
+      ...this.state.draftSource,
+      [e.target.name]: e.target.value,
+    }
+
+    this.setState({draftSource})
+  }
+
+  private handleChangeType = type => {
```
jaredscheib: type `type`? :)

ui/src/sources/components/DeleteSourceButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+        color={ComponentColor.Danger}
+        size={ComponentSize.ExtraSmall}
+        status={buttonStatus}
+        onClick={this.handleClick}
+      />
+    )
+  }
+
+  private handleClick = async () => {
+    this.setState({status: RemoteDataState.Loading})
```
jaredscheib: might you want the `setState` to succeed before calling `props.onClick()`?

ui/src/sources/components/SourcesList.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const sources = getSources(state)
+
+  return {sources}
+}
+
+const mdtp = {}
+
+export default connect<StateProps, DispatchProps, OwnProps>(
+  mstp,
+  mdtp
```
jaredscheib: can't be `null`?

ui/src/sources/components/SourcesListRow.tsx
```diff
@@ -0,0 +1,98 @@
...
+  )
+}
+
+const mstp = (state: AppState) => {
+  return {
+    activeSourceID: state.sources.activeSourceID,
+  }
+}
+
+const mdtp = dispatch => ({
```
jaredscheib: possibly type?

ui/src/sources/reducers/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    case 'SET_SOURCE': {
+      const {source} = action.payload
+
+      return {...state, sources: {...state.sources, [source.id]: source}}
+    }
+
+    case 'REMOVE_SOURCE': {
+      const sources = {...state.sources}
+
+      delete sources[action.payload.sourceID]
```
jaredscheib: a more functional style would be `.filter`, but this works just as well (and should be quicker). i approve!

ui/src/sources/reducers/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    case 'SET_ACTIVE_SOURCE': {
+      const {activeSourceID} = action.payload
+
+      return {...state, activeSourceID}
+    }
+
+    case 'SET_SOURCES': {
+      const sources = {...state.sources}
+
+      for (const source of action.payload.sources) {
```
jaredscheib: there's prob a lodash function for this.. escaping me atm

ui/src/sources/selectors/index.ts
```diff
@@ -0,0 +1,20 @@
+import {AppState, Source} from 'src/types/v2'
+
+export const getActiveSource = (state: AppState): Source | null => {
+  const {activeSourceID} = state.sources
+
+  if (!activeSourceID) {
+    return null
+  }
+
+  const source = state.sources.sources[activeSourceID]
```
jaredscheib: any chance this lookup fails at some level of nesting?

ui/src/sources/selectors/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  if (!source) {
+    return null
+  }
+
+  return source
+}
+
+export const getSources = (state: AppState): Source[] =>
+  Object.values(state.sources.sources)
```
jaredscheib: you loooove `Object.values` :)

ui/src/shared/containers/SetActiveSource.tsx
chnn: I actually moved it from the `componentDidMount` hook... the issue was that the children of this component (the entire rest of the app) was rendering before an active source had been resolved. Putting the active source resolution in the constructor instead fixed a few errors in the app where components were expecting a source to exist.

I can't think of any reason why it would create problems to be in the constructor, but it does seem a bit strange.

jaredscheib: Is it setting state in other components? I don't have full context/understanding but just recall things like this historically: https://reactjs.org/docs/react-component.html – `Avoid introducing any side-effects or subscriptions in the constructor. For those use cases, use componentDidMount() instead.`

ui/src/sources/actions/index.ts
chnn: for the most part they end up getting caught (or should be caught) by the components that dispatch the action. My thinking was that different components might want to handle errors from the same action differently

ui/src/sources/actions/index.ts
chnn: whoops, nice catch

ui/src/sources/reducers/index.ts
chnn: I opted for the janky-immutability since there's no built-in filter method on object types (it would have to be the lodash `filter` or something).

was it you that was sharing a link to [immer](https://github.com/mweststrate/immer) a while ago? super cool idea for reducers

ui/src/sources/selectors/index.ts
chnn: it may return undefined, but it should never fail with a `property of undefined` error

ui/src/sources/selectors/index.ts
chnn: haha totally a consequence of indexing collections by `id` in Redux

ui/src/shared/containers/SetActiveSource.tsx
```diff
@@ -0,0 +1,82 @@
...
+      onSetActiveSource(resolvedSourceID)
+    }
+
+    if (querySourceID !== resolvedSourceID) {
+      updateQueryParams({sourceID: resolvedSourceID})
+    }
+  }
+}
+
+const mstp = (state: AppState) => ({
```
AlirieGray: could type the return as `ConnectStateProps`

ui/src/shared/utils/queryParams.ts
```diff
@@ -0,0 +1,27 @@
+import {browserHistory} from 'react-router'
+import qs from 'qs'
+import {pickBy} from 'lodash'
+
+export const readQueryParams = (): {[key: string]: any} => {
+  return qs.parse(window.location.search, {ignoreQueryPrefix: true})
+}
+
+/*
+  Given an object of query parameter keys and values, updates any corresponding
```
AlirieGray: nice

ui/src/shared/containers/SetActiveSource.tsx
chnn: I think any type error should be caught by the generics that are passed to `connect`

ui/src/sources/actions/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  type: 'SET_ACTIVE_SOURCE'
+  payload: {
+    activeSourceID: string | null
+  }
+}
+
+export const setActiveSource = (
+  activeSourceID: string | null
+): SetActiveSourceAction => ({
+  type: 'SET_ACTIVE_SOURCE',
```
AlirieGray: could use an enum for the action types instead of a hard-coded string, that way you can use the enum in the reducer as well

ui/src/sources/actions/index.ts
```diff
@@ -0,0 +1,112 @@
...
+) => {
+  const source = getState().sources.sources[sourceID]
+
+  if (!source) {
+    throw new Error(`no source with ID "${sourceID}" exists`)
+  }
+
+  await deleteSourceAJAX(source)
+
+  dispatch(removeSource(sourceID))
```
AlirieGray: maybe check that the delete source call was successful, and only dispatch `removeSource` if it was successful?

ui/src/sources/actions/index.ts
chnn: I've been using strings instead of enums since it cuts down on the amount of boilerplate for creating actions while still offering comparable type safety in a reducer

ui/src/sources/actions/index.ts
chnn: good call 👍

chnn: actually I think the function would throw an error at that point, so the `dispatch` line would never be hit

ui/src/sources/apis/index.ts
```diff
@@ -1,6 +1,53 @@
 import AJAX from 'src/utils/ajax'
+import {Source} from 'src/types/v2'
 
-export const getSourceHealth = async (url: string) => {
+export const readSources = async (url): Promise<Source[]> => {
```
AlirieGray: why `readSources` instead of `getSources`?

ui/src/sources/components/CreateSourceOverlay.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    this.setState({creationStatus: RemoteDataState.Loading})
+
+    try {
+      await onCreateSource(draftSource)
+      onHide()
+    } catch (error) {
+      this.setState({creationStatus: RemoteDataState.Error})
+      onNotify(sourceCreationFailed(error.toString()))
+    }
```
AlirieGray: should the `creationStatus` be set to Done after the try catch block?

ui/src/sources/apis/index.ts
chnn: for consistency with the [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) vocabulary, and since `get` tends to have so many different meanings

ui/src/sources/components/CreateSourceOverlay.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  private handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
AlirieGray: nice

ui/src/sources/components/CreateSourceOverlay.tsx
chnn: since `onHide` gets called, the entire component is unmounted. I think setting the `creationStatus` would cause one of the those “setState on an unmounted component” errors.

ui/src/sources/components/SourcesList.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </IndexList.Header>
+        <IndexList.Body emptyState={<div />} columnCount={4}>
+          {rows}
+        </IndexList.Body>
+      </IndexList>
+    </div>
+  )
+}
+
+const mstp = (state: AppState) => {
```
AlirieGray: return type?

ui/src/sources/apis/index.ts
AlirieGray: got it, thanks for clarifying 👍

PR #1408, opened 2018-11-14
Update task creation request body in swagger

bthesorceror:
Reflect current state of the implementation

  - [ ] Rebased/mergeable
  - [ ] Tests pass


http/swagger.yml
```diff
@@ -3749,6 +3743,25 @@ components:
           readOnly: true
           description: Link to the full logs for a run.
           type: string
+    TaskCreateRequest:
+      properties:
+        organizationId:
+          description: The ID of the organization that owns this Task.
+          type: string
+        status:
+          description: The current status of the task. When updated to 'inactive', cancels all queued jobs of this task.
```
goller: Description should be something like:

> Starting state of the task. 'inactive' tasks are not run until they are updated to 'active'

http/swagger.yml
```diff
@@ -3749,6 +3743,25 @@ components:
...
+          default: active
+          type: string
+          enum:
+            - active
+            - inactive
+        flux:
+          description: The Flux script to run for this task.
+          type: string
+        owner:
```
goller: No need to specify owner on creation.  The owner would be the user of the token.

http/swagger.yml
```diff
@@ -2508,19 +2508,13 @@ paths:
       tags:
         - Tasks
       summary: Create a new task
-      parameters:
```
goller: all other endpoints are _supposed_ to use org; we need to keep this

http/swagger.yml
```diff
@@ -3749,6 +3743,25 @@ components:
           readOnly: true
           description: Link to the full logs for a run.
           type: string
+    TaskCreateRequest:
+      properties:
+        organizationId:
```
goller: with org in the params we wouldn't need it in the request.

goller: Also needs `name`

task/platform_adapter.go
```diff
@@ -80,7 +80,19 @@ func (p pAdapter) CreateTask(ctx context.Context, t *platform.Task) error {
 
 	// TODO(mr): decide whether we allow user to configure scheduleAfter. https://github.com/influxdata/platform/issues/595
 	scheduleAfter := time.Now().Unix()
-	req := backend.CreateTaskRequest{Org: t.Organization, User: t.Owner.ID, Script: t.Flux, ScheduleAfter: scheduleAfter}
+
+	if t.Status == "" {
+		t.Status = "active"
```
mark-rushakoff: Nit, but this should be backend.DefaultTaskStatus in case we ever decide to change the default.

https://godoc.org/github.com/influxdata/platform/task/backend#TaskStatus

PR #1411, opened 2018-11-15
feat(kit/grpc): add translation between gRPC status and platform.Error

goller:
Closes #

Add new kit library to standardize translate between gRPC status errors and platform.Errors.

_What was the problem?_
gRPC services have a "status" error that is returned.  We need to translate them to our standard error format.

_What was the solution?_
Translates between gRPC status errors and platform.Error

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/kit/grpc/errors.go
+++ b/kit/grpc/errors.go
@@ -1,0 +1,54 @@
+package grpc
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/influxdata/platform"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ToStatus converts a platform.Error to a gRPC status message.
+func ToStatus(err *platform.Error) (*status.Status, error) {
+	if err == nil {
+		return status.New(codes.OK, ""), nil
+	}
+
+	c := codes.Unknown
+	switch err.Code {
+	case platform.EInternal:
+		c = codes.Internal
+	case platform.ENotFound:
+		c = codes.NotFound
+	case platform.EConflict, platform.EInvalid, platform.EEmptyValue: // not really sure how to express this as gRPC only has Invalid
+		c = codes.InvalidArgument
+	case platform.EUnavailable:
+		c = codes.Unavailable
+	}
+
+	buf, jerr := json.Marshal(err)
+	if jerr != nil {
+		return nil, jerr
+	}
+
+	return status.New(c, string(buf)), nil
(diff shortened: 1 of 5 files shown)


kit/grpc/errors.go
```diff
@@ -0,0 +1,52 @@
...
+func FromStatus(s *status.Status) *platform.Error {
+	if s == nil || s.Code() == codes.OK {
+		return nil
+	}
+
+	msg := s.Message()
+	var perr platform.Error
+	if err := json.Unmarshal([]byte(msg), &perr); err != nil {
+		return &platform.Error{
+			Err:  fmt.Errorf(msg),
```
mark-rushakoff: Should be `errors.New`. We don't want to apply fmt to any accidental % verbs in msg.

kit/grpc/errors.go
```diff
@@ -0,0 +1,52 @@
...
+
+	c := codes.Unknown
+	switch err.Code {
+	case platform.EInternal:
+		c = codes.Internal
+	case platform.ENotFound:
+		c = codes.NotFound
+	case platform.EConflict, platform.EInvalid, platform.EEmptyValue: // not really sure how to express this as gRPC only has Invalid
+		c = codes.InvalidArgument
+	}
```
mark-rushakoff: Looks like you'd want to handle `case platform.EUnavailable: c = codes.Unavailable` in this switch.

kit/grpc/errors.go
```diff
@@ -0,0 +1,52 @@
...
+
+	buf, jerr := json.Marshal(err)
+	if jerr != nil {
+		return nil, jerr
+	}
+
+	return status.New(c, string(buf)), nil
```
mark-rushakoff: I don't know enough about gRPC statuses, but putting JSON in there seems a little weird. I wonder if we should look into adding a MarshalText method to platform.Error.

goller: @mark-rushakoff marshaling JSON is just a convenience so that unmarshaling the error across the gRPC link is easy.

We could do some sort of MarshalText, but, we'd need the corresponding parser for the unmarshal.  Any ideas for an easy marshaling/unmarshaling scheme?

mark-rushakoff: Now that I've looked more closely at platform.Error, I'm fine with MarshalJSON since it's straightforward on our end. I was mainly concerned about what is conventional for gRPC status.

PR #1419, opened 2018-11-15
Create a dropdown filter to filter tasks through Organizations

### LINKED ISSUE #1372: Add the ability to filter tasks by organization via dropdown
as a tasks user
when i view the tasks list page
then i would like the ability to filter tasks by the organization associated with it

Palakp41:
Co-authored-by: Palak Bhojani <palak@influxdata.com>

Closes #1372 

_What was the solution?_Created a dropdown on Task Header to show all organizations and an ability to filter tasks using the dropdown

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/tasks/components/TasksHeader.tsx
+++ b/ui/src/tasks/components/TasksHeader.tsx
@@ -1,5 +1,8 @@
+// Libraries
 import React, {PureComponent} from 'react'
 import {Page} from 'src/pageLayout'
+
+// Components
 import {
   Button,
   ComponentColor,
@@ -8,6 +11,7 @@ import {
   SlideToggle,
 } from 'src/clockface'
 import SearchWidget from 'src/shared/components/search_widget/SearchWidget'
+import TaskOrgDropdown from 'src/tasks/components/TasksOrgDropdown'
 
 interface Props {
   onCreateTask: () => void
@@ -41,6 +45,7 @@ export default class TasksHeader extends PureComponent<Props> {
             placeholderText="Filter tasks by name..."
             onSearch={setSearchTerm}
           />
+          <TaskOrgDropdown />
           <Button
             color={ComponentColor.Primary}
             onClick={onCreateTask}
--- a/ui/src/tasks/components/TasksOrgDropdown.tsx
+++ b/ui/src/tasks/components/TasksOrgDropdown.tsx
@@ -1,0 +1,76 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {connect} from 'react-redux'
+
+// Components
+import {Dropdown} from 'src/clockface'
+
+// Actions
+import {setDropdownOrgID as setDropdownOrgIDAction} from 'src/tasks/actions/v2'
+
+// Constants
+import {defaultAllOrgs} from 'src/tasks/constants'
+
+// Types
+import {Organization} from 'src/types/v2'
+
+interface ConnectedDispatchProps {
+  setDropdownOrgID: typeof setDropdownOrgIDAction
+}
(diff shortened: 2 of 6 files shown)


ui/src/tasks/components/TasksHeader.tsx
```diff
@@ -41,6 +49,11 @@ export default class TasksHeader extends PureComponent<Props> {
             placeholderText="Filter tasks by name..."
             onSearch={setSearchTerm}
           />
+          <TaskOrgDropdown
+            orgs={orgs}
```
chnn: Rather than connect these three props to the `TaskPage` and then pass them to the `TaskHeader` and then to the `TaskOrgDropdown`, could they be connected directly to the `TaskOrgDropdown` instead?

I would prefer this, since it would help keep the other components smaller and more focused (although the `TaskPage` would still need to be connected to the `dropdownOrgID`).

ui/src/tasks/containers/TasksPage.tsx
```diff
@@ -113,18 +125,28 @@ class TasksPage extends PureComponent<Props> {
       if (!showInactive) {
         activeFilter = t.status === TaskStatus.Active
       }
-
-      return searchTermFilter && activeFilter
+      let orgIDFilter = true
+      if (dropdownOrgID) {
+        orgIDFilter = t.organizationId === dropdownOrgID
```
chnn: how does this work if the `dropdownOrgID` is `'All Organizations'`?

ui/src/tasks/components/TasksOrgDropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+          </Dropdown.Item>
+        ))}
+      </Dropdown>
+    )
+  }
+
+  private get selectedID(): string {
+    const {dropdownOrgID} = this.props
+    if (!dropdownOrgID) {
+      return this.orgsArray[0].id
```
jaredscheib: is it ever possible that there will be no orgs and this lookup could fail with property undefined?

ui/src/tasks/components/TasksOrgDropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return dropdownOrgID
+  }
+
+  private get orgsArray(): Array<Partial<Organization>> {
```
jaredscheib: what do you think about a name that reflects what it _is_, rather than the data structure? such as `private get allOrgs`?

ui/src/tasks/constants/index.ts
```diff
@@ -0,0 +1,13 @@
+import {Organization} from 'src/types/v2'
+
+export const allOrgsElement: Organization = {
```
jaredscheib: same question here about a name that reflects what it is, rather than the underlying implementation. e.g. `Element` --> `defaultAllOrgs`?

ui/src/tasks/components/TasksOrgDropdown.tsx
ebb-tide: it's not possible that this.orgsArray is empty, because we are adding allOrgsElement by default.

ui/src/tasks/containers/TasksPage.tsx
Palakp41: Fixed it thanks!

ui/src/tasks/components/TasksHeader.tsx
Palakp41: Connected TaskOrgDropdown to redux, thanks

PR #1420, opened 2018-11-15
Move org name location on org page

### LINKED ISSUE #1376: Update organization page to include the name of the organization in the top left
as an org manager
when i view the detail page for an org
then i would like to see the name of the org in the top left corner of the page

chnn:
Closes #1376

After:

screen shot 2018-11-15 at 12 08 23 pm

Also fixes an issue with broken typechecking wrt the `ProfilePage`.


ui/src/organizations/components/Dashboards.tsx
```diff
@@ -45,7 +45,7 @@ export default class Dashboards extends PureComponent<Props, State> {
             onChange={this.handleFilterChange}
             placeholder="Filter Dashboards..."
           />
-        </ProfilePage.Header>
+        </ProfilePageHeader>
```
OfTheDelmer: Interesting, what was happening to the type check?

ui/src/organizations/containers/OrganizationView.tsx
```diff
@@ -58,7 +59,7 @@ class OrganizationView extends PureComponent<Props> {
       <Page>
         <Page.Header fullWidth={false}>
           <Page.Header.Left>
-            <Page.Title title="Organization" />
+            <Page.Title title={org.name ? org.name : 'Organization'} />
```
OfTheDelmer: Nice!

ui/src/organizations/components/Dashboards.tsx
chnn: basically doing `withRouter(ProfilePage)` (the default export) would export a component type like `ComponentClass<any, any>`, so that no meaningful typechecking would happen when invoking the component.

I did `withRouter<OwnProps>(ProfilePage)` instead, which exports something of type `ComponentClass<OwnProps>`, so the prop typechecking was back. But then `static Header` doesn't exist on this more generic type, so I removed the fancy namespace invocation of the `ProfilePageHeader` and related components.

Issue #1305, opened 2018-11-05
Add the ability to configure flux task options from the create task page


bthesorceror:
Given I am a platform user
When I go to the tasks creation page
And I supply the tasks options through the GUI
And Click create
Then I see that a task was created with the appropriate task options prepended to the flux script

bthesorceror:
Given I am a platform user
When I go to the tasks creation page
And I supply the tasks options through the GUI
And Click create
Then I see that a task was created with the appropriate task options prepended to the flux script

Issue #1306, opened 2018-11-05
Update task from update tasks page


bthesorceror:

ebb-tide:
blocked on https://github.com/influxdata/flux/issues/286

PR #1426, opened 2018-11-15
Tasks/configure task options from flux page

AlirieGray:
Closes: #1305 
Closes: #1469

_What was the problem?_
Previously, the only way to create a task was to type in the options into the Flux script field, like so: 

screen shot 2018-11-15 at 2 06 50 pm

_What was the solution?_
This PR introduces a UI for updating task options, shown here: 

screen shot 2018-11-15 at 1 57 40 pm

Once the user has selected the options they want, the options are converted into a string and attached to the flux script. When they open the "update script" page, they will then see the full Flux script with the options prepended, as in the first picture.

NOTE THAT: the "retry attempts" field does not save to the new task, and that updating options on an already created script with the options UI is not supported in this PR. That will be addressed in: https://github.com/influxdata/platform/issues/1306. Also, this UI does not contain a field to set the task as active/inactive, that will also be addressed in another PR.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/components/TaskForm.tsx
+++ b/ui/src/tasks/components/TaskForm.tsx
@@ -1,33 +1,170 @@
 import _ from 'lodash'
-import React, {PureComponent} from 'react'
+import React, {PureComponent, ChangeEvent} from 'react'
 
-import {Form, Columns} from 'src/clockface'
+import {
+  ComponentSpacer,
+  Form,
+  Columns,
+  Input,
+  Radio,
+  ButtonShape,
+} from 'src/clockface'
+import TaskOptionsOrgDropdown from 'src/tasks/components/TasksOptionsOrgDropdown'
 import FluxEditor from 'src/shared/components/FluxEditor'
+import TaskScheduleFormField from 'src/tasks/components/TaskScheduleFormField'
+
+import {TaskOptions, TaskSchedule} from 'src/utils/taskOptionsToFluxScript'
+
+import {Organization} from 'src/types/v2'
+import {Alignment, Direction} from 'src/clockface/types'
 
 interface Props {
   script: string
-  onChange: (script) => void
+  orgs: Organization[]
+  taskOptions: TaskOptions
+  onChangeScheduleType: (schedule: TaskSchedule) => void
+  onChangeScript: (script: string) => void
+  onChangeInput: (e: ChangeEvent<HTMLInputElement>) => void
+  onChangeScheduleUnit: (unit: string, scheduleType: string) => void
+  onChangeTaskOrgID: (orgID: string) => void
 }
 
-export default class TaskForm extends PureComponent<Props> {
+interface State {
+  selectedIntervalUnit: string
+  cron: string
+  selectedDelayUnit: string
+  delayTime: string
+  retryAttempts: string
+  schedule: TaskSchedule
+}
+
+export default class TaskForm extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      cron: '',
+      selectedIntervalUnit: 'd',
+      delayTime: '20',
+      selectedDelayUnit: 'm',
+      retryAttempts: '1',
+      schedule: props.taskOptions.taskScheduleType,
+    }
+  }
+
   public render() {
+    const {
+      script,
+      onChangeInput,
+      onChangeScript,
+      onChangeTaskOrgID,
+      taskOptions: {
+        name,
+        taskScheduleType,
+        intervalTime,
+        intervalUnit,
+        delayTime,
+        delayUnit,
+        cron,
+        orgID,
+      },
+      orgs,
+    } = this.props
+
     return (
-      <Form style={{height: '90%'}}>
-        <Form.Element
-          label="Script"
-          colsXS={Columns.Six}
-          offsetXS={Columns.Three}
-          errorMessage={''}
-        >
+      <div className="task-page">
+        <div className="task-page--options">
+          <Form>
+            <Form.Element label="Name" colsXS={Columns.Twelve}>
+              <Input
+                name="name"
+                placeholder="Name your task"
+                onChange={onChangeInput}
+                value={name}
+              />
+            </Form.Element>
+            <Form.Element label="Owner">
+              <TaskOptionsOrgDropdown
+                orgs={orgs}
+                selectedOrgID={orgID}
+                onChangeTaskOrgID={onChangeTaskOrgID}
+              />
+            </Form.Element>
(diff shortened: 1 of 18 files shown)


ui/src/utils/taskOptionsToFluxScript.ts
```diff
@@ -0,0 +1,28 @@
...
+  cron = 'CRON',
+}
+
+export const taskOptionsToFluxScript = (options: TaskOptions): string => {
+  let fluxScript = `option task = { \n  name: "${options.name}",\n`
+
+  if (options.taskScheduleType === TaskSchedule.interval) {
+    fluxScript = `${fluxScript}  every: ${options.interval},`
+  } else if (options.taskScheduleType === TaskSchedule.cron) {
+    const formattedCron = options.cron
```
bthesorceror: I don't think this is going to work for something like "* * 15 40", I also think that it would be better to lean toward validation over manipulating their input.

ui/src/tasks/components/TaskForm.tsx
```diff
@@ -1,33 +1,246 @@
...
+    const {onChangeTaskName} = this.props
+
+    onChangeTaskName(e.target.value)
+  }
+
+  private handleChangeScheduleType = (schedule: TaskSchedule): void => {
+    this.props.onChangeScheduleType(schedule)
+  }
+
+  private handleChangeCron = (cron: string): void => {
```
121watts: Many of these `handleChange` functions can be consolidated into one function.

ui/src/tasks/components/TaskForm.tsx
```diff
@@ -1,33 +1,246 @@
...
+
+  private get ownerDropdownItems(): JSX.Element[] {
+    const {orgs} = this.props
+
+    return orgs.map(org => {
+      return (
+        <Dropdown.Item id={org.id} key={org.id} value={org.id}>
+          {org.name}
+        </Dropdown.Item>
+      )
```
121watts: This could be its own component

ui/src/tasks/components/TaskForm.tsx
```diff
@@ -1,33 +1,246 @@
...
+      return (
+        <Dropdown.Item
+          value={timeUnitStringToAbbreviation[unit]}
+          key={`delay${unit}`}
+          id={timeUnitStringToAbbreviation[unit]}
+        >
+          {unit}
+        </Dropdown.Item>
+      )
+    })
```
121watts: This could be its own component

ui/src/tasks/components/TaskForm.tsx
```diff
@@ -1,33 +1,246 @@
...
+      return (
+        <Dropdown.Item
+          value={timeUnitStringToAbbreviation[unit]}
+          key={`interval${unit}`}
+          id={timeUnitStringToAbbreviation[unit]}
+        >
+          {unit}
+        </Dropdown.Item>
+      )
+    })
```
121watts: This could be its own component

ui/src/tasks/components/TaskForm.tsx
```diff
@@ -1,33 +1,246 @@
...
+            </Form.Element>
+          </Form>
+        </div>
+        <div className="task-page--editor">
           <FluxEditor
-            script={this.props.script}
-            onChangeScript={this.props.onChange}
+            script={script}
+            onChangeScript={this.props.onChangeScript}
             visibility={'visible'}
```
121watts: string props don't need the curlys

ui/src/tasks/components/TaskScheduleFormField.tsx
```diff
@@ -0,0 +1,88 @@
...
+  private onChangeDelayTime = (e: ChangeEvent<HTMLInputElement>) => {
+    this.props.onChangeDelayTime(e.target.value)
+  }
+
+  private onChangeIntervalTime = (e: ChangeEvent<HTMLInputElement>) => {
+    this.props.onChangeIntervalTime(e.target.value)
+  }
+
+  private onChangeCron = (e: ChangeEvent<HTMLInputElement>) => {
+    this.props.onChangeCron(e.target.value)
```
121watts: These can be pulled up into the parent into a single `handleChange`

ui/src/tasks/components/TaskScheduleFormField.tsx
```diff
@@ -0,0 +1,88 @@
...
+      )
+    }
+    return (
+      <>
+        <Input
+          placeholder="02***"
+          onChange={this.onChangeCron}
+          value={this.props.cron}
+        />
+      </>
```
121watts: Don't think you need a fragment here

ui/src/tasks/components/TaskScheduleFormField.tsx
```diff
@@ -0,0 +1,88 @@
...
+              onChange={this.onChangeDelayTime}
+            />
+            <Dropdown
+              selectedID={this.props.selectedDelayUnit}
+              onChange={this.props.onChangeDelayUnit}
+            >
+              {this.props.delayDropdownItems}
+            </Dropdown>
+          </ComponentSpacer>
+        </>
```
121watts: Don't think you need a fragment here.

ui/src/tasks/reducers/v2/index.ts
```diff
@@ -18,10 +28,40 @@ const defaultState: State = {
   searchTerm: '',
   showInactive: true,
   dropdownOrgID: null,
+  interval: '1d',
+  taskOptions: defaultTaskOptions,
 }
 
 export default (state: State = defaultState, action: Action): State => {
   switch (action.type) {
+    case ActionTypes.SetTaskName:
```
121watts: These actions could be one action like `SetTaskKey` whose action creator returns something like `{key: 'name', value: 'foo'}` and reducer could return something like {...state.taskOptions, [key]: value}`

ui/src/utils/taskOptionsToFluxScript.ts
AlirieGray: ok, I'll just send the raw string

ui/src/tasks/components/TaskScheduleFormField.tsx
AlirieGray: I think we do in this case, bc the ComponentSpacers are siblings

ui/src/tasks/components/TaskForm.tsx
AlirieGray: I made a component called `TaskScheduleUnitDropdown` that does both interval and delay units

PR #1449, opened 2018-11-16
feat(ui/onboarding): Add step for choosing csv, streaming, or line protocol

### LINKED ISSUE #1422: Select different data source types
as a new user for platform
with admin permissions
when i go through onboarding for the first time
then i would like the ability to select the data source id like to configure (csv, streaming, line protocol)
and if streaming is selected, allow multiple streaming sources to be selected.

ischolten:
Closes #1422 

Add step for choosing csv, streaming or line protocol. If you choose streaming,  you can add the different streaming data source options.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/onboarding/components/ConfigureDataSourceSwitcher.tsx
+++ b/ui/src/onboarding/components/ConfigureDataSourceSwitcher.tsx
@@ -1,0 +1,20 @@
+// Libraries
+import React, {PureComponent} from 'react'
+
+// Components
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+export interface Props {
+  dataSource: string
+}
+
+@ErrorHandling
+class ConfigureDataSourceSwitcher extends PureComponent<Props> {
+  public render() {
+    const {dataSource} = this.props
+
+    return <div>{dataSource}</div>
+  }
+}
+
+export default ConfigureDataSourceSwitcher
--- a/ui/src/onboarding/components/SelectDataSourceStep.tsx
+++ b/ui/src/onboarding/components/SelectDataSourceStep.tsx
@@ -9,24 +9,39 @@ import {
   ComponentSize,
   ComponentStatus,
 } from 'src/clockface'
-import CardSelectCard from 'src/clockface/components/card_select/CardSelectCard'
-import GridSizer from 'src/clockface/components/grid_sizer/GridSizer'
+import DataSourceSelector from 'src/onboarding/components/DataSourceSelector'
+import StreamingDataSourceSelector from 'src/onboarding/components/StreamingDataSourcesSelector'
 
 // Types
 import {OnboardingStepProps} from 'src/onboarding/containers/OnboardingWizard'
 import {DataSource, ConfigurationState} from 'src/types/v2/dataSources'
-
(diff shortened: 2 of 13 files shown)


ui/src/onboarding/components/SelectDataSourceStep.tsx
```diff
@@ -74,25 +76,79 @@ class SelectDataSourceStep extends PureComponent<Props> {
...
+      return
+    }
+
+    this.props.onDecrementCurrentStepIndex()
+  }
+
+  private handleSelectDataSource = (dataSource: string) => {
+    switch (dataSource) {
+      case 'Streaming':
+        this.setState({streaming: StreamingOptions.Selected})
```
jaredscheib: any possible race conditions to be careful about if `setState()` here ends up setting state after `this.props.onSetDataSources()`? this same question applies to all instances above and below this

ui/src/onboarding/components/StreamingDataSourcesSelector.tsx
```diff
@@ -0,0 +1,61 @@
...
+  dataSources: DataSource[]
+  onToggleDataSource: (dataSource: string, isSelected: boolean) => void
+}
+
+const STREAMING_DATA_SOURCES_OPTIONS = [
+  'CPU',
+  'Docker',
+  'InfluxDB',
+  'Kubernetes',
+  'NGNIX',
```
jaredscheib: typo --> `NGINX`

ui/src/onboarding/components/ConfigureDataSourceSwitcher.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+@ErrorHandling
+class ConfigureDataSourceSwitcher extends PureComponent<Props> {
+  public render() {
+    const {dataSource} = this.props
+
+    switch (dataSource) {
+      case 'System':
+        return <div> system</div>
```
jaredscheib: nit: there's an extra space before `system` here :)

ui/src/onboarding/reducers/steps.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface OnboardingStepsState {
+  currentStepIndex: number
+  stepStatuses: StepStatus[]
+  setupParams: SetupParams
+}
+
+const INITIAL_STATE: OnboardingStepsState = {
+  currentStepIndex: 0,
+  stepStatuses: [
```
jaredscheib: this works, though you _could_ generate an Array of length populated with this value for each, based on the total number of desired states

ischolten: can do

ui/src/onboarding/components/ConfigureDataSourceSwitcher.tsx
ischolten: good catch. i think im just going to return that default since its essentially the same thing for now

ui/src/onboarding/components/SelectDataSourceStep.tsx
ischolten: it doesnt really matter here which happens before the other.

PR #1458, opened 2018-11-16
WIP: Add vault implementation of secret service

### LINKED ISSUE #1365: Add Vault implementation of secret service.

desa:
~**Note**: I believe we are hitting https://github.com/golang/go/issues/28489 and/or https://github.com/golang/go/issues/26208 with this PR. Essentially, the import of docker depends on an old logrus which collides with the dependency of vaults newer logrus. I've tried removing the chronograf dependency of logrus, but that still did not seem to fix the issue--likely for the reason I mention above. See https://github.com/golang/go/issues/26208#issuecomment-411955266 for more details.~

Closes #1365

This PR introduces a vault implementation of the secret service. It is intended to be used by those who would prefer to securely store their secrets in vault (as opposed to in plaintext in bolt).

  - [x] Rebased/mergeable
  - [x] Tests pass


desa:
@jademcgough this should be ready for another review

PR #1491, opened 2018-11-19
Add .DS_Store to .gitignore file

AlirieGray:
Add .DS_Store to .gitignore file

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ vendor
 .vscode
 .vs
 .tern-project
+.DS_Store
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@ vendor
 .vscode
 .vs
 .tern-project
+.DS_Store
```
OfTheDelmer: Ahh nice to see `.vscode` is also in there already

PR #1493, opened 2018-11-19
Add Flux functions list to VEO

### LINKED ISSUE #1424: Migrate 1.7 flux explorer to platform
- [x] migrate script editor component to platform (ignore script wizard for now)
- [x] add the ability to submit a query
- [x] add the ability to view a line graph of the query results (no line graph config controls for now, no viz options)
- [x] add the ability to download the raw results of the query as a csv
- [x] add the ability to view the raw results of a query
- [x] add the functions list (ignore the explore tool for now)

chnn:
Closes #1424 

Ports the `FluxFunctionsToolbar` and related components from Chronograf 1.7:

screen shot 2018-11-19 at 12 32 44 pm

Also does some miscellaneous cleanup, which is described by the individual commits.


ui/src/shared/actions/v2/timeMachines.ts
```diff
@@ -40,6 +42,18 @@ export const setActiveTimeMachine = (
...
 
+interface SetActiveTabAction {
+  type: 'SET_ACTIVE_TAB'
+  payload: {activeTab: TimeMachineTab}
+}
+
+export const setActiveTab = (
+  activeTab: TimeMachineTab
+): SetActiveTabAction => ({
+  type: 'SET_ACTIVE_TAB',
```
Palakp41: I don't know if we have converged upon a solution for this but could we use enums for action types?

ui/src/shared/components/TimeMachine.tsx
```diff
@@ -18,65 +18,52 @@ import {getActiveTimeMachine} from 'src/shared/selectors/timeMachines'
...
-interface OwnProps {
-  activeTab: TimeMachineTab
-}
-
-type Props = OwnProps & StateProps
-
-class TimeMachine extends PureComponent<Props> {
-  public render() {
-    const {activeTab, queryLink, queries} = this.props
+const TimeMachine: SFC<StateProps> = props => {
```
Palakp41: Would you consider renaming Stateprops to props?

ui/src/shared/components/TimeMachine.tsx
```diff
@@ -18,65 +18,52 @@ import {getActiveTimeMachine} from 'src/shared/selectors/timeMachines'
...
+              </div>
+            </>
+          )
+        }}
+      </TimeSeries>
+    </div>
+  )
 }
 
 const mstp = (state: AppState) => {
```
Palakp41: Would you add stateprops as the return type of mstp?

ui/src/shared/components/TimeMachineBottom.tsx
```diff
@@ -31,4 +40,13 @@ const TimeMachineBottom: SFC<Props> = props => {
   return null
 }
 
-export default TimeMachineBottom
+const mstp = (state: AppState) => {
```
Palakp41: Would you add stateprops to the return type of mstp?

ui/src/shared/components/TimeMachineQueryEditor.tsx
```diff
@@ -26,29 +37,60 @@ interface DispatchProps {
...
+      queryStatus,
+      draftScript,
+      onSetDraftScript,
+      onSubmitScript,
+    } = this.props
 
-    return (
-      <div className="time-machine-query-editor">
-        <div className="time-machine-query-editor--controls">
+    const buttonStatus =
```
Palakp41: Would it be better to have a private getter for the buttonStatus?

ui/src/shared/components/flux_functions_toolbar/FluxFunctionsToolbar.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+interface State {
+  searchTerm: string
+}
+
+class FluxFunctionsToolbar extends PureComponent<Props, State> {
+  public state: State = {searchTerm: ''}
+
+  public render() {
+    const {searchTerm} = this.state
```
Palakp41: What do you think about putting this in redux state?

ui/src/shared/components/flux_functions_toolbar/TooltipLink.tsx
```diff
@@ -0,0 +1,23 @@
+import React, {PureComponent} from 'react'
+
+interface Props {
+  link: string
+}
+
+class TooltipLink extends PureComponent<Props> {
```
Palakp41: Would you consider transforming this into a SFC?

ui/src/shared/reducers/v2/timeMachines.ts
```diff
@@ -156,6 +159,11 @@ const timeMachineReducer = (
       return {...state, isViewingRawData}
     }
 
+    case 'SET_ACTIVE_TAB': {
```
Palakp41: Would you consider using enums here?

ui/src/shared/actions/v2/timeMachines.ts
chnn: I've seen this come up in code review a few times before, so I posted about it in the #chronograf-code channel. I'll update this and the other action types if we can come to a consensus

ui/src/shared/components/TimeMachine.tsx
chnn: I thought it was a bit clearer this way that those props are coming from Redux. Do you feel strongly about this? If so I'll update it

ui/src/shared/components/TimeMachine.tsx
chnn: `mstp` will get type checked by the `connect` function, which accepts `StateProps` as props.

do you think adding the type here helps or hinders readability?

ui/src/shared/components/flux_functions_toolbar/TooltipLink.tsx
chnn: can do 👍

ui/src/shared/components/flux_functions_toolbar/FluxFunctionsToolbar.tsx
chnn: Since it's only used here and once elsewhere I thought it would be simpler to keep it in local state

ui/src/shared/components/FluxEditor.tsx
```diff
@@ -185,6 +186,11 @@ class FluxEditor extends PureComponent<Props, State> {
 
   private handleKeyUp = (__, e: KeyboardEvent) => {
     const {ctrlKey, metaKey, key} = e
+    const {onSubmitScript} = this.props
+
+    if (ctrlKey && key === 'Enter' && onSubmitScript) {
```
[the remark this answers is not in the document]
chnn: I was copying what we had before in the Flux editor. I agree that command + enter would be nice, but probably in addition to ctrl + enter so that windows and linux users can use the shortcut too. I'll update it.

ui/src/shared/components/TimeMachine.tsx
```diff
@@ -18,65 +18,52 @@ import {getActiveTimeMachine} from 'src/shared/selectors/timeMachines'
...
-                render: () => <TimeMachineBottom activeTab={activeTab} />,
-                headerOrientation: HANDLE_HORIZONTAL,
-                size: 0.67,
-              },
-            ]
+  return (
+    <div className="time-machine">
+      <TimeSeries link={queryLink} queries={queries}>
+        {queriesState => {
+          const divisions = [
```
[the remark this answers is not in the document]
chnn: yep totally

ui/src/shared/components/FluxEditor.tsx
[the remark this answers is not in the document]
chnn: I attempted to add the command key too, but am having trouble detecting key up events when the meta key is pressed. I'm going to punt for now.

ui/src/shared/components/TimeMachine.tsx
ebb-tide: nope.

ui/src/shared/actions/v2/timeMachines.ts
Palakp41: We just use enums for action types in tasks page so I thought it would be nice to have it consistent :)

PR #1494, opened 2018-11-19
Fix issue#1467

### LINKED ISSUE #1467: Can't POST to /api/v2/query with a session cookie
I have a valid session, and I'm attempting to run a query via:

```
curl -v -X POST 'http://localhost:9999/api/v2/query' \
  -H "Cookie: session=$INFLUX_SESSION" \
  -d '{"query":"from(bucket: \"telegraf\")\n  |> range(start: -1m)"}'
```
I receive the error
```
X-Influx-Error: authorization not found
```

If I use an invalid session, I receive a different error:
```
X-Influx-Error: unauthorized
```

However, using a `Authorization` header works fine:
```
curl -v -X POST 'http://localhost:9999/api/v2/query' \
  -H "Authorization: Token $INFLUX_TOKEN" \
  -d '{"query":"from(bucket: \"telegraf\")\n  |> range(start: -1m)"}'
```

I can also query other endpoints just fine using a session (e.g. `/api/v2/orgs`).

desa:
Closes #1467 

This PR introduces a mechanism for turning a user resource mapping into a set of permissions that can be used to authorize requests. Additionally, it removes the write endpoints dependence on the authorization service and uses the authorizer middleware instead.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/user_resource_mapping.go
+++ b/user_resource_mapping.go
@@ -3,6 +3,7 @@ package platform
 import (
 	"context"
 	"errors"
+	"fmt"
 )
 
 type UserType string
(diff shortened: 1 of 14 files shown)


user_resource_mapping.go
```diff
@@ -66,3 +67,33 @@ type UserResourceMappingFilter struct {
 	UserID       ID
 	UserType     UserType
 }
+
+var ownerActions = []action{WriteAction, CreateAction, DeleteAction}
+var memberActions = []action{ReadAction}
+
+// ToPermission converts a user resource mapping into a set of permissions.
+func (m *UserResourceMapping) ToPermissions() []Permission {
```
kelwang: maybe use a map [resource]permission, so it's o(1) for search

PR #1534, opened 2018-11-20
Switch Tasks UI updates to use generated client

bthesorceror:
- [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/tasks/api/v2/index.ts
+++ b/ui/src/tasks/api/v2/index.ts
@@ -10,38 +10,33 @@ const getBasePath = () => {
   return `${protocol}//${host}:${port}/api/v2`
 }
 
+const createTaskAPI = () => {
+  return new TasksApi({basePath: getBasePath()})
+}
+
 export const submitNewTask = async (
   organizationId: string,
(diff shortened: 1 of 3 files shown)


ui/src/tasks/api/v2/index.ts
```diff
@@ -10,38 +10,33 @@ const getBasePath = () => {
...
-export const updateTaskFlux = async (url, id, flux: string): Promise<Task> => {
-  const completeUrl = `${url}/${id}`
-  const request = {
-    flux,
-  }
-
-  const {data} = await AJAX({url: completeUrl, data: request, method: 'PATCH'})
+export const updateTaskFlux = async (id, flux: string): Promise<Task> => {
+  const api = createTaskAPI()
+  const {data} = await api.tasksTaskIDPatch(id, {flux})
```
AlirieGray: could this be called `taskIDPatch` instead of `tasksTaskIDPatch`?

bthesorceror: that method name is from the generator

PR #1535, opened 2018-11-20
Enable viewing a bar chart

### LINKED ISSUE #1527: should be able to view a bar chart
as a query builder
when i add a cell to my dashboard
then i would like the ability to visualize my data as a column bar chart
then i would like the ability to name my y1 axis
then i would like the ability to choose the color of my bar chart
then i would like to specify a min and max value for the y1 axis (or auto)
then i would like to specify a prefix or postfix for the y axis
then i would like to specify the scale as linear or logarithmic

chnn:
Closes #1527 

- Consolidates the `LineGraph` / `Stacked` / `StepChart` components into a single `DygraphContainer` component, since they differ only by a single option passed to a inner `Dygraph` component
- Enable viewing a bar chart, which also uses the `DygraphContainer` component
- Fixes an issue where cells could not be saved to the dashboard due to their `id` being erased

Note that while a bar chart can be viewed in the DE / VEO, it cannot be saved to a cell yet. This is due to an API bug which I will fix or open an issue for shortly.


ui/src/shared/components/dygraph/Dygraph.tsx
```diff
@@ -380,7 +390,11 @@ class Dygraph extends Component<Props, State> {
           valueRange: this.getYRange(timeSeries),
         },
       },
-      ...this.props.options,
+      ...passedOptions,
+      // The following options are explicitly coerced to booleans, since
+      // dygraphs will not update if they change from `true` to `undefined`
```
OfTheDelmer: Nice

PR #1544, opened 2018-11-21
Add styles for single stat charts

### LINKED ISSUE #1526: should be able to view a single stat
as a query builder
when i build a dashboard cell
then i would like the ability to visualize that data as a single stat graph
then i would like the ability to display a trend line behind the single stat
then i would like the ability to display text or numbers
then i would like the ability to set a prefix or postfix for the value
then i would like the ability to set the precision of a single stat
then i would like the ability to set any number of thresholds for this single state
then i would like the ability to change the color of the text or the background of the single stat
then the size of the single stat should change relative to the size of the graph cell

chnn:
Closes #1526

screen shot 2018-11-20 at 4 54 56 pm


ui/src/shared/components/SingleStatTransform.tsx
```diff
@@ -2,12 +2,18 @@
...
+// Components
+import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
+
 // Parsing
 import getLastValues from 'src/shared/parsing/flux/fluxToSingleStat'
 
 // Types
 import {FluxTable} from 'src/types'
 
+const NON_NUMERIC_ERROR =
```
OfTheDelmer: Nice

ui/src/shared/components/cells/Cell.scss
```diff
@@ -16,8 +16,10 @@ $cell--header-size: 36px;
 .cell--view {
   position: absolute;
   width: 100%;
-  height: 100%;
-  top: 0;
+  height: calc(100% - $cell--header-size);
```
OfTheDelmer: nice adjustment

PR #1545, opened 2018-11-21
Fix staticcheck errors and make tool a required CI step

e-dard:
This PR fixes a lot of staticcheck (megacheck) errors, and makes the tool a required step during CI.

I have taken some time to add some documentation around this tool, and hope that it's enough for people to get a handle on what it does and how to satisfy the checks. I could elaborate further if there are some concrete needs.

The only annoying thing at the moment is that we need to ignore the generated `promql.go` file with a linter directive. I have added that to the `README` should someone go and regenerate the file and find that their PR now fails CI.


e-dard:
@mark-rushakoff thanks for grabbing those. I was rather blind to them all by the time I had finished as you might imagine...

In terms of:

> There were some other spots, where I thought it was acceptable for the error to start with an 
> uppercase letter when the first word matched the method name. I'd prefer those to remain 
> unchanged.

I think we need to commit all or nothing to the tool so the solution for those cases is:

 a) rewrite the error, e.g., `method SetFoo did not get what it wanted`
 b) add linter directive to ignore those cases.

e-dard:
@mark-rushakoff oh I. Did. Not. Know. That! ⭐️

PR #1549, opened 2018-11-21
Fix various XY chart options

### LINKED ISSUE #1527: should be able to view a bar chart

chnn:
Closes #1527 

- Fixes ability to set a y-axis label
- Fixes a screen tearing bug when setting y-tick prefixes and suffixes
- Fixes ability to choose a custom color scheme
- Reimplements the color scheme selector to use the clockface `Dropdown`
- Removes dead code

colors

screen shot 2018-11-21 at 12 29 48 pm


ui/src/shared/components/ColorSchemeDropdown.scss
```diff
@@ -0,0 +1,13 @@
...
+  justify-content: flex-start;
+  align-items: center;
+}
+
+.color-scheme-dropdown--swatches {
+  width: 100px;
+  height: 10px;
+  border-radius: 5px;
+  margin-right: 10px;
+
```
alexpaxton: So minimal!

ui/src/shared/components/dygraph/Dygraph.tsx
```diff
@@ -237,22 +218,16 @@ class Dygraph extends Component<Props, State> {
...
-        coloredDygraphSeries[seriesName] = {...series, color}
-      }
-    }
+  private get yLabel(): string | null {
+    return get(this.props, 'axes.y.label', null)
+  }
+
+  private get colors(): string[] {
+    const {timeSeries, colors} = this.props
+    const numSeries = get(timeSeries, '0.length', colors.length)
```
bthesorceror: Are we losing any type safety by not using getDeep here?

PR #1551, opened 2018-11-21
Add http endpoints for secret service

desa:
**Note**: this PR is currently blocked on https://github.com/influxdata/platform/pull/1458

This PR introduces the HTTP endpoints for the secret service. Additionally, it updates both the boltdb and vault secret services to support methods required in the http endpoints.

  - [x] Rebased/mergeable
  - [ ] Tests pass


bolt/secret.go
```diff
@@ -189,3 +189,58 @@ func encodeSecretValue(v string) []byte {
...
+					return err
+				}
+			}
+		}
+		return nil
+	})
+}
+
+// PatchSecrets patches all provided secrets and updates any previous values.
+func (c *Client) PatchSecrets(ctx context.Context, orgID platform.ID, m map[string]string) error {
```
gianarb: Why do we need to have both? Put and Patch? They looks the same to me

desa: PutSecrets will replace all of the existing secrets with the new secrets.
For example
If we had
```
s1=a
s2=b
```
and we did a `PUT s3=x` all that we'd end up storing is
```
s3=x
```
If we did the same thing with a patch we'd get
```
s1=a
s2=b
s3=x
```

We probably don't need both, but it doesn't hurt to have around just in case.

PR #1560, opened 2018-11-25
fix(http/task): add validation for beforeTime/afterTime params

zhulongcheng:
- [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -578,8 +578,8 @@ func decodeGetRunsRequest(ctx context.Context, r *http.Request, orgs platform.Or
 		req.filter.BeforeTime = bt
 	}
 
-	if at != "" && bt != "" && afterTime.Sub(beforeTime).Nanoseconds() > 0 {
-		return nil, kerrors.InvalidDataf("beforeTime must be greater than afterTime")
+	if at != "" && bt != "" && !beforeTime.After(afterTime) 0 {
+		return nil, kerrors.InvalidDataf("beforeTime must be later than afterTime")
 	}
 
 	return req, nil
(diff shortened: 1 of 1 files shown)


http/task_service.go
```diff
@@ -560,14 +560,26 @@ func decodeGetRunsRequest(ctx context.Context, r *http.Request, orgs platform.Or
...
+		if err != nil {
+			return nil, err
+		}
+		req.filter.BeforeTime = bt
 	}
 
-	if time := qp.Get("beforeTime"); time != "" {
-		// TODO (jm): verify valid RFC3339
-		req.filter.BeforeTime = time
+	if &afterTime != nil && &beforeTime != nil && afterTime.Sub(beforeTime).Nanoseconds() > 0 {
```
mark-rushakoff: `&afterTime != nil` will always be true because you're taking the address of `afterTime`.

If you break apart `if at := qp.Get("afterTime"); at != ""` to `at := qp.Get("afterTime"); if at != "" { ...`, then this line should be `if at != "" && bt != "" && !afterTime.After(beforeTime) {`. This is somewhat less brittle than checking `afterTime.IsZero()`.

http/task_service.go
```diff
@@ -560,14 +560,26 @@ func decodeGetRunsRequest(ctx context.Context, r *http.Request, orgs platform.Or
...
-		// TODO (jm): verify valid RFC3339
-		req.filter.AfterTime = time
+	var afterTime, beforeTime time.Time
+	if at := qp.Get("afterTime"); at != "" {
+		afterTime, err = time.Parse(time.RFC3339, at)
+		if err != nil {
+			return nil, err
+		}
+		req.filter.AfterTime = at
+
```
mark-rushakoff: Remove blank line.

PR #1565, opened 2018-11-26
Support for median in transpiler

### LINKED ISSUE #1488: Add median to the transpiler
Take a look at how "percentile" is done and copy that.

affo:
Closes #1488 

_What was the problem?_
No support for median in transpiler
_What was the solution?_
Fake the median for a 50th percentile.  
The side effect on the AST could cause problems. In that case, I will update the code to treat the median as a percentile only when creating the opSpec in `createCursor`

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/influxql/function.go
+++ b/query/influxql/function.go
@@ -47,7 +47,7 @@ func parseFunction(expr *influxql.Call) (*function, error) {
 		default:
 			return nil, fmt.Errorf("expected field argument in %s()", expr.Name)
 		}
-	case "min", "max", "sum", "first", "last", "mean":
+	case "min", "max", "sum", "first", "last", "mean", "median":
 		if exp, got := 1, len(expr.Args); exp != got {
 			return nil, fmt.Errorf("invalid number of arguments for %s, expected %d, got %d", expr.Name, exp, got)
 		}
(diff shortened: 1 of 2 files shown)


query/influxql/function.go
```diff
@@ -65,6 +65,14 @@ func parseFunction(expr *influxql.Call) (*function, error) {
 		default:
 			return nil, fmt.Errorf("expected field argument in %s()", expr.Name)
 		}
+	case "median":
+		if exp, got := 1, len(expr.Args); exp != got {
+			return nil, fmt.Errorf("invalid number of arguments for %s, expected %d, got %d", expr.Name, exp, got)
+		}
+
+		// fake it for a percentile
+		expr.Args = append(expr.Args, &influxql.NumberLiteral{Val: 0.5})
```
jsternberg: A fallthrough probably makes this more difficult. This modifies a copy of the AST that it may not really own so I would prefer it didn't do that, but it should also use the `exact_mean` method instead of `exact_selector`.

I think it would be easier to copy lines 219-228 and just put it here instead of using fallthrough and running through all of the checking code that percentile uses.

query/influxql/function.go
```diff
@@ -201,7 +193,20 @@ func createFunctionCursor(t *transpilerState, call *influxql.Call, in cursor, no
...
 	case "median":
-		fallthrough
+		value, ok := in.Value(call.Args[0])
+		if !ok {
+			return nil, fmt.Errorf("undefined variable: %s", call.Args[0])
+		}
+		cur.id = t.op("median", &transformations.PercentileOpSpec{
+			Percentile:  0.5,
+			Compression: 0,
+			Method:      "exact_selector",
```
affo: Kept `exact_selector` for coherence with percentile for now.  
In the case we want to aggregate, there is there any reason to prefer `exact_mean` over `estimate_tdigest`?

Note: `estimate_tdigest` is the default mode in flux.

jsternberg: Yea, it's because influxql itself uses the `exact_mean` method so the transpiler wants to get the same result. So even if it's a worse way of doing it, it keeps compatibility.

affo: Oh ok, I see

Issue #1291, opened 2018-11-05
convert /buckets endpoint error to platform error


kelwang:
https://github.com/influxdata/platform/issues/897

PR #1567, opened 2018-11-26
convert /buckets endpoint error to platform error 

kelwang:
Closes #https://github.com/influxdata/platform/issues/1291


bucket.go
```diff
@@ -27,6 +27,16 @@ type Bucket struct {
 	RetentionPeriod     time.Duration `json:"retentionPeriod"`
 }
 
+// ops for buckets
```
desa: What are ops? I think that more comments here might be useful.

http/bucket_service.go
```diff
@@ -474,6 +461,7 @@ type BucketService struct {
 	Addr               string
 	Token              string
 	InsecureSkipVerify bool
+	OpPrefix           string
```
desa: What does this do? I'm a little confused why this is needed.

kelwang:
wait platform master resolved in idpe, will get this merged

PR #1568, opened 2018-11-26
refactor(query): explicitly shut down query controller

mark-rushakoff:
The flux query controller was updated to include a Shutdown method a
while ago. Explicitly handle query controller creation and shutdown
where applicable.

In influxd, this ensures that outstanding queries are handled before the
process dies. In tests, this ensures that query controller goroutines
aren't leaked, which drastically simplifies reading full stack traces.

This change also registers query controller metrics with the prometheus
registry in influxd.


query/control/controller.go
```diff
@@ -37,3 +38,8 @@ func (c *Controller) Query(ctx context.Context, req *query.Request) (flux.Query,
 func (c *Controller) PrometheusCollectors() []prometheus.Collector {
 	return c.c.PrometheusCollectors()
 }
+
+// Shutdown shuts down the underlying Controller.
+func (c *Controller) Shutdown(ctx context.Context) error {
```
nathanielc: So passing a context to Shutdown allows you to cancel the shutdown and force it to terminate immediately?

mark-rushakoff: Yes, that's my understanding based on the godoc for the flux controller: https://godoc.org/github.com/influxdata/flux/control#Controller.Shutdown

storage/readservice/service.go
```diff
@@ -39,36 +38,16 @@ func NewProxyQueryService(engine *storage.Engine, bucketSvc platform.BucketServi
...
-			},
-		},
-	}, nil
-}
-
-type queryAdapter struct {
-	Controller *control.Controller
-}
-
-func (q *queryAdapter) Query(ctx context.Context, req *query.Request) (flux.Query, error) {
```
zeebo: I'm surprised this is no longer needed. Did you test the resulting binary and run some flux queries? Did I miss where this happens somewhere else, or is it not needed for some reason?

mark-rushakoff: It was a duplicate definition of https://github.com/influxdata/platform/blob/ce2674279609e3593347e74a33baf5fe0b047e5b/query/control/controller.go#L27-L34, so we're just relying on that other implementation, since query/control seems more canonical than storage/readservice, as far as managing query-related context values are concerned.

PR #1572, opened 2018-11-26
Tasks/fill in options in update task page

### LINKED ISSUE #1475: fill in the options on the left when editing a task
as a task user
when i edit a task
then i would like to see the options reflected in the ui

AlirieGray:
Closes #1475

_What was the problem?_
When opening the `TaskEditPage` to update an existing Task, the options pane defaulted to empty. We wanted the UI to be filled in with the options of the task being edited.

_What was the solution?_
Populate the task options UI with the options on the task object returned from the server.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/tasks/components/TaskScheduleFormField.tsx
```diff
@@ -5,96 +5,57 @@ import React, {PureComponent, ChangeEvent} from 'react'
...
-    }
     return (
       <>
-        <Input
-          name="cron"
-          placeholder="02***"
-          onChange={this.props.onChangeInput}
-          value={this.props.cron}
-        />
+        {this.scheduleInput}
```
chnn: Does this need to be a getter?

ui/src/tasks/actions/v2/index.ts
```diff
@@ -38,22 +41,18 @@ export type Action =
...
-  SetNewScript = 'SET_NEW_SCRIPT',
-  SetTasks = 'SET_TASKS',
-  SetSearchTerm = 'SET_TASKS_SEARCH_TERM',
-  SetCurrentScript = 'SET_CURRENT_SCRIPT',
-  SetCurrentTask = 'SET_CURRENT_TASK',
-  SetShowInactive = 'SET_TASKS_SHOW_INACTIVE',
-  SetDropdownOrgID = 'SET_DROPDOWN_ORG_ID',
+export interface SetAllTaskOptions {
+  type: 'SET_ALL_TASK_OPTIONS'
+  payload: Task
```
chnn: Should this be `TaskAPI`? Not sure what to make of the import
```
import {Task as TaskAPI, Organization} from 'src/api'
```

ui/src/tasks/reducers/v2/index.ts
```diff
@@ -15,18 +15,15 @@ export interface State {
...
   name: '',
-  intervalTime: '1',
-  intervalUnit: 'd',
-  delayTime: '20',
-  delayUnit: 'm',
+  interval: '',
+  delay: '',
   cron: '',
-  taskScheduleType: TaskSchedule.interval,
+  taskScheduleType: TaskSchedule.unselected,
```
chnn: Would it make more sense to default to `TaskSchedule.interval`? Seems like `unselected` isn't really ever a valid `TaskSchedule` variant other than here

ui/src/tasks/components/TaskScheduleFormField.tsx
AlirieGray: good catch

ui/src/tasks/actions/v2/index.ts
AlirieGray: I'm not 100% certain either, but I think it's because of this line in the file:
```
interface Task extends TaskAPI {
  organization: Organization
}
```
which adds another field to the type. maybe @bthesorceror can explain more?

PR #1575, opened 2018-11-26
 Enable switching the source for a dashboard cell

### LINKED ISSUE #1528: Allow the user to select the source for a cell
as a query builder
when i build a query in the explorer
then i would like the ability to select a query read endpoint for that query
then i would like the ability to dynamically change the query endpoint without editing the query

chnn:
Closes #1528 

screen shot 2018-11-26 at 3 27 54 pm


ui/src/clockface/components/dropdowns/Dropdown.tsx
```diff
@@ -234,7 +234,10 @@ class Dropdown extends Component<Props, State> {
       throw new Error('Dropdowns in ActionList mode require a titleText prop.')
     }
 
-    if (mode === DropdownMode.Radio && selectedID === '') {
+    if (
+      mode === DropdownMode.Radio &&
+      (isUndefined(selectedID) || isNull(selectedID))
```
AlirieGray: nice

ui/src/shared/components/RefreshingView.tsx
```diff
@@ -100,15 +87,4 @@ class RefreshingView extends PureComponent<Props> {
...
-  return {link}
-}
-
-const mdtp = {}
-
-export default connect<StateProps, DispatchProps, OwnProps>(
-  mstp,
-  mdtp
-)(RefreshingView)
+export default RefreshingView
```
AlirieGray: much cleaner!

ui/src/shared/reducers/v2/timeMachines.ts
```diff
@@ -243,6 +245,32 @@ const timeMachineReducer = (
...
+      const selectedQuery: string | undefined = get(
+        view,
+        `properties.queries.${activeQueryIndex}`
+      )
+
+      if (!selectedQuery) {
+        return state
+      }
+
+      const queries = view.properties.queries.map(
```
AlirieGray: lodash get here?

ui/src/shared/utils/view.ts
```diff
@@ -179,22 +186,28 @@ export function convertView<T extends View | NewView>(
...
 
-export function replaceQuery<T extends View | NewView>(
-  view: T,
-  text,
-  type = InfluxLanguage.Flux
-): T {
+// Replaces the text of the first query in a view
+export function replaceQuery<T extends View | NewView>(view: T, text): T {
   const anyView: any = view
+  const queries = anyView.properties.queries
```
AlirieGray: lodash get here too?

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -93,8 +102,20 @@ class TimeSeries extends Component<Props, State> {
     })
   }
 
+  private get queries(): URLQuery[] {
+    const {sources, queries, dynamicSourceURL} = this.props
+
+    return queries.filter(query => !!query.text).map(query => {
+      const source = sources.find(source => source.id === query.sourceID)
+      const url: string = source ? source.links.query : dynamicSourceURL
```
AlirieGray: I'm confused why `links.query` counts as a URL?

ui/src/shared/reducers/v2/timeMachines.ts
chnn: It should be guaranteed to exist based on the guard on line 258.

ui/src/shared/utils/view.ts
chnn: while `queries` might not exist on a view, `properties` always does, so I think it should be ok (the guard on line 194) checks if `queries` exists)

ui/src/shared/components/TimeSeries.tsx
chnn: `links` is an object that looks like:

```
  links: {
    buckets: '/api/v2/sources/02f8749edda60000/buckets',
    health: '/api/v2/sources/02f8749edda60000/health',
    query: '/api/v2/sources/02f8749edda60000/query',
    self: '/api/v2/sources/02f8749edda60000'
  }
```

does that make sense? maybe it should be called `queryLink` or something else instead

PR #1576, opened 2018-11-27
Switch Dashboard UI to use generated client

bthesorceror:
- [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/cur_swagger.yml
+++ b/http/cur_swagger.yml
@@ -4513,7 +4513,7 @@ components:
         x:
           type: integer
           format: int32
-        y:
+        "y":
           type: integer
           format: int32
         w:
@@ -4531,6 +4531,8 @@ components:
     Cell:
       type: object
       properties:
+        id:
+          type: string
         links:
           type: object
           properties:
@@ -4543,7 +4545,7 @@ components:
         x:
           type: integer
           format: int32
-        y:
+        "y":
           type: integer
           format: int32
         w:
--- a/ui/src/dashboards/actions/v2/index.ts
+++ b/ui/src/dashboards/actions/v2/index.ts
@@ -12,7 +12,6 @@ import {
   updateCells as updateCellsAJAX,
   addCell as addCellAJAX,
   deleteCell as deleteCellAJAX,
-  copyCell as copyCellAJAX,
 } from 'src/dashboards/apis/v2'
 import {createView as createViewAJAX} from 'src/dashboards/apis/v2/view'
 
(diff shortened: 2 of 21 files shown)


ui/src/types/v2/index.ts
```diff
@@ -15,6 +13,8 @@ import {
   ViewProperties,
   DashboardQuery,
 } from 'src/types/v2/dashboards'
+
+import {Cell, Dashboard} from 'src/api'
```
OfTheDelmer: nice

ui/src/types/v2/dashboards.ts
```diff
@@ -198,21 +199,6 @@ export enum ViewType {
   LogViewer = 'log-viewer',
 }
 
-interface DashboardLinks {
```
OfTheDelmer: :)

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -107,7 +107,14 @@ class Cells extends Component<Props & WithRouterProps, State> {
   }
 
   private get cells(): Layout[] {
-    return this.props.cells.map(c => ({...c, i: c.id}))
+    return this.props.cells.map(c => ({
+      ...c,
+      x: c.x,
```
OfTheDelmer: If the intent is letting `undefined` be a default value here, we could define a `DEFAULT_LAYOUT` constant and then spread it with each iteration.

http/cur_swagger.yml
```diff
@@ -4513,7 +4513,7 @@ components:
         x:
           type: integer
           format: int32
-        y:
+        "y":
```
OfTheDelmer: why quotes on y?

bthesorceror: generator requires it, otherwise it skips the y

ui/src/dashboards/actions/v2/index.ts
```diff
@@ -312,10 +312,9 @@ export const copyDashboardCellAsync = (
 ) => async (dispatch: Dispatch<Action>): Promise<void> => {
   try {
     const clonedCell = getClonedDashboardCell(dashboard, cell)
-    const cellFromServer = await copyCellAJAX(cell.links.copy, clonedCell)
```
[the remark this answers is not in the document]
bthesorceror: we never get a `cell.links.copy` value so this would have never worked anyway.

ui/src/shared/components/cells/Cells.tsx
bthesorceror: This is a work around for typescript, from the api x, y, w, h are optional.

ui/src/dashboards/actions/v2/index.ts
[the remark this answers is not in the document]
OfTheDelmer: ahh I see

PR #1578, opened 2018-11-27
Show inactive task count

### LINKED ISSUE #1472: A user might want to know that there are inactive tasks being hidden
as a task user
with admin permissions of higher
when i view the list of tasks 
then i would like to know the number of active and inactive tasks

it should be clear to the user that there are hidden inactive tasks

alexpaxton:
Closes #1472 

tasks-inactive-count

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/tasks/containers/TasksPage.tsx
```diff
@@ -171,6 +173,36 @@ class TasksPage extends PureComponent<Props, State> {
 
     return matchingTasks
   }
+
+  private get totalTaskCount(): number {
```
mavarius: Does this need a getter?

ui/src/tasks/containers/TasksPage.tsx
```diff
@@ -171,6 +173,36 @@ class TasksPage extends PureComponent<Props, State> {
...
+      t => t.status === TaskAPI.StatusEnum.Inactive
+    ).length
+
+    const allTasksAreHidden = hiddenCount === tasks.length
+
+    if (allTasksAreHidden) {
+      return null
+    }
+
+    if (hiddenCount === 1 && !showInactive) {
```
mavarius: Could early return if showInactive is true then combine the template strings for the count.

ui/src/tasks/containers/TasksPage.tsx
alexpaxton: Probably not, I just like how it reads on L104

PR #1580, opened 2018-11-27
Refactor time machines

chnn:
- Adds tests for the less trivial cases in the `timeMachinesReducer`
- Rename `RefreshingViewProperties` to `QueryViewProperties` to more effectively communicate its meaning (a `QueryViewProperties` is the type of any view's `properties` field that has `queries`)
- Narrows the type of the `view` in the `TimeMachineState` to `View<QueryViewProperties>`, so that we can avoid widening the `view` to `any` when accessing its `queries` property.


ui/src/shared/reducers/v2/timeMachines.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import {submitScript, setQuerySource} from 'src/shared/actions/v2/timeMachines'
+import {DE_TIME_MACHINE_ID} from 'src/shared/constants/timeMachine'
+import {DashboardQuery, InfluxLanguage} from 'src/types/v2/dashboards'
+
+describe('timeMachinesReducer', () => {
+  describe('SUBMIT_SCRIPT', () => {
+    test('replaces the first queries text if it exists', () => {
+      const state = initialState()
+      const timeMachine = state.timeMachines[DE_TIME_MACHINE_ID]
+      const viewProps: any = timeMachine.view.properties
```
Palakp41: Should this type be any?

ui/src/shared/utils/view.ts
```diff
@@ -195,6 +195,22 @@ export function replaceQuery<T extends View | NewView>(view: T, text): T {
     return view
   }
 
+  if (!queries[0]) {
```
Palakp41: Will there always be a queries[0] or should you check for that in the above if statement?

ui/src/shared/reducers/v2/timeMachines.test.ts
Palakp41: Nevermind! Looks good

ui/src/shared/utils/view.ts
chnn: there may not be, but in that case `queries[0]` is `undefined` and the contents of the if block will be executed

Issue #93, opened 2018-05-31
add derivative support to transpiler


aanthony1243:
need to add derivative as a valid function call. currently returns a 'not implemented' error

PR #1590, opened 2018-11-27
generate derivative tests from HT

jlapacik:
Un-skip these tests when https://github.com/influxdata/platform/issues/93 is implemented.


jlapacik:
I did not know that. I will rename random to testdata.

Issue #1587, opened 2018-11-27
GET to /api/v2/telegrafs/{telegrafID} does not fuzzy match on Accept header


ebb-tide:
A GET request to '/telegrafs/{telegrafID}' with header `{"Accept" : "application/json, text/plain, */*"}` should return what is specified as expected from a request with header `{"Accept: "application/json"}`
 as defined in: https://github.com/influxdata/platform/blob/e2959c61b7824185d16858b815c0b725497d6225/http/swagger.yml#L148

This is because https://github.com/influxdata/platform/blob/e2959c61b7824185d16858b815c0b725497d6225/http/telegraf.go#L155 is performing an exact match on the Accept header. 

Also: This issue was obscured because the default behavior of the API is not specified in the Swagger spec, even though the API does have a default behavior. It would be really nice to have default behavior described, when it exists.

goller:
I have a basic branch, but, I found several bugs in the API.  This lead me to writing the swagger test framework.

I'll circle back to this once the swagger tests are complete

PR #1598, opened 2018-11-28
fix(http): change default behavior match swagger

kelwang:
Closes #https://github.com/influxdata/platform/issues/1587


goller:
Closed in favor of #2138

PR #1599, opened 2018-11-28
Improve bulk series file writes.

benbjohnson:

This pull request preallocates space in the RHH hashmap before we bulk import into a partition. Normally the hashmap would keep doubling in space but that can take many iterations for large imports.

Previous benchmark:

```
BenchmarkSeriesPartition_CreateSeriesListIfNotExists/1000000-4         	       1	3611973385 ns/op
```

New benchmark:

```
BenchmarkSeriesPartition_CreateSeriesListIfNotExists/1000000-4         	       1	2519296210 ns/op
```

There is an approximately 30% improvement per partition (`3.6s` to `2.5s`) when performing a large import. The preallocation also occurs in a staggered lock so that the continuous lock time is broken up.

_What was the problem?_

We wanted to improve bulk import performance.

_What was the solution?_

Preallocate hashmap space and remove duplicate effort caused by constantly resizing the hashmap.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ vendor
 .vs
 .tern-project
 .DS_Store
+*.test
 
 # binary databases
 influxd.bolt
--- a/pkg/rhh/rhh.go
+++ b/pkg/rhh/rhh.go
@@ -41,6 +41,8 @@ func (m *HashMap) Reset() {
 	m.n = 0
 }
 
+func (m *HashMap) LoadFactor() int { return m.loadFactor }
+
 func (m *HashMap) Get(key []byte) interface{} {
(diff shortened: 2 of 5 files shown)


mark-rushakoff:

1. Download the patch from GitHub, by adding .patch to the end of the pull request URL, e.g. wget -P /tmp https://github.com/influxdata/platform/pull/1599.patch to download 1599.patch to /tmp (so that you don't accidentally check in 1599.patch in your influxdb directory).
1. Use git-am to apply the commits in the patch: git am /tmp/1599.patch. Then push the branch to GitHub and open a PR like you would normally.

PR #1604, opened 2018-11-28
Deal with yarn bug

docmerlin:
This is to workaround a yarn bug,

https://github.com/yarnpkg/yarn/issues/6312


docmerlin:
The tests passed, then I cleaned up some whitespace, and circleci started having an outage.
if you look at the run history, the tests passed.

PR #1616, opened 2018-11-28
update to go 1.11.2 in ci

docmerlin:


mark-rushakoff:
Is 1.11 not an alias to 1.11.2? I thought if we left it on 1.11, it would track the most recent 1.11.X release, which seems less work than manually updating.

docmerlin:
Yes, I was wrong.  Closing.

PR #1617, opened 2018-11-28
Add AutoRefresh to TimeMachine component

### LINKED ISSUE #1531: should be able to set an auto refresh timer in the DE/VEO
as a query builder
when i am building a query in the explorer or the cell editor
then i would like the ability to select a refresh rate for the data
then i would like the default to the paused
then i would like the ability to manually refresh the query

AlirieGray:
Closes #1531

_What was the problem?_
Previously, AutoRefresh only worked on the Dashboard page, not for the DE or VEO. Also, the "submit" button in the DE/VEO did not cause the query to be re-submitted.

_What was the solution?_
Add the `AutoRefreshDropdown` component to the `TimeMachineControls` component and add a `submitCount` property to the TimeMachine redux state that gets updated when the "submit" button is pressed or when the graph is auto-refreshed.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/actions/v2/timeMachines.ts
```diff
@@ -269,3 +270,11 @@ export const setQuerySource = (sourceID: string): SetQuerySourceAction => ({
   type: 'SET_QUERY_SOURCE',
   payload: {sourceID},
 })
+
+interface IncrementSubmitCount {
+  type: 'INCREMENT_SUBMIT_COUNT'
+}
+
+export const incrementSubmitCount = (): IncrementSubmitCount => ({
+  type: 'INCREMENT_SUBMIT_COUNT',
```
ebb-tide: Yeah, so.. my one gripe here is that "submit_count" is not a very self-explanatory variable name. On this particular line I'd say that "increment" is not a very useful concept either. How about "Refresh_Token" and "Update_Refresh_Token" ?

ui/src/shared/components/TimeMachineRefreshDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+interface DispatchProps {
+  onIncrementSubmitCount: typeof incrementSubmitCount
+}
+
+interface State {
+  autoRefreshInterval: number
+}
+
+class TimeMachineRefreshDropdown extends PureComponent<DispatchProps, State> {
+  public state: State = {autoRefreshInterval: 1000 * 5}
```
ebb-tide: nit: magic number I feel like we had durations in an enum somewhere?

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -155,6 +140,10 @@ class TimeSeries extends Component<Props, State> {
   }
 
   private shouldReload(prevProps: Props) {
+    if (prevProps.submitCount !== this.props.submitCount) {
+      return true
```
ebb-tide: where the magic happens!

PR #1620, opened 2018-11-28
test cases ported from HT

jlapacik:
Closes https://github.com/influxdata/flux/issues/236.

Generated test cases ported from HT. Tests with failures have been skipped with the appropriate issue linked.


query/functions/query_test.go
```diff
@@ -284,6 +317,21 @@ func testGeneratedInfluxQL(t testing.TB, prefix, queryExt string) {
 	}
 
 	if ok, err := executetest.EqualResults(exp, got); !ok {
+		log.Println("Expected Tables:")
```
wolffcm: I think Go's testing framework will produce better output if you use the testing objects's `t.Log()` method for diagnostics.  Likewise, instead of `os.Stdout` to capture the formatted tables you can use a `bytes.Buffer` and then log that through the testing object.

PR #1621, opened 2018-11-29
refactor(task): store meta objects by raw ID, not ID string

mark-rushakoff:
The keys used to be strings, back when platform.ID was a byte slice, not
a uint64.

And rename the field from runners to meta, which is much more accurate.


mark-rushakoff:
Jenkins is failing due to the not-yet-resolved `tables.Statistics undefined (type flux.TableIterator has no field or method Statistics)` issue.

PR #1627, opened 2018-11-29
fix(task): add limit field to TaskFilter

zhulongcheng:
https://github.com/influxdata/platform/blob/master/task/platform_adapter.go#L45-L48

_What was the problem?_
Currently, the `FindTasks` method uses a fixed `pageSize` to filter tasks.
The number of tasks to return should be specified by callers.

_What was the solution?_
Update swagger to include limit param.
Add limit field to `TaskFilter`.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/cmd/influx/task.go
+++ b/cmd/influx/task.go
@@ -173,6 +173,7 @@ type TaskFindFlags struct {
 	user  string
 	id    string
 	orgID string
+	limit int
 }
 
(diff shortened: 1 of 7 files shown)


zhulongcheng:
Hi @mark-rushakoff , request a review. Thanks!

mark-rushakoff:
LGTM, thanks.

PR #1628, opened 2018-11-29
WIP: Spike for generic key value store

desa:
This PR is a spike intended to be a proof of concept before we go on to adapt other services.

This PR introduces a new directory `kv` that contains an interface for a generic key value store. The intention is that all of the services to be implemented in this package. This way we can have a single location where the service implementations live and we have those implementations backed by anything that implements the `kv.Store` interface.

The interface as it is now should closely resemble the boltdb/stm style. It is likely that this interface may need to change a bit to allow for other implementations.

My intention is to have this be a place where we can discuss the design decisions and work on a suitable interface for us to build off of.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/kv/cursor.go
+++ b/kv/cursor.go
@@ -1,0 +1,80 @@
+package kv
+
+import (
+	"bytes"
+	"sort"
+)
+
+// staticCursor implements the Cursor interface for a slice of
+// static key value pairs.
+type staticCursor struct {
+	idx   int
+	pairs []Pair
+}
+
+// Pair is a struct for key value pairs.
+type Pair struct {
+	Key   []byte
+	Value []byte
+}
+
+// NewStaticCursor returns an instance of a StaticCursor. It
+// destructively sorts the provided pairs to be in key ascending order.
+func NewStaticCursor(pairs []Pair) Cursor {
+	sort.Slice(pairs, func(i, j int) bool {
+		return bytes.Compare(pairs[i].Key, pairs[j].Key) < 0
+	})
+	return &staticCursor{
+		pairs: pairs,
+	}
+}
+
+// Seek searches the slice for the first key with the provided prefix.
+func (c *staticCursor) Seek(prefix []byte) ([]byte, []byte) {
+	// TODO: do binary search for prefix since pairs are ordered.
+	for i, pair := range c.pairs {
+		if bytes.HasPrefix(pair.Key, prefix) {
+			c.idx = i
+			return pair.Key, pair.Value
+		}
+	}
+
+	return nil, nil
+}
+
+func (c *staticCursor) getValueAtIndex(delta int) ([]byte, []byte) {
+	idx := c.idx + delta
+	if idx < 0 {
+		return nil, nil
+	}
+
+	if idx >= len(c.pairs) {
+		return nil, nil
+	}
+
+	c.idx = idx
+
+	pair := c.pairs[c.idx]
+
+	return pair.Key, pair.Value
+}
+
+// First retrieves the first element in the cursor.
+func (c *staticCursor) First() ([]byte, []byte) {
+	return c.getValueAtIndex(-c.idx)
+}
+
+// Last retrieves the last element in the cursor.
+func (c *staticCursor) Last() ([]byte, []byte) {
+	return c.getValueAtIndex(len(c.pairs) - 1 - c.idx)
+}
+
+// Next retrieves the next entry in the cursor.
+func (c *staticCursor) Next() ([]byte, []byte) {
+	return c.getValueAtIndex(1)
+}
+
+// Prev retrieves the previous entry in the cursor.
+func (c *staticCursor) Prev() ([]byte, []byte) {
+	return c.getValueAtIndex(-1)
+}
--- a/kv/store.go
+++ b/kv/store.go
@@ -1,0 +1,52 @@
+package kv
+
+import (
+	"context"
+	"errors"
+)
+
+var (
+	// ErrKeyNotFound is the error returned when the key requested is not found.
+	ErrKeyNotFound = errors.New("key not found")
+	// ErrTxNotWritable is the error returned when an mutable operation is called during
+	// a non-writable transaction.
+	ErrTxNotWritable = errors.New("transaction is not writable")
+)
+
+// Store is an interface for a generic key value store. It is modeled after
+// the boltdb database struct.
+type Store interface {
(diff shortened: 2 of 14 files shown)


kv/store.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+)
+
+var (
+	ErrKeyNotFound = fmt.Errorf("key not found")
+)
+
+type Store interface {
+	Tx(bool) (Tx, error)
```
desa: We could probably get rid of this here. I don't know that we need to explicitly expose it in this way.

kv/store.go
```diff
@@ -0,0 +1,37 @@
...
+
+type Bucket interface {
+	Get(key []byte) ([]byte, error)
+	Put(key, value []byte) error
+	Delete(key []byte) error
+	Cursor() (Cursor, error)
+}
+
+type Cursor interface {
+	Seek(prefix []byte) (k []byte, v []byte, err error)
```
desa: We might be able to drop there error here, but I'm not sure.

kv/cursor.go
```diff
@@ -0,0 +1,67 @@
...
+	return c.getValueAtIndex()
+}
+
+func (c *StaticCursor) Next() ([]byte, []byte, error) {
+	c.idx += 1
+	return c.getValueAtIndex()
+}
+
+func (c *StaticCursor) Prev() ([]byte, []byte, error) {
+	c.idx += 1
```
leodido: I suspect you to want this to be subtraction, not addition :)

kv/cursor.go
```diff
@@ -0,0 +1,67 @@
...
+			c.idx = i
+			return pair.Key, pair.Value, nil
+		}
+	}
+
+	return nil, nil, fmt.Errorf("prefix not found")
+}
+
+func (c *StaticCursor) getValueAtIndex() ([]byte, []byte, error) {
+	if c.idx < 0 {
```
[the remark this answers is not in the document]
desa: I think I follow the idea, but it'd be good to clarify. Ping me when you get a chance :)

desa:
Just requested a review from everyone. I'd like to get feedback about whats going on here. Before moving on to cleanup.

kv/store.go
```diff
@@ -0,0 +1,43 @@
...
+	// View opens up a transaction that will not write to any data. Implementing interfaces
+	// should take care to ensure that all view transactions do not mutate any data.
+	View(func(Tx) error) error
+	// Update opens up a transaction that will mutate data.
+	Update(func(Tx) error) error
+}
+
+// Tx is a transaction in the store.
+type Tx interface {
+	CreateBucketIfNotExists(b []byte) error
```
goller: Document that this will error if used in a View...
Also, would you document that `Put` and `Delete` also error.

kv/store.go
```diff
@@ -0,0 +1,43 @@
...
+	// ErrKeyNotFound is the error returned when the key requested is not found.
+	ErrKeyNotFound = errors.New("key not found")
+)
+
+// Store is an interface for a generic key value store. It is modeled after
+// the boltdb database struct.
+type Store interface {
+	// View opens up a transaction that will not write to any data. Implementing interfaces
+	// should take care to ensure that all view transactions do not mutate any data.
+	View(func(Tx) error) error
```
goller: I'm not sure that View needs transactions.

kv/store.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Tx is a transaction in the store.
+type Tx interface {
+	CreateBucketIfNotExists(b []byte) error
+	Bucket(b []byte) (Bucket, error)
+}
+
+// Bucket is the abstraction used to perform get/put/delete/get-many operations
+// in a key value store.
+type Bucket interface {
```
goller: Maybe add a basic "conformance" test for transactions where you do a `Put`, `Get`, `Delete` followed by a `View` checking for the key.

kv/store.go
```diff
@@ -0,0 +1,43 @@
...
+import "errors"
+
+var (
+	// ErrKeyNotFound is the error returned when the key requested is not found.
+	ErrKeyNotFound = errors.New("key not found")
+)
+
+// Store is an interface for a generic key value store. It is modeled after
+// the boltdb database struct.
+type Store interface {
```
goller: Probably add context to all these interfaces

kv/store.go
goller: Perhaps make this an internal detail of the bolt impl.

kv/store.go
desa: I think this is correct, but I'd like to keep it there so that there is a symmetry for the method calls for view vs update. Happy to change this if you think it'd be best.

kv/store.go
desa: Added methods 
```go
// Tx is a transaction in the store.
type Tx interface {
	Bucket(b []byte) (Bucket, error)
	Context() context.Context
	WithContext(ctx context.Context)
}
```
to the transaction so that we don't can use context without modifying the function signatures.

kv/cursor.go
```diff
@@ -0,0 +1,82 @@
...
+	sort.Slice(pairs, func(i, j int) bool {
+		return bytes.Compare(pairs[i].Key, pairs[j].Key) < 0
+	})
+	return &StaticCursor{
+		pairs: pairs,
+	}
+}
+
+// Seek searches the slice for the first key with the provided prefix.
+func (c *StaticCursor) Seek(prefix []byte) ([]byte, []byte, error) {
```
leodido: What do you think to return `(Pair, error)` ?

Or `([]byte, []byte)`?

kv/store.go
leodido: Yep doable, imho. Anyway we should think about user readability.

kv/store.go
leodido: Also a test with two actors concurrently updating the same key.

kv/cursor.go
desa: I think I like `([]byte,[]byte)`. Which would you prefer to see? Either works for me.

leodido: I think I like more `([]byte, []byte)` too.

kv/store.go
desa: Good idea. I'll add a test for that.

kv/cursor.go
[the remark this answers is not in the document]
leodido: We did it :). Resolving.

PR #1631, opened 2018-11-29
feat(ui/DataLoaders): Streaming/Listening step

### LINKED ISSUE #1459: Add a card in onboarding for listening for streaming data
as a new user
when i configure a set of streaming data sources
then i would like to confirm that the data is loading properly before leaving the data loading section of the onboarding

ischolten:
Closes #1459

Add a streaming data step in the onboarding process.
If you have a telegraf config and have passed the setup step, it displays the commands for you to run to have telegraf put data into the bucket created in step one.
Clicking the listen button will cause it to check for data in the created buck and then wait five seconds and check again for no longer than a minute. As soon as data is found it will inform you of such. if no data is found and the timer passes 60 seconds, it will say there's an error.

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/types/v2/dataSources.ts
```diff
@@ -7,6 +7,13 @@ export enum ConfigurationState {
   Error = 'error',
 }
 
+export enum DataSourceType {
```
alexpaxton: Moar enums!

ui/src/onboarding/components/DataListening.tsx
```diff
@@ -0,0 +1,134 @@
...
+    this.state = {loading: RemoteDataState.NotStarted}
+  }
+
+  public componentWillUnmount() {
+    clearInterval(this.intervalID)
+  }
+
+  public render() {
+    return (
+      <div className="wizard-step--body-streaming">
```
alexpaxton: More of a nitpick, but considering how many components are part of onboarding now it might make sense to give this each one its own stylesheet with namespacing. I'm starting to think that namespacing all the onboarding components to `wizard` or `onboarding` gets excessively verbose.

In this case the classname could be simply `.data-listening`

ischolten: i like that! i agree with your overall comment that we should wait until some of the dust settles with this. i think this whole dataloaders concept is going to be more of a separate thing rather than only in onboarding so that will probs be a good time to separate everything out

PR #1633, opened 2018-11-30
Update flux dependency

e-dard:
Fixes the `master` build.


e-dard:
I don't think we need this.

PR #1668, opened 2018-12-03
Labels

jademcgough:
Add label service.


http/dashboard_test.go
```diff
@@ -275,7 +278,8 @@ func TestService_handleGetDashboard(t *testing.T) {
   "links": {
     "self": "/api/v2/dashboards/020f755c3c082000",
     "log": "/api/v2/dashboards/020f755c3c082000/log",
-    "cells": "/api/v2/dashboards/020f755c3c082000/cells"
+    "cells": "/api/v2/dashboards/020f755c3c082000/cells",
+		"labels": "/api/v2/dashboards/020f755c3c082000/labels"
```
gianarb: This as too many tabs or spaces for what I can see.

http/dashboard_test.go
```diff
@@ -417,7 +422,8 @@ func TestService_handlePostDashboard(t *testing.T) {
...
```
gianarb: Same here

http/dashboard_test.go
```diff
@@ -643,7 +651,8 @@ func TestService_handlePatchDashboard(t *testing.T) {
...
```
gianarb: Same here

PR #1674, opened 2018-12-03
feat(ui/dataLoaders): Move the listening/verify step into a separate onboarding step 

### LINKED ISSUE #1648: dataLoaders: the data listening step should be its own step in the wizard and not under configure. should be renamed to verify
as a new user
when i finish configuring my bundles
then i would like to see a step at the top nav after configure named Verify

the data listening step should be its own step in the wizard and not under configure. should be renamed to verify

ischolten:
Closes #1648

Move listen to data/ verify data in bucket step to a separate step in the wizard rather than nested under configure

  - [x] Rebased/mergeable
  - [ ] Tests pass


ischolten:
Ah yes! I shall add some tests

PR #1686, opened 2018-12-04
Introduce Draggable Resizer Component

alexpaxton:
Goodbye `Threesizer`, hello `DraggableResizer`!

eight-sizer
vertical-8-sizer

### How to use `DraggableResizer`

- `DraggableResizer` is a clockface component so just import like so `import {DraggableResizer} from 'src/clockface'`
- Purely presentational
- Composable API
- State is managed by the parent component that consumes `DraggableResizer`, which will likely end up as local storage or component state
- Ability to specify minimum sizes per section in pixels
- You can use as many sections as you like
- Uses just flexbox for positioning so no cropping or context stacking woes

### To-Do

- Some fine tuning for minimum sizing of sections, going to do in another PR

### Example

```
import {DraggableResizer, Direction} from 'src/clockface'
```
```
this.state = {
  dragHandlePositions: [0.125, 0.25, 0.375, 0.5, 0.625, 0.75, 0.875],
}
```
```
const {dragHandlePositions} = this.state

return (
  <DraggableResizer
    onChangePositions={this.handleResizerChange}
    handlePositions={dragHandlePositions}
    handleDirection={Direction.Horizontal}
  >
    <DraggableResizer.Panel id="panel-a" minSizePixels={0}>
      <div className="test">A</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-b" minSizePixels={0}>
      <div className="test">B</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-c" minSizePixels={0}>
      <div className="test">C</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-d" minSizePixels={0}>
      <div className="test">D</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-e" minSizePixels={0}>
      <div className="test">E</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-f" minSizePixels={0}>
      <div className="test">F</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-g" minSizePixels={0}>
      <div className="test">G</div>
    </DraggableResizer.Panel>
    <DraggableResizer.Panel id="panel-h" minSizePixels={0}>
      <div className="test">H</div>
    </DraggableResizer.Panel>
  </DraggableResizer>
)
```
```
private handleResizerChange = dragHandlePositions => {
  this.setState({dragHandlePositions})
}
```

### Checklist

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+// Decorators
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+// Styles
+import './DraggableResizer.scss'
+
+interface Props {
+  children: JSX.Element[]
+  handleDirection: Direction
```
chnn: what do you think about using `Row` / `Column` lingo rather than `Vertical` / `Horizontal`?

I think `Row` / `Column` might signal a bit since people are familiar with the flexbox api, whereas `Vertical` / `Horizontal` feels a bit ambiguous (is it the handle that's `Horizontal`, or the way in which divisions are laid out?)

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
```diff
@@ -0,0 +1,191 @@
...
+    this.state = {
+      initialized: false,
+      dragIndex: null,
+    }
+  }
+
+  public componentDidMount() {
+    // This allows for one extra render which
+    // ensures the panels don't render without positioning information
+    this.setState({initialized: true})
```
chnn: Is this still needed?

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
```diff
@@ -0,0 +1,191 @@
...
+              />
+              <DraggableResizerHandle
+                {...child.props}
+                dragIndex={i}
+                onStartDrag={this.handleStartDrag}
+                dragging={dragging}
+              />
+            </>
+          )
+        } else if (child.type === DraggableResizerPanel) {
```
chnn: It felt a little bit difficult to read which parts of these if blocks were the same and which were different. What about something like
```tsx
      return React.Children.map(children, (child: JSX.Element, i: number) => {
        if (child.type !== DraggableResizerPanel) {
          return
        }

        const isLastPanel = i === panelsCount - 1

        return (
          <>
            <DraggableResizerPanel
              {...child.props}
              sizePercent={this.calculatePanelSize(i)}
            />
            {!isLastPanel && (
              <DraggableResizerHandle
                {...child.props}
                dragIndex={i}
                onStartDrag={this.handleStartDrag}
                dragging={dragIndex === i}
              />
            )}
          </>
        )
      })

```

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    if (panelIndex === 0) {
+      return handlePositions[panelIndex]
+    }
+
+    if (panelIndex === handlePositions.length) {
+      return 1 - handlePositions[prevPanelIndex]
+    }
+
+    return handlePositions[panelIndex] - handlePositions[prevPanelIndex]
```
chnn: woop awesome

ui/src/clockface/components/draggable_resizer/DraggableResizerPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+import React, {Component, CSSProperties} from 'react'
+
+// Decorators
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+interface Props {
+  id: string
+  children: JSX.Element
+  minSizePixels: number
+  sizePercent?: number
```
chnn: a percent seems like it's a number between 0 and 100. I'm not sure what to call this... maybe just `size`? dunno how to communicate that it should be between 0 and 1 and that all sizes should sum up to 1

ui/src/clockface/components/draggable_resizer/DraggableResizer.scss
```diff
@@ -0,0 +1,154 @@
...
+    &.draggable-resizer--dragging,
+    &.draggable-resizer--dragging:hover {
+      cursor: col-resize !important;
+    }
+  }
+}
+
+.draggable-resizer--dragging {
+  position: relative;
+  z-index: 9999;
```
chnn: maybe use a `$z--` variable for this and subsequent z-indices?

ui/src/clockface/components/draggable_resizer/DraggableResizer.scss
```diff
@@ -0,0 +1,154 @@
...
+.draggable-resizer--horizontal .draggable-resizer--handle-dragging:after {
+  width: calc(100% - 60px);
+}
+
+.draggable-resizer--vertical .draggable-resizer--handle-dragging:after {
+  height: calc(100% - 60px);
+}
+
+
+.test {
```
chnn: can probably remove this

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
alexpaxton: Ah that is much cleaner

ui/src/clockface/components/draggable_resizer/DraggableResizer.scss
alexpaxton: good call, I actually have an area in the css modules for this exact reason

ui/src/clockface/components/draggable_resizer/DraggableResizerPanel.tsx
alexpaxton: I wish we could make a special type for that

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
alexpaxton: If I delete it the component renders strangely the first time. There is probably a way around it, seems like a race condition

ui/src/clockface/components/draggable_resizer/DraggableResizer.tsx
alexpaxton: Not opposed, does seem slightly more semantic. There are a handful of other components that use the `Direction` enum, I think it won't affect the intuitiveness of those

alexpaxton:
Going to build in the support for minimum pixels sizes in a separate PR

PR #1687, opened 2018-12-04
feat(ui/dataLoaders): Update redux to store telegraf plugin arrays rather than dataSource

ischolten:
The concept of data sources and telegraf plugins were conflated. Now we only store the array of plugins for the data source rather than the entire data source which is unnecessary for data loaders that are not type streaming.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/onboarding/actions/dataLoaders.ts
```diff
@@ -1,11 +1,11 @@
...
-import {DataSource, DataLoaderType} from 'src/types/v2/dataLoaders'
+import {TelegrafPlugin, DataLoaderType} from 'src/types/v2/dataLoaders'
 
 export type Action =
   | SetDataLoadersType
-  | AddDataSource
-  | RemoveDataSource
-  | SetActiveDataSource
+  | AddTelegrafPlugin
+  | RemoveTelegrafPlugin
```
OfTheDelmer: much clearer

ui/src/onboarding/components/selectionStep/SelectDataSourceStep.tsx
```diff
@@ -119,29 +120,29 @@ class SelectDataSourceStep extends PureComponent<Props, State> {
...
-    const active = dataSources.length === 0
-    this.props.onAddDataSource({
-      name: dataSource,
+    const active = telegrafPlugins.length === 0
+    this.props.onAddTelegrafPlugin({
+      name: telegrafPlugin,
       configured: ConfigurationState.Unconfigured,
       active,
-      configs: null,
+      config: {},
```
OfTheDelmer: Was `configs: null` previously creating an issue?

ui/src/onboarding/reducers/dataLoaders.test.ts
```diff
@@ -7,117 +7,113 @@ import dataLoadersReducer, {
...
   setDataLoadersType,
-  addDataSource,
-  removeDataSource,
+  addTelegrafPlugin,
+  removeTelegrafPlugin,
+  setActiveTelegrafPlugin,
 } from 'src/onboarding/actions/dataLoaders'
 
+// Types
+import {TelegrafRequestPlugins} from 'src/api'
```
OfTheDelmer: great to see usage of the `src/api`

ui/src/onboarding/components/selectionStep/SelectDataSourceStep.tsx
ischolten: configs needed to be changed to config, and either null or {} seems to work here in the backend. this is just a place holder for now until the swagger gets updated and we have more type safety from the generated client.

PR #1694, opened 2018-12-04
Move View Type Selection

alexpaxton:
Taking incremental steps towards a bigger plan for the VEO
View type selection is now a dropdown in the top left of the graph instead of a large card-based grid.

move-view-type-switcher
screen shot 2018-12-04 at 10 18 46 am

This makes room for 2 sets of axis controls, as well as any other options that we introduce in the future

  - [x] Rebased/mergeable
  - [ ] Tests pass


jademcgough:
That failure's a flaky test. I'll try to deal with it today.

Issue #1530, opened 2018-11-20
should see syntax errors in editor as I type


russorat:
as a query builder
with admin permissions or higher
when i am writing a query from scratch
then i would like to see the errors in my query as i type them

desa:
```
{
  "valid": false,
  "errors": [
    {
      "line": 11,
      "column": 41,
      "character": 674,
      "message": "rule StringLiteral: string literal not terminated"
    },
    {
      "line": 11,
      "column": 48,
      "character": 681,
      "message": "rule ArrayElements: interface conversion: string is not ast.Expression: missing method Copy"
    }
  ]
}
```

desa:
@russorat @121watts backend parts of this are complete https://github.com/influxdata/platform/pull/1753

PR #1753, opened 2018-12-05
add /query/analyze endpoint query http handler

desa:
Backend Portion of #1530

_What was the problem?_
Wasn't able to analyze query request without actually executing the query.

_What was the solution?_
Add an analyze endpoint that returns friendly json like
```json
{
  "errors": [
    {
      "line": 11,
      "column": 41,
      "character": 674,
      "message": "rule StringLiteral: string literal not terminated"
    },
    {
      "line": 11,
      "column": 48,
      "character": 681,
      "message": "rule ArrayElements: interface conversion: string is not ast.Expression: missing method Copy"
    }
  ]
}
```

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/query.go
+++ b/http/query.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
+	"strconv"
 	"time"
 	"unicode/utf8"
 
@@ -12,8 +14,10 @@ import (
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/csv"
 	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/parser"
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/values"
+	"github.com/influxdata/influxql"
 	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/kit/errors"
 	"github.com/influxdata/platform/query"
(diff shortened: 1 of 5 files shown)


http/query.go
```diff
@@ -97,6 +101,133 @@ func (r QueryRequest) Validate() error {
 	return nil
 }
 
+type QueryAnalysis struct {
```
goller: Hey @desa would you comment all the public things?

http/query.go
```diff
@@ -97,6 +101,133 @@ func (r QueryRequest) Validate() error {
...
+	Valid  bool              `json:"valid"`
```
goller: This seems redundant as len(Errors) == 0 is the same information

http/query_handler.go
```diff
@@ -50,6 +50,7 @@ func NewFluxHandler() *FluxHandler {
 
 	h.HandlerFunc("POST", fluxPath, h.handlePostQuery)
 	h.HandlerFunc("POST", "/api/v2/query/ast", h.postFluxAST)
+	h.HandlerFunc("POST", "/api/v2/query/analyze", h.postQueryAnalyze)
```
goller: I think this route needs to be added to links

desa:
@goller should be ready for another review.

desa:
@goller should be ready for another review

http/swagger.yml
```diff
@@ -2211,6 +2211,51 @@ paths:
...
+      - Query
+    summary: analyze an influxql or flux query
+    parameters:
+      - in: header
+        name: Content-Type
+        schema:
+          type: string
+          enum:
+            - application/json
+      - in: header
```
goller: No need to have authorization in the header like this.  We'll have use the openapi authorization specification.

desa:
@goller this good to merge?

Issue #1754, opened 2018-12-05
Add ability to build queries with a v1 source


chnn:
Behind the scenes, the current query builder executes Flux metaqueries in order to populate lists of buckets, measurements, and fields.

The query builder should be updated so that when connected to a `v1` source, it executes InfluxQL metaqueries instead.

chnn:
Closing in favor of #1775

PR #1755, opened 2018-12-05
Add ability to select tags in query builder

### LINKED ISSUE #1683: Ability to select buckets, measurements, fields, and functions
As a user building a query in the Data Explorer or CEO with the visual query builder, I should be able to select

- 1 or more buckets
- 1 or more measurements for that bucket
- 0 or more fields for that bucket
- 0 or more functions 

For each list of items, I should be able to select all or none of the items in the list. I should also be able to filter the items with a search field.

chnn:
Closes #1683 

query builder

Some differences from the specification:

- The builder won't work if connected to a `v1` source. This is tracked in #1754.
- There's no “select all“ option for any list in the builder. This is because in order to support `v1` sources, we need to execute a separate metaquery for each possible pairing of a selected bucket and selected measurement (in order to show fields). For a user with many buckets and measurements, these queries would be too expensive to run.


ui/src/shared/components/BuilderCard.tsx
```diff
@@ -0,0 +1,124 @@
...
+      return (
+        <div className={this.containerClass}>
+          <div className="builder-card--empty">Failed to load</div>
+        </div>
+      )
+    }
+
+    if (status === RemoteDataState.NotStarted) {
+      return (
+        <div className={this.containerClass}>
```
bthesorceror: Could we extract a component here, looks like we have the same markup 3 times with just the inner text changing.

bthesorceror: I could encapsulate the container class logic as well

ui/src/shared/components/BuilderCardLimitMessage.tsx
```diff
@@ -0,0 +1,22 @@
...
+    return null
+  }
+
+  return (
+    <div className="builder-card-limit-message">
+      Showing first {limitCount} results. Use the search bar to find more.
+    </div>
+  )
+
+  return null
```
bthesorceror: this can never be reached.

ui/src/shared/components/TimeMachineQueryBuilder.tsx
```diff
@@ -1,114 +1,330 @@
...
+interface StateProps {
+  queryURL: string
+  sourceType: SourceType
+}
 
 interface State {
   buckets: string[]
   bucketsStatus: RemoteDataState
-  selectedBuckets: string[]
+  bucketsSelection: string[]
```
bthesorceror: The amount of state this one component has seems excessive, is there anyway this could be broken up?

ui/src/shared/apis/v2/queryBuilder.ts
```diff
@@ -0,0 +1,153 @@
...
+  sourceType: SourceType,
+  bucket: string,
+  searchTerm: string = ''
+): Promise<string[]> {
+  if (sourceType === SourceType.V1) {
+    throw new Error('metaqueries not yet implemented for SourceType.V1')
+  }
+
+  const resp = await findMeasurementsFlux(url, bucket, searchTerm)
+  const parsed = extractCol(resp, '_measurement')
```
bthesorceror: same as above

ui/src/shared/apis/v2/queryBuilder.ts
```diff
@@ -0,0 +1,153 @@
...
+  searchTerm: string = ''
+): Promise<string[]> {
+  if (sourceType === SourceType.V1) {
+    throw new Error('metaqueries not yet implemented for SourceType.V1')
+  }
+
+  const resp = await findFieldsFlux(url, bucket, measurements, searchTerm)
+  const parsed = extractCol(resp, '_field')
+
+  return parsed
```
bthesorceror: same as above

ui/src/shared/components/TimeMachineQueryBuilder.tsx
chnn: I'm certainly open to ideas, but I couldn't think of a way. Each entry in the state has a necessary tight coupling to many other entries—e.g. the fields need to be reloaded when a different measurement or bucket is selected, and so on.

We could move the state to Redux, but that might be messy in a different way (huge thunks, trickier handling of race conditions, and a boatload of actions).

ui/src/shared/components/BuilderCardLimitMessage.tsx
chnn: thanks, will remove

ui/src/shared/apis/v2/queryBuilder.ts
```diff
@@ -0,0 +1,153 @@
...
+  url: string,
+  sourceType: SourceType,
+  searchTerm?: string
+) {
+  if (sourceType === SourceType.V1) {
+    throw new Error('metaqueries not yet implemented for SourceType.V1')
+  }
+
+  const resp = await findBucketsFlux(url, searchTerm)
+  const parsed = extractCol(resp, 'name')
```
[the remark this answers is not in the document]
chnn: It's a stylistic choice I picked up somewhere... it makes it easier to quickly set a debugger or a console.log

PR #1762, opened 2018-12-06
Change delay to offset in tasks to match the flux spec

### LINKED ISSUE #1757: rename delay to be offset to match flux
In task options delay needs to be offset; this is a breaking change.
This was requested by Russ so that we would match the flux spec.

docmerlin:
Closes #1757 
This is just a renaming of delay to offset for task options so it matches the flux spec.


task/options/options.go
```diff
@@ -106,11 +106,11 @@ func FromScript(script string) (Options, error) {
 		opt.Every = everyVal.Duration().Duration()
 	}
 
-	if delayVal, ok := optObject.Get("delay"); ok {
+	if delayVal, ok := optObject.Get("offset"); ok {
```
mark-rushakoff: Rename local variable to `offsetVal`?

task/options/options.go
```diff
@@ -165,7 +165,7 @@ func (o *Options) Validate() error {
 		}
 	}
 
-	if o.Delay.Truncate(time.Second) != o.Delay {
+	if o.Offset.Truncate(time.Second) != o.Offset {
 		// For now, allowing negative delays. Maybe they're useful for forecasting?
 		errs = append(errs, "delay option must be expressible as whole seconds")
```
mark-rushakoff: Should be `offset` option. May need to take a closer look at other error messages in this file.

task/servicetest/servicetest.go
```diff
@@ -158,8 +158,8 @@ func testTaskCRUD(t *testing.T, sys *System) {
 		if f.Cron != "* * * * *" {
 			t.Fatalf(`%s: wrong cron returned; want "* * * * *", got %q`, fn, f.Cron)
 		}
-		if f.Delay != "5s" {
-			t.Fatalf(`%s: wrong delay returned; want "5s", got %q`, fn, f.Delay)
+		if f.Offset != "5s" {
+			t.Fatalf(`%s: wrong delay returned; want "5s", got %q`, fn, f.Offset)
```
mark-rushakoff: wrong `offset`

PR #1763, opened 2018-12-06
Add ability to file upload and call write api

### LINKED ISSUE #1482: add a card for configuring line protocol uploads
as a new user
when i select the line protocol option for loading data
then i would like a card to configure the line protocol file

Palakp41:
Co-authored-by: Palak Bhojani <palak@influxdata.com>

Closes #1482 

Add the ability to upload file


ui/src/onboarding/components/configureStep/lineProtocol/LineProtocolTabs.tsx
```diff
@@ -120,6 +125,14 @@ export class LineProtocolTabs extends PureComponent<Props> {
     const {setLineProtocolText} = this.props
     setLineProtocolText(e.target.value)
   }
+
+  private handleFileUpload = async (uploadContent: string): Promise<any> => {
```
bthesorceror: I think this should be a `promise<void>`, also what do you think about moving this api call to an action as this is already a connected component?

PR #1765, opened 2018-12-06
feat(task): Allow the most recent run time to be shown in the api.

### LINKED ISSUE #1705: task: include timestamp of latest completed run when listing tasks
This is to support https://github.com/influxdata/platform/issues/1474.

The swagger doc needs to be updated to mention the new field. I think we can expose it as `latestCompleted`, formatted as an RFC3339 timestamp matching the now/scheduledFor of a run, which is actually stored as the task meta object's LatestCompleted field. The description in swagger will need to make it clear that this is the latest matching schedule, not the timestamp of when the run occurred.

lyondhill:
When showing tasks in the api latest_completed will now show in the api if it has been run.

fixes #1705


http/swagger.yml
```diff
@@ -3954,6 +3954,10 @@ components:
         offset:
           description: Duration to delay after the schedule, before executing the task; parsed from flux.
           type: string
+        latest_completed:
+          description: Time task was last run, RFC3339.
```
mark-rushakoff: I think this needs `readOnly: true`.

The description should be a little more clear that this is the latest completed run. How about "Timestamp of latest scheduled, completed run"?

task/platform_adapter.go
```diff
@@ -235,5 +235,8 @@ func toPlatformTask(t backend.StoreTask, m *backend.StoreTaskMeta) (*platform.Ta
 	if m != nil {
 		pt.Status = string(m.Status)
 	}
+	if m.LatestCompleted != 0 {
```
mark-rushakoff: Probably don't need the `if` here. LatestCompleted should always be set on meta creation.

lyondhill: the if was under the condition that the task has never been run. But before using m I do need to check if m exists :)

lyondhill: Thanks. I see that now :)

task/servicetest/servicetest.go
```diff
@@ -225,6 +230,44 @@ func testTaskCRUD(t *testing.T, sys *System) {
...
+	if err := sys.ts.CreateTask(sys.Ctx, task); err != nil {
+		t.Fatal(err)
+	}
+
+	st, err := sys.ts.FindTaskByID(sys.Ctx, task.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if st.LatestCompleted == "" {
```
mark-rushakoff: What do you think about parsing the LatestCompleted and asserting that it's within, say, +/- 10 seconds of when we created the task?

task/servicetest/servicetest.go
```diff
@@ -225,6 +230,49 @@ func testTaskCRUD(t *testing.T, sys *System) {
...
+		t.Fatal(err)
+	}
+
+	ti, err := time.Parse(time.RFC3339, st.LatestCompleted)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if time.Now().Sub(ti) > 10*time.Second {
+		t.Fatal("latest created not accurate")
```
mark-rushakoff: Include the current and parsed time here so that if it fails, we have detail on how wrong it was.

task/servicetest/servicetest.go
```diff
@@ -225,6 +230,49 @@ func testTaskCRUD(t *testing.T, sys *System) {
...
+	if err := sys.S.FinishRun(sys.Ctx, task.ID, rc.Created.RunID); err != nil {
+		t.Fatal(err)
+	}
+
+	st2, err := sys.ts.FindTaskByID(sys.Ctx, task.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if st.LatestCompleted == st2.LatestCompleted {
```
mark-rushakoff: RFC3339 strings should be lexicographically comparable, so this comparison can be `if st2.LatestCompleted <= st.LatestCompleted`. Include the values in the error message here too.

task/servicetest/servicetest.go
```diff
@@ -268,8 +269,8 @@ func testMetaUpdate(t *testing.T, sys *System) {
 		t.Fatal(err)
 	}
 
-	if st.LatestCompleted == st2.LatestCompleted {
-		t.Fatal("non started task has a latest completed")
+	if st2.LatestCompleted <= st.LatestCompleted {
+		t.Fatalf("executed task has not updated latest complete: expected !%s, got %s", st.LatestCompleted, st2.LatestCompleted)
```
mark-rushakoff: `expected > %s` probably?

lyondhill: yah probably. I dont need to show both if I do it your way

PR #1777, opened 2018-12-06
Add query building to query builder

### LINKED ISSUE #1685: Selections in the query builder should build a editable and submittable query
Depends on #1683.

- Selections in the visual query builder should build a query behind the scenes that can be submitted to view data
- ~~A user should be able edit this query as text (Flux or InfluxQL)~~ #1775
- ~~If the user selects a function not supported in InfluxQL, the ability to edit the query as InfluxQL should be disabled~~ #1775
- The user should need to press the submit button explicitly
- The submit button should be disabled if the query is not fully built (e.g. no bucket selected)

chnn:
Closes #1685

builder


ui/src/shared/components/SubmitQueryButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+        text="Submit"
+        size={ComponentSize.Small}
+        status={this.buttonStatus}
+        onClick={this.handleClick}
+        color={ComponentColor.Primary}
+      />
+    )
+  }
+
+  private get buttonStatus(): ComponentStatus {
```
AlirieGray: nice

ui/src/shared/utils/queryBuilder.test.ts
```diff
@@ -0,0 +1,87 @@
+import {buildQuery} from 'src/shared/utils/queryBuilder'
+
+import {BuilderConfig} from 'src/types/v2'
+
+describe('buildQuery', () => {
```
AlirieGray: yay tests!

ui/src/shared/utils/queryBuilder.ts
```diff
@@ -0,0 +1,89 @@
+import {BuilderConfig} from 'src/types/v2'
+import {FUNCTIONS} from 'src/shared/constants/queryBuilder'
+
+const DEFAULT_WINDOW_INTERVAL = '10s'
+
+const WINDOW_INTERVALS = {
```
AlirieGray: what does this do?

chnn: It's a mapping of time ranges to what interval we should use when viewing that time range.

So like if we are viewing 1 hour of data, we should see it aggregated by 1 minute intervals.

It's kind of a temporary thing... I think we'll eventually use `autoInterval` or something like that.

PR #1780, opened 2018-12-07
ci: reset go module cache

mark-rushakoff:
This should make the gotest stage on Circle noticeably faster. We should
do this every once in a while. I used YYYYMMDD format for the
cachebuster in case we need to reset more than once in a single month.

The arbitrary builds mentioned in the comment are:

https://circleci.com/gh/influxdata/platform/8811
https://circleci.com/gh/influxdata/platform/4701

The docs for cachebusting are:

https://circleci.com/docs/2.0/caching/#clearing-cache


mark-rushakoff:
In https://circleci.com/gh/influxdata/platform/8814 (for this PR), the module cache restore took 10s and the save took 39s in https://circleci.com/gh/influxdata/platform/8813 (also this PR); down from a 39s restore and 68s save in https://circleci.com/gh/influxdata/platform/8811 (a separate PR using the previous, bloated module cache). So, the new total is 49s, down from 107s.

BondAnthony:
Hey @mark-rushakoff the numbers look really good. However, does introducing a date now require a maintenance task for the team? Could we possibly change our key names to use a generated hash? Maybe something like `platform-gomod-{{ checksum go.sum }}`, thoughts?

mark-rushakoff:
> However, does introducing a date now require a maintenance task for the team?

Kind of. When module caching in Circle takes noticeably long and you want to reset it, you would update the date to clear the old cache. So it could on demand / as needed, or we could make a habit of doing it on a schedule.

> Could we possibly change our key names to use a generated checksum?

That would clear the entire cache any time we change any dependency. The upside is, the module cache would contain precisely the bare minimum set of modules we need, so save and restore times should be optimal. The downside is, there's a higher cost to warm the cache from empty whenever a dependency changes.

But since dependency changes are infrequent compared to other activity, `platform-gomod-{{ checksum go.sum }}` as you've described seems like a good fit. Want to make a PR for that to supersede this change?

mark-rushakoff:
The cleared cache _should_ only affect the first CI job for a given go.sum.

Come to think of it, I'm not sure whether we need to consider go.mod too, or if just go.sum will suffice.  I didn't get a clear answer skimming through `go help modules`.

Thanks for picking up the work.

PR #1781, opened 2018-12-07
Feat/filter view type

### LINKED ISSUE #1003: Filter /api/v2/views by view type
As a follow up to #918 the logs page needs an endpoint that returns some log page struct that contains a single view of type `log-viewer`. Making a `GET /api/v2/log` could either find or create the log page with a [default log viewer view](https://github.com/influxdata/chronograf/blob/2f2263d7acbbac16e0cc4d5ef6e48775678993c5/bolt/org_config.go#L90-L234) similar to the `org_config` endpoint in chronograf.

jademcgough:
Closes #1003 

Allows users to pass an optional type param to `GET /views`.

I'm not a pro at working with go interfaces, so lmk if there's a cleaner way to do this.


bolt/view.go
```diff
@@ -108,14 +109,25 @@ func (c *Client) FindView(ctx context.Context, filter platform.ViewFilter) (*pla
...
-		return func(d *platform.View) bool {
-			return d.ID == *filter.ID
+		return func(v *platform.View) bool {
+			return v.ID == *filter.ID
 		}
 	}
 
-	return func(d *platform.View) bool { return true }
+	if len(filter.Types) > 0 {
+		var sm sync.Map
```
kelwang: Just one question, do you think a regular loop is good enough? sync.Map only benefit when using goroutines with heavier store and load.

jademcgough: A regular map, you mean? Yeah, I think that'd be better in terms of performance, I just wasn't sure whether it'd be concurrency safe.

kelwang: 👍

desa: I think a regular map should work, but either is fine.

http/view_service.go
```diff
@@ -104,6 +106,20 @@ func (h *ViewHandler) handleGetViews(w http.ResponseWriter, r *http.Request) {
...
+type getViewsRequest struct {
+	filter platform.ViewFilter
+}
+
+func decodeGetViewsRequest(ctx context.Context, r *http.Request) *getViewsRequest {
+	qp := r.URL.Query()
+
+	return &getViewsRequest{
+		filter: platform.ViewFilter{
+			Types: qp["type"],
```
desa: 👍

view.go
```diff
@@ -70,13 +71,16 @@ type ViewContents struct {
 // ViewProperties is used to mark other structures as conforming to a View.
 type ViewProperties interface {
 	viewProperties()
+	GetType() string
```
desa: Is there a reason why we don't just call this `Type`?

jademcgough: A lot of these already have a `Type` field, so the compiler won't allow it.

PR #1789, opened 2018-12-07
Allow TSI cache to be dynamically altered

e-dard:

This PR allows the TSI cache (which caches bitsets of series IDs for key/value pairs) to have its size altered.

_What was the problem?_

The cache size was fixed at 100.

_What was the solution?_

Make it a configuration variable and increase the default size.


e-dard:
@stuartcarnie AFAIK we don't yet have a config sample.

PR #1791, opened 2018-12-07
Feat/auth description

### LINKED ISSUE #1769: Authorizations API:  Add a `name` and `description` field to Authorization.

121watts:
Closes #1769 

Adding a string description for tokens

_What was the problem?_
There is no human readable way to distinguish tokens

_What was the solution?_
Add an description field!


goller:
@121watts you'll need to update the swagger doc as well.

PR #1792, opened 2018-12-07
chore(ci): Reset CircleCI cache logic

BondAnthony:
Update how CircleCI saves cached data. Since dependencies don't change often builds will reuse the same cache based on the checksum of go.sum. When dependencies update the checksum will change resulting in a new cache rebuild. This first rebuild will take a few seconds longer but all additional builds will speed up.

Additional discussion around this can be found in PR #1780.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,9 +44,7 @@ jobs:
       - restore_cache:
           name: Restoring GOPATH/pkg/mod
           keys:
-            - platform-gomod-{{ .Branch }}-{{ .Revision }} # Matches when retrying a single run.
-            - platform-gomod-{{ .Branch }}-                # Matches a new commit on an existing branch.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           when: always
       - save_cache:
           name: Saving GOPATH/pkg/mod
-          key: platform-gomod-{{ .Branch }}-{{ .Revision }}
+          key: platform-gomod-{{ checksum "go.sum" }}-{{ .Branch }}-{{ .Revision }}
```
mark-rushakoff: Since the module cache is intended to be a function of the content of go.sum, I think we can probably drop branch and revision here - which means restoring the cache can also drop branch and revision.

BondAnthony:
Build passed and cache response times look good.

Issue #1503, opened 2018-11-20
add the ability to name the new query


russorat:
as a query builder
with admin permissions or higher
when i use the explorer tool to build out queries
then i would like the ability to name each query something i choose

PR #1794, opened 2018-12-07
feat(ui): add name to dashboard query

OfTheDelmer:
Closes https://github.com/influxdata/platform/issues/1503

Add query name and editing field
_What was the problem?_
A user should be able to specify a name that refers to a query
_What was the solution?_
Update the TimeMachineTab name to be editable using either double click or right click.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/types/v2/dashboards.ts
```diff
@@ -100,19 +101,23 @@ export interface ViewLinks {
...
   | TableView
   | GaugeView
-
-export type ViewProperties =
-  | QueryViewProperties
   | MarkdownView
   | EmptyView
   | LogViewerView
 
+export type QueryViewProperties = Extract<
```
OfTheDelmer: [Conditional Typing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)

```
Extract<T, U> – Extract from T those types that are assignable to U
```

Selects only `ViewProperties` with queries.

ischolten: woaaa

OfTheDelmer:
I can open a separate issues to discuss two things that came up:

- Add a var to disable/enable right click actions (so we don't gotta have right click inspect in dev mode)
- Development assertions instead of throwing errors in clockface components

ui/src/App.tsx
```diff
@@ -10,6 +11,7 @@ interface Props {
 const App: SFC<Props> = ({children}) => (
   <div className="chronograf-root">
     <Notifications />
+    <RightClickLayer />
```
ischolten: cool

PR #1798, opened 2018-12-09
doc(http): delete a bucket

### LINKED ISSUE #1636: Verify swagger correctly represents DELETE on buckets
The implementation allows a delete against a bucket resource - but I think we need to doc the DELETE in swagger: https://github.com/influxdata/platform/blob/master/http/swagger.yml#L1911

/cc @e-dard

zhulongcheng:
Closes #1636

  - [x] Rebased/mergeable
  - [x] Tests pass


jademcgough:
❤️

PR #1799, opened 2018-12-09
feat(http/paging): add paging implementation

zhulongcheng:
https://github.com/influxdata/platform/blob/master/http/swagger.yml#L3809-L3827
https://github.com/influxdata/platform/blob/master/http/bucket_service.go#L207-L213
https://github.com/influxdata/platform/blob/master/http/auth_service.go#L66-L72

_What was the problem?_
Paging has been defined in swagger, but it is not implemented in http server.

_What was the solution?_
Try to add an implementation:
1. Add `PagingLinks`
```
// PagingLinks represents paging links.
type PagingLinks struct {
    Prev string `json:"prev,omitempty"`
    Self string `json:"self"`
    Next string `json:"next,omitempty"`
}
```

2. Add `PagingFilter` interface
```
// PagingFilter represents a filter containing url query params.
type PagingFilter interface {
    // QueryParams converts `[Resource]Filter` fields to query params.
    QueryParams() map[string][]string
}
```

3. Add `decodeFindOptions` function
```
// decodeFindOptions returns a FindOptions decoded from http request.
func decodeFindOptions(ctx context.Context, r *http.Request) (*platform.FindOptions, error) 
```

4. Add `newPagingLinks` function
```
// newPagingLinks returns a PagingLinks.
// num is the number of returned results.
func newPagingLinks(basePath string, opts platform.FindOptions, f platform.PagingFilter, num int) *platform.PagingLinks
```

---

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/paging.go
+++ b/paging.go
@@ -4,6 +4,11 @@ import (
 	"strconv"
 )
 
+const (
+	DefaultPageSize = 20
+	MaxPageSize     = 100
+)
+
 // PagingFilter represents a filter containing url query params.
 type PagingFilter interface {
 	// QueryParams returns a map containing url query params.
--- a/paging.go
+++ b/paging.go
@@ -17,9 +17,9 @@ type PagingFilter interface {
 
 // PagingLinks represents paging links.
 type PagingLinks struct {
-	Prev map[string]string `json:"prev"`
-	Self map[string]string `json:"self"`
-	Next map[string]string `json:"next"`
+	Prev string `json:"prev"`
+	Self string `json:"self"`
+	Next string `json:"next"`
 }
 
 // FindOptions represents options passed to all find methods with multiple results.
--- a/paging.go
+++ b/paging.go
@@ -17,9 +17,9 @@ type PagingFilter interface {
 
 // PagingLinks represents paging links.
 type PagingLinks struct {
-	Prev string `json:"prev"`
+	Prev string `json:"prev,omitempty"`
 	Self string `json:"self"`
-	Next string `json:"next"`
+	Next string `json:"next,omitempty"`
 }
 
 // FindOptions represents options passed to all find methods with multiple results.
--- a/paging.go
+++ b/paging.go
@@ -1,0 +1,38 @@
+package platform
+
+import (
+	"strconv"
+)
+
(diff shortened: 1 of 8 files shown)


zhulongcheng:
@desa @goller @kelwang  request some suggestions, thanks.

paging.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// PagingFilter represents a filter containing url query params.
+type PagingFilter interface {
+	// QueryParams returns a map containing url query params.
+	QueryParams() map[string][]string
+}
+
+// PagingLinks represents paging links.
+type PagingLinks struct {
```
desa: What's the reasoning behind making each of these value a map?

zhulongcheng: Thanks for reply @desa . 
Links are embedded in a map to match [Links in swagger](https://github.com/influxdata/platform/blob/master/http/swagger.yml#L3809-L3827) . 
But like this [comment](https://github.com/influxdata/platform/blob/master/http/task_service.go#L115-L136), links are flat in http service.
So, embedded or flat?

desa: @zhulongcheng thanks for pointing that out. After discussing it internally, we decided that we'd like to see a flat structure. Would you mind making the changes necessary for that.

Great work by the way!

zhulongcheng: I also prefer a flat structure. 👍

paging.go
```diff
@@ -17,9 +17,9 @@ type PagingFilter interface {
 
 // PagingLinks represents paging links.
 type PagingLinks struct {
-	Prev map[string]string `json:"prev"`
-	Self map[string]string `json:"self"`
-	Next map[string]string `json:"next"`
+	Prev string `json:"prev"`
+	Self string `json:"self"`
+	Next string `json:"next"`
```
desa: What do you think about making these `omitempty`?

zhulongcheng: `self` is required,  `prev` and `next` to be optional ?

desa: Yeah I think so

zhulongcheng:
@desa  request a review again, thanks.

zhulongcheng:
😀

PR #1813, opened 2018-12-10
Feat/filter token

### LINKED ISSUE #1439: add the ability to filter available tokens
as a user
when i view the list of available tokens
then i would like to be able to filter by the name or org of the token

121watts:
Closes #1439 

Filters tokens by description or status


ui/src/me/components/account/Tokens.tsx
```diff
@@ -18,30 +19,66 @@ interface StateProps {
   authorizationsLink: string
 }
 
-export class Tokens extends PureComponent<StateProps> {
+interface State {
+  searchTerm: string
+}
+
+enum AuthSearchKeys {
```
Palakp41: What if this was under types folder like all the other enums?

121watts: These are local to the scope of this component.  So, I think they are best here for now

PR #1818, opened 2018-12-10
Feat/selection step type next

### LINKED ISSUE #1657: dataLoaders: clicking on the type in the selection step should take you to the next "step" (configure for csv and line protocl but streaming selection for streaming)
as a new user
when i click on the source for loading data
then i should immediately jump to the next page

see also: https://github.com/influxdata/platform/issues/1656

Palakp41:
Closes #1657 

Add the call to handle next when the type on selection step is selected.

### DIFF
--- a/ui/src/onboarding/components/selectionStep/SelectDataSourceStep.tsx
+++ b/ui/src/onboarding/components/selectionStep/SelectDataSourceStep.tsx
@@ -156,8 +156,12 @@ class SelectDataSourceStep extends PureComponent<Props, State> {
     this.props.onDecrementCurrentStepIndex()
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/onboarding/components/configureStep/lineProtocol/LineProtocolTabs.tsx
```diff
@@ -108,76 +116,82 @@ export class LineProtocolTabs extends PureComponent<Props, State> {
...
-              placeholder="http://..."
-              widthPixels={700}
-              value={urlInput}
-              onChange={this.handleURLChange}
-              autoFocus={true}
-            />
-          </Form.Element>
-        </Form>
-      )
+    switch (activeLPTab) {
```
121watts: Very nice use of the `switch` here

PR #1825, opened 2018-12-10
Data loader/remove sidebar buttons selection step

Palakp41:
Closes 1652

Create the condition on sidebar to display buttons only when on configuration step.


http/cur_swagger.yml
```diff
@@ -3972,6 +3964,14 @@ components:
           description: err is a stack of errors that occurred during processing of the request. Useful for debugging.
           type: string
       required: [code, message, op, err]
+    WritePrecision:
```
ischolten: noice!

PR #1857, opened 2018-12-12
fix(testing): compare expected error messages against actual

### LINKED ISSUE #1856: tests are not comparing actual vs expected error messages
Our `diffPlatformErrors` util function is only checking error codes and ops against expectations in tests. Several existing tests specify an expected error message, but this is not currently being checked in our tests.

jademcgough:
Closes #1856


kelwang:
thanks

Issue #1832, opened 2018-12-11
tsdb/tsi1 tests failing in CI


wolffcm:
After I merged just now, CI started failed on master in the `make go-test-race` step.  After I reran, it passed.

https://circleci.com/gh/influxdata/platform/9163?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

The failure seems to be happening here:
```
=== RUN   TestIndex_MeasurementCardinalityStats
=== RUN   TestIndex_MeasurementCardinalityStats/Empty
=== RUN   TestIndex_MeasurementCardinalityStats/Simple
=== RUN   TestIndex_MeasurementCardinalityStats/SimpleWithDelete
=== RUN   TestIndex_MeasurementCardinalityStats/Large
FAIL	github.com/influxdata/platform/tsdb/tsi1	97.896s
```
This doesn't seem to be related to my change though, since it also failed on a feature branch similarly:
https://circleci.com/gh/influxdata/platform/9074?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mark-rushakoff:
My guess at this point would be that an individual test's main goroutine completed successfully, then printed `PASS`, and then a separate goroutine for that test called t.Error, which caused that test to "retroactively" fail and caused the suite to fail.

mark-rushakoff:
Happened again at https://circleci.com/gh/influxdata/platform/9179. I'm going to take a couple minutes to try and repro locally.

mark-rushakoff:
I ran `while go test -count=1 ./tsdb/tsi1/ -v -race -p=8; do date; done` for about 30 minutes and wasn't able to reproduce the error on my machine.

I was wrong about the separate goroutine failing the test after it completed; when that happens, there's a `panic: Fail in goroutine after TestFoo has completed`. So at this point, I'm really unsure what's going on with this failure.

mark-rushakoff:
Another failure: https://circleci.com/gh/influxdata/platform/9341

Thinking about using `go test -json` in CI to see if that somehow captures more failure detail.

It's really difficult to find that failure in the logs, because it doesn't match `FAIL:`; when you just search for "fail" without the colon, there's 132 case-insensitive matches in the link above, from other test names and test log output.

mark-rushakoff:
[...]

PR #1871, opened 2018-12-12
ci: use -p=4 on circle

mark-rushakoff:
The docker container we're using has 2 vCPUs. -p=8 was working fine for
a while, but diagnosing #1832 looks like the tsi1.test process was
likely getting OOM-killed. Reducing from -p=8 to -p=4 ought to reduce
memory pressure enough to avoid OOM kills, hopefully.

Also remove a stale TODO.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - image: circleci/golang:1.11
     environment:
       GOCACHE: /tmp/go-cache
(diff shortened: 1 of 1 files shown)


mark-rushakoff:
This change ran go test -race 4 times in a row successfully. (The first build failed due to an error I made in the circle config file.)

It seems to make a negligible difference in CI duration.

PR #1890, opened 2018-12-12
Introduce Clockface Labels Components

alexpaxton:
### Overview

This PR introduces 3 presentational components intended for use in building out the Labels feature. The components are atomic and context independent.

#### Label

screen shot 2018-12-12 at 10 40 26 am

Displays a single label.

```
import {Label} from 'src/clockface'

<Label />
```

| Prop | Type | Notes |
|-------------|------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| id | `string` |  |
| text | `string` | Displayed on the label |
| description | `string` | Displayed as the `title` attribute if `onClick` and `onDelete` are not specified |
| colorHex | `string` | Will throw an error if an invalid value is passed in. Must be 7 chars starting with # |
| onClick? | `(id: string) => void` | Intended for seeing a list of all resources sharing the label. Allows the label to light up on hover and adds tooltip text: `Click to see all resources with the "text" label` |
| onDelete? | `(id: string) => void` | If present, the delete `X` button appears on the label. Intended for managing labels on a resource |
|size? | `ComponentSize` | Set to `ExtraSmall` by default |

### Label Container

For displaying collections of labels. Can optionally limit how many labels can appear before truncation via tooltip

screen shot 2018-12-12 at 10 47 45 am

```
import {Label} from 'src/clockface'

<Label.Container>
    {labels.map(label => <Label />)}
</Label.Container>
```

| Prop | Type | Notes |
|------------------|-------------------------------|-----------------------------------------------------------------------------------------------------|
| children | `JSX.Element | JSX.Element[]` | Must be of type `<Label />` |
| className? | `string` | Allows for a custom classname, useful when you want the container to have special positioning rules |
| limitChildCount? | `number` | Any children beyond this number will be displayed in a tooltip. Is set to `999` by default |

### Label Selector Widget

Intended for the user to assign labels to a resource in a quick way. Supports keyboard and mouse navigation. Makes use of a fuzzy-ish search so spelling errors don't prevent the user from finding their desired label. This component only has enough state to manage the interactions it offers, and most state is to be managed by the parent component that consumes `LabelSelector`.

labels-widget-fuzzy-search

```
import {LabelSelector, LabelType} from 'src/clockface'

<LabelSelector />
```

| Prop | Type | Notes |
|-------------------|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
| selectedLabels | `LabelType[]` | Tells the component which labels have already been selected. These will be subtracted from the suggestion list |
| allLabels | `LabelType[]` | List of all labels to filter against |
| onAddLabel | `(label: LabelType) => void` | Gets called when the user adds a label via this component. The entire label object is passed so that adding it to the selected array is easy |
| onRemoveLabel | `(label: LabelType) => void` | Gets called when the user clicks the delete button on an individual label. The entire label object is passed so that removing it from the selected list is easy |
| onRemoveAllLabels | `() => void` | Gets called when the user clicks the `Remove All Labels` button. Should clear out selected state |
| resourceType | `string` | Used in the component to refer to the type of resource the labels are being added to. In the dashboard example here I have passed in `this Dashboard` |
| inputSize? | `ComponentSize` | Controls the size of the input within the `LabelSelector` component |

### Adding Labels to a `IndexList` row

screen shot 2018-12-12 at 11 01 32 am

The limit of `3` is entirely arbitrary here, use your best judgement

```
<IndexList.Cell>
  <ComponentSpacer align={Alignment.Left} stackChildren={Stack.Rows}>
    <Link to={`/dashboards/${id}`}>{name}</Link>
    <LabelContainer limitChildCount={3}>
      {labels go here}
    </LabelContainer>
  </ComponentSpacer>
</IndexList.Cell>
```

### Checklist

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/clockface/components/label/LabelContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+      </div>
+    )
+  }
+
+  private get children(): JSX.Element[] {
+    const {children, limitChildCount} = this.props
+
+    return React.Children.map(children, (child: JSX.Element, i: number) => {
+      if (i < limitChildCount) {
```
ebb-tide: 👍

ui/src/clockface/components/label/LabelContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface Props {
+  children: JSX.Element | JSX.Element[]
+  className?: string
+  limitChildCount?: number
+}
+
+class LabelContainer extends Component<Props> {
+  public static defaultProps: Partial<Props> = {
+    limitChildCount: 999,
```
ebb-tide: so generous!

ui/src/clockface/components/label/LabelSelector.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const adjacentID = availableLabels[adjacentIndex].id
+
+    this.setState({highlightedID: adjacentID})
+  }
+
+  private handleKeyDown = (e: KeyboardEvent<HTMLInputElement>): void => {
+    const {highlightedID} = this.state
+
+    switch (e.key) {
+      case 'Escape':
```
ebb-tide: can these be in an enum please? I feel like they should already exist as part of react or type of e.key??

ui/src/clockface/components/label/LabelSelector.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const {allLabels, selectedLabels} = this.props
+    const filterValue = e.target.value
+
+    const availableLabels = _.differenceBy(
+      allLabels,
+      selectedLabels,
+      l => l.text
+    )
+
+    const filteredLabels = availableLabels.filter(label => {
```
ebb-tide: I feel like this function might exist in something search bar related  as well? Consider moving function to a utility or export the existing function?

ui/src/clockface/components/label/LabelSelectorMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return <div className="label-selector--empty">{this.emptyText}</div>
+  }
+
+  private get emptyText(): string {
+    const {allLabelsUsed} = this.props
+
+    if (allLabelsUsed) {
+      return 'You have somehow managed to add all the labels, wow!'
```
ebb-tide: cute, but a little forward?  maybe "no labels left to add!"

ui/src/clockface/components/label/LabelContainer.tsx
alexpaxton: =D

ui/src/clockface/components/label/LabelSelector.tsx
alexpaxton: Just found this guy: https://www.npmjs.com/package/ts-keycode-enum

ui/src/clockface/components/label/LabelSelector.tsx
alexpaxton: that's not a bad idea. I am thinking of building out a similar component to this one as a general purpose input with suggestions that would exist in all the filter widgets

PR #1891, opened 2018-12-12
Add the ability to count down timer on listening for data

### LINKED ISSUE #1646: dataLoader: timeout should be counting down
as a new user
when i listen for data
then i would like the timer to show the seconds remaining

Palakp41:
Closes #1646 

Add new timer that is used for counting down every second while the request processes

### DIFF
--- a/ui/src/onboarding/components/verifyStep/DataListening.tsx
+++ b/ui/src/onboarding/components/verifyStep/DataListening.tsx
@@ -25,24 +25,38 @@ export interface Props {
 
 interface State {
   loading: RemoteDataState
+  timePassedInSeconds: number
+  secondsLeft: number
 }
 
 const MINUTE = 60000
-const WAIT = 5000
+const FETCH_WAIT = 5000
+const SECONDS = 60
+const TIMER_WAIT = 1000
 
 @ErrorHandling
 class DataListening extends PureComponent<Props, State> {
   private intervalID: NodeJS.Timer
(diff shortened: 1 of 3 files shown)


ui/src/onboarding/components/verifyStep/DataListening.tsx
```diff
@@ -91,6 +107,7 @@ class DataListening extends PureComponent<Props, State> {
   }
 
   private handleClick = (): void => {
```
OfTheDelmer: Do we need a guard here for `RemoteDataState.Error`?

ui/src/onboarding/components/verifyStep/DataListening.tsx
```diff
@@ -126,9 +143,24 @@ class DataListening extends PureComponent<Props, State> {
       this.setState({loading: RemoteDataState.Error})
       return
     }
-
     this.intervalID = setTimeout(this.checkForData, WAIT)
   }
+
+  private startTimer() {
```
OfTheDelmer: Before we start counting down we could set the initial state for the `timePassedInSeconds` and `secondsLeft`. This will prevent negative seconds after the `timePassed >= MINUTE`  and ensure we reach our terminal condition at `secondsLeft === 0`

Palakp41: Fixed it!

PR #1893, opened 2018-12-12
Feat(ui/dataLoaders): Change plugin selection to bundle selection

### LINKED ISSUE #1655: dataLoaders: replace list of streaming sources with list of streaming bundles
as a new user
when i view the list of streaming sources
then i would like to view the config bundles, not the sources.

the first one is:
- System (contains these plugins: cpu, disk, diskio, kernel, mem, processes, swap, system)
- more tbd

ischolten:
Closes #1655

Instead of individually adding plugins, users will select a bundle of plugins.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/mocks/dummyData.ts
+++ b/ui/mocks/dummyData.ts
@@ -12,7 +12,17 @@ import {Links} from 'src/types/v2/links'
 import {Task, TaskStatus} from 'src/types/v2/tasks'
 import {OnboardingStepProps} from 'src/onboarding/containers/OnboardingWizard'
 import {ConfigurationState} from 'src/types/v2/dataLoaders'
-import {TelegrafPluginInputCpu, TelegrafPluginInputRedis} from 'src/api'
+import {
+  TelegrafPluginInputCpu,
+  TelegrafPluginInputRedis,
+  TelegrafPluginInputDisk,
+  TelegrafPluginInputKernel,
+  TelegrafPluginInputDiskio,
+  TelegrafPluginInputMem,
+  TelegrafPluginInputSwap,
+  TelegrafPluginInputSystem,
+  TelegrafPluginInputProcesses,
+} from 'src/api'
 
 export const links: Links = {
   authorizations: '/api/v2/authorizations',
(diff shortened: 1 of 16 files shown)


ui/mocks/dummyData.ts
```diff
@@ -289,7 +299,41 @@ export const cpuPlugin = {
 export const telegrafPlugin = {
   name: TelegrafPluginInputCpu.NameEnum.Cpu,
   configured: ConfigurationState.Unconfigured,
-  active: true,
+  active: false,
+}
+
+export const cpuTelegrafPlugin = {...telegrafPlugin}
```
AlirieGray: nit: what do you think about spacing out these constants?

ui/mocks/dummyData.ts
```diff
@@ -348,6 +392,10 @@ export const telegrafConfigsResponse = {
   request: {},
 }
 
+export const createTelegrafConfigResponse = {
+  data: telegrafConfig,
```
AlirieGray: just to see if I'm understanding correctly, this is referring to `telegrafConfig` above which has name `TelegrafPluginInputCpu.NameEnum.Cpu`? is there a reason?

ischolten: correct; its just dummy data. i just want the output to be the the config that i created

ui/mocks/dummyData.ts
ischolten: can do

ui/src/onboarding/utils/pluginConfigs.ts
```diff
@@ -27,3 +31,30 @@ export const updateConfigFields = <T extends Plugin>(
 export const createNewPlugin = (name: TelegrafPluginName): Plugin => {
   return telegrafPluginsInfo[name].defaults
 }
+
+export const isPluginUniqueToBundle = (
```
AlirieGray: nice

ui/mocks/dummyData.ts
AlirieGray: oh I see, I should have looked at the file name. now I feel silly :P

ischolten: all good :)

PR #1901, opened 2018-12-12
Pass source id instead of source to resolvedSourceID

ebb-tide:


ui/src/shared/containers/SetActiveSource.tsx
```diff
@@ -52,14 +52,14 @@ class SetActiveSource extends PureComponent<Props> {
 
     const querySourceID = readQueryParams().sourceID
 
-    let resolvedSourceID
+    let resolvedSourceID: string
```
alexpaxton: TYPES! Huzzah

ui/src/shared/containers/SetActiveSource.tsx
```diff
@@ -52,14 +52,14 @@ class SetActiveSource extends PureComponent<Props> {
...
 
     if (sources.find(s => s.id === activeSourceID)) {
       resolvedSourceID = activeSourceID
     } else if (sources.find(s => s.id === querySourceID)) {
       resolvedSourceID = querySourceID
     } else if (sources.length) {
-      resolvedSourceID = sources[0]
+      resolvedSourceID = sources[0].id
```
alexpaxton: Does this need a guard?

ui/src/shared/containers/SetActiveSource.tsx
ebb-tide: I KNOWWW! :)

ui/src/shared/containers/SetActiveSource.tsx
ebb-tide: good idea, thanks

PR #1909, opened 2018-12-12
Switch buckets ui to use generated client

bthesorceror:
- [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/logs/actions/index.ts
```diff
@@ -334,17 +332,13 @@ export const getSourceAndPopulateBucketsAsync = (id: string) => async (
 ): Promise<void> => {
   const source = await readSource(id)
 
-  const bucketsLink = getDeep<string | null>(source, 'links.buckets', null)
```
OfTheDelmer: After it fetches a source the links include associated buckets of the following form:

```
/api/v2/sources/{sourceID}/buckets
```

Here getBuckets is fetching source buckets. I don't believe these can be retrieved using `/api/v2/buckets`

PR #1917, opened 2018-12-13
Fix(ui/dataLoaders): Hide download button if there's nothing to download

### LINKED ISSUE #1889: Download config should not be available until config is created
as a new user
when i am in the configure step 
then i should not see the download config button if there is no telegraf config created

ischolten:
Closes #1889 

_What was the problem?_
The download button was visible, even if there was nothing to download
_What was the solution?_
Dont show the download button if there's nothing to download

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/onboarding/components/OnboardingSideBar.test.tsx
+++ b/ui/src/onboarding/components/OnboardingSideBar.test.tsx
@@ -4,16 +4,22 @@ import {shallow} from 'enzyme'
 
 // Components
 import OnboardingSideBar from 'src/onboarding/components/OnboardingSideBar'
-import {cpuPlugin, influxDB2Plugin} from 'mocks/dummyData'
+import {cpuTelegrafPlugin, diskTelegrafPlugin} from 'mocks/dummyData'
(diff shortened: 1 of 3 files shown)


ui/src/onboarding/components/OnboardingSideBar.test.tsx
```diff
@@ -4,16 +4,22 @@ import {shallow} from 'enzyme'
 
 // Components
 import OnboardingSideBar from 'src/onboarding/components/OnboardingSideBar'
-import {cpuPlugin, influxDB2Plugin} from 'mocks/dummyData'
+import {cpuTelegrafPlugin, diskTelegrafPlugin} from 'mocks/dummyData'
+import {Button} from 'src/clockface'
+import SideBarTab from './side_bar/SideBarTab'
```
ebb-tide: can we define the full path here?

ischolten: oops yep. oh that auto import gets me every time

PR #1922, opened 2018-12-13
tsm1: tombstone performance

zeebo:
# Summary

`DeleteRange` would start at the first key in the index and do a linear walk until it reached the first key to be deleted. This is obviously bad when attempting to delete a large key. Instead, do a seek to the first key to be deleted.

## Performance changes

I did very short benchmarks, so lots of noise, but:

```
BenchmarkIndirectIndex_DeleteRangeLast-8     310663434       1550          -100.00%
BenchmarkIndirectIndex_DeleteRangeFull-8     18776033480     773649708     -95.88%
```

`DeleteRangeFull` is the closest thing to what retention does right now.

### DIFF
--- a/tsdb/tsm1/reader.go
+++ b/tsdb/tsm1/reader.go
@@ -1033,7 +1033,7 @@ func (d *indirectIndex) DeleteRange(keys [][]byte, minTime, maxTime int64) {
 	tombstones := map[string][]TimeRange{}
 	var ie []IndexEntry
 
-	for i := 0; len(keys) > 0 && i < d.KeyCount(); i++ {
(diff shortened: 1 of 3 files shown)


tsdb/tsm1/reader.go
```diff
@@ -1033,7 +1033,7 @@ func (d *indirectIndex) DeleteRange(keys [][]byte, minTime, maxTime int64) {
 	tombstones := map[string][]TimeRange{}
 	var ie []IndexEntry
 
-	for i := 0; len(keys) > 0 && i < d.KeyCount(); i++ {
+	for i := d.searchOffset(keys[0]); len(keys) > 0 && i < d.KeyCount(); i++ {
```
e-dard: Haha this is so obvious 😂.

PR #1925, opened 2018-12-13
Add ready checking into swagger and update the doc to match health check

### LINKED ISSUE #1906: document ready endpoint in swagger

lyondhill:
closes #1906


http/swagger.yml
```diff
@@ -723,6 +723,30 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /ready:
+    get:
+      tags:
+        - Ready
+      summary: Get the readiness of an instance
```
goller: would you describe the differences between ready and health here?

http/swagger.yml
```diff
@@ -6006,6 +6036,22 @@ components:
           type: string
       required:
         - password
+    Ready:
```
goller: Because ready and health follow the same schema, I'd say use just one and rename it to "Check" or something.

http/swagger.yml
```diff
@@ -6006,6 +6036,22 @@ components:
...
+      type: object
+      properties:
+        name:
```
goller: I think name and status should be required fields.  Do you think message should be?

PR #1929, opened 2018-12-13
Feat/nav logout link

### LINKED ISSUE #1772: add ability to log out under the my profile icon
as a user
with viewer permission or higher
when i am using the application
then i would like to be able to log out from anywhere in the app

similar to other apps, let's add a sign out option under the my profile section of the left menu

alexpaxton:
Closes #1772 

logout-link

Refactored the navbar to be truly composable and support sub-menu items.
Created a logout route `/logout` and a corresponding page component
Replaced the existing logout button with a link to `/logout`

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/Logout.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+// Components
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+type Props = WithRouterProps
+
+@ErrorHandling
+export class Logout extends PureComponent<Props> {
+  public async componentDidMount() {
+    this.handleSignOut()
```
OfTheDelmer: does this need to await?

ui/src/Logout.tsx
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  public render() {
+    return null
+  }
+
+  private handleSignOut = async () => {
+    await logout()
```
OfTheDelmer: What happens if they are offline?

ui/src/pageLayout/components/NavMenuItem.tsx
```diff
@@ -25,12 +33,16 @@ const NavMenuItem: SFC<Props> = ({
...
     <div className={classnames('nav--item', {active: isActive})}>
       <Link className="nav--item-icon" to={link}>
-        {children}
+        <span className={`icon sidebar--icon ${icon}`} />
       </Link>
       <div className="nav--item-menu">
         <Link className="nav--item-header" to={link}>
           {title}
         </Link>
+        {React.Children.map(
```
OfTheDelmer: We could utilize the `Select` component here

```jsx
<Select type={NavMenuSubItem}>
   {this.props.children}
</Select>
```

ui/src/Logout.tsx
alexpaxton: They get redirected back to signin no matter what

ui/src/pageLayout/components/NavMenuItem.tsx
alexpaxton: Good call!

PR #1943, opened 2018-12-14
Handle setting step status when skipping a step

AlirieGray:
Handle setting step status for the select and configure data steps when the user skips the step

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/onboarding/components/selectionStep/SelectDataSourceStep.tsx
```diff
@@ -195,6 +187,22 @@ class SelectDataSourceStep extends PureComponent<Props, State> {
     this.props.onAddPluginBundle(bundle)
   }
 
+  private handleSetStepStatus = () => {
+    const {
+      handleSetStepStatus,
```
chnn: confusing that there's a prop and method named `handleSetStepStatus`. maybe the prop could be called `onSetStepStatus`?

AlirieGray: can do 👍

PR #1961, opened 2018-12-14
feat(api/health): Add a standardized health and ready check system

lyondhill:
Build a system that can be used for both ready and health checking.
It can be used directly as a http.Handler and can be given a passthrough handler,
allowing it to operate as middleware.

### DIFF
--- a/kit/check/check.go
+++ b/kit/check/check.go
@@ -1,0 +1,208 @@
+// Package check standardizes /health and /ready endpoints.
+// This allows you to easily know when your server is ready and healthy.
+package check
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sort"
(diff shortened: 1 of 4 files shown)


kit/check/check.go
```diff
@@ -0,0 +1,208 @@
...
+	switch r.URL.Path {
+	case "/ready":
+		resp = c.CheckReady(r.Context())
+	case "/health":
+		resp = c.CheckHealth(r.Context())
+	}
+
+	// Check for a manual override state change.
+	query := r.URL.Query()
+	switch query.Get("force") {
```
goller: Because force can change the state of the platform, I think that this should not be in this handler.  I think it could be a middleware that could be reused by other programs using kit.

lyondhill: A middleware would be much more appropriate.

PR #1980, opened 2018-12-14
Add delete button to token index row

### LINKED ISSUE #1427: add the ability to delete a token from the token's list
as a platform user
when i view the My Account -> Tokens page
then i would like the ability to delete a token from that page

user should see a confirmation

AlirieGray:
Closes #1427

Adds a delete button to the `TokenRow` component that allows the user to delete a token. This works by optimistically removing the token from the state of the `TokensList` component, and then making a call to remove it from the backend. The user will be notified if the deletion was successful or unsuccessful.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/me/components/account/__snapshots__/Account.test.tsx.snap
+++ b/ui/src/me/components/account/__snapshots__/Account.test.tsx.snap
@@ -38,7 +38,7 @@ exports[`Account rendering renders! 1`] = `
         title="Tokens"
         url="tokens"
       >
-        <Tokens />
+        <Connect(Tokens) />
       </ProfilePageSection>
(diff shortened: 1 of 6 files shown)


ui/src/me/components/account/__snapshots__/Account.test.tsx.snap
```diff
@@ -38,7 +38,7 @@ exports[`Account rendering renders! 1`] = `
         title="Tokens"
         url="tokens"
       >
-        <Tokens />
+        <Connect(Tokens) />
```
Palakp41: nice!

PR #1988, opened 2018-12-17
Write a write service to support influxdb1

gianarb:
Currently we don't have a client that supports influxdb v1.
This code changes the interface for the HTTP.WriteService in order to be
a bit more general, and I wrote a WriteService that target InfluxDB v1

Signed-off-by: Gianluca Arbezzano <gianarb92@gmail.com>

### DIFF
--- a/http/write_handler.go
+++ b/http/write_handler.go
@@ -209,17 +209,101 @@ type postWriteRequest struct {
 	Precision string
 }
 
+// WriteServiceV1 sends data over HTTP to influxdb v1 via line protocol.
+type WriteServiceV1 struct {
+	Addr               string
+	Database           string
+	Rp                 string
+	Precision          string
+	Consistency        string
+	Username           string
+	Password           string
+	InsecureSkipVerify bool
+}
+
+var _ platform.WriteService = (*WriteServiceV1)(nil)
+
+// convertPrecisionToInfluxDB1 is an helper function that converts the SI units
+// used in Platform to what it is used for InfluxDB
+func convertPrecisionToInfluxDB1(precision string) string {
+	switch precision {
+	case "ns":
+		return "n"
+	case "us":
+		return "u"
+	default:
+		return precision
+	}
+}
+
+func (s *WriteServiceV1) Write(ctx context.Context, r io.Reader) error {
+	precision := s.Precision
+	if precision == "" {
+		precision = "ns"
+	}
+	consistency := s.Consistency
+	if consistency == "" {
+		consistency = "one"
+	}
+
+	if !models.ValidPrecision(precision) {
+		return fmt.Errorf("invalid precision")
+	}
+
+	precision = convertPrecisionToInfluxDB1(precision)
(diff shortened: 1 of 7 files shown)


http/write_handler.go
```diff
@@ -188,17 +188,88 @@ type postWriteRequest struct {
 	Precision string
 }
 
+// WriteService sends data over HTTP to influxdb v1 via line protocol.
```
goller: Should be WriteServiceV1 sends ...

http/write_handler.go
```diff
@@ -188,17 +188,88 @@ type postWriteRequest struct {
...
+type WriteServiceV1 struct {
```
goller: Also, add interface conformance check:

```go
var _ platform.WriteService = (*WriteServiceV1)(nil)
```

http/write_handler.go
```diff
@@ -188,17 +188,88 @@ type postWriteRequest struct {
...
+	Consistency        string
+	Username           string
+	Password           string
+	InsecureSkipVerify bool
+}
+
+func (s *WriteServiceV1) Write(ctx context.Context, r io.Reader) error {
+	precision := s.Precision
+	if precision == "" {
+		precision = "ns"
```
goller: Influx 1.x didn't use SI units consistently.  nanoseconds was `n` and microseconds was `u`
Because we want uniformity in the interface and hide the underlying inconsistency you'll need to add a conversion of `ns` and `us` to `n` and `u` after models.ValidPrecision()

http/write_handler.go
```diff
@@ -188,17 +188,88 @@ type postWriteRequest struct {
...
+	req.Header.Set("Content-Type", "text/plain; charset=utf-8")
+	req.Header.Set("Content-Encoding", "gzip")
+
+	params := req.URL.Query()
+	params.Set("db", s.Database)
+	params.Set("rp", s.Rp)
+	params.Set("precision", s.Precision)
+	params.Set("consistency", consistency)
+	if s.Username != "" {
+		params.Set("username", s.Username)
```
goller: Rather than set query parameters, you should set basic auth headers like this:

```go
req.SetBasicAuth(s.Username, s.Password)
```

Here is some helper code I wrote for chronograf that encompasses basic auth and JWT:

https://github.com/influxdata/platform/blob/master/chronograf/influx/authorization.go#L49

gianarb:
Do you think the interface has sense @goller ? It means that for every combination of bucket/org you need a different WriterService.

mark-rushakoff:

1. Download the patch from GitHub, by adding .patch to the end of the pull request URL, e.g. wget -P /tmp https://github.com/influxdata/platform/pull/1988.patch to download 1988.patch to /tmp (so that you don't accidentally check in 1988.patch in your influxdb directory).
1. Use git-am to apply the commits in the patch: git am /tmp/1988.patch. Then push the branch to GitHub and open a PR like you would normally.

PR #1989, opened 2018-12-17
Add support for prefix tombstone entries

### LINKED ISSUE #1833: Create prefix-based tombstone entries
All series keys contain a measurement comprising an org and bucket id. If one wished to remove all keys for a bucket or org, then rather than generating tombstone entries for every single series key, one entry could be created, containing the appropriate prefix and time range.

If one were entirely removing all org or bucket data, the time range would be maximal.

To enable this we need to allow tombstone entries to represent a prefix key.

e-dard:
Closes #1833.

This PR adds support for prefix-based tombstone keys, along with a new API to creating them. Prefix keys are to be used when the intention is to delete all series keys matching a prefix, which commonly might be an organisation or an oganisaton+bucket combination.

In these cases we can use a single prefix key rather than adding a tombstone entry for each and every matching key.

_What was the problem?_

The existing `tsm1` engine does not provide support for prefix base tombstone keys.

_What was the solution?_

For the `v4` tombstone version we use a `32 bit` value to represent the length of a tombstone entry's key. We don't need that many bits, so we reduce it to 24 bits. This leaves us with 8 spare bits when encoding a tombstone entry.

We use the first of these 8 bits to indicate if the encoded entry contains a normal key or a prefix key.

### DIFF
--- a/tsdb/tsm1/tombstone.go
+++ b/tsdb/tsm1/tombstone.go
@@ -1,5 +1,45 @@
 package tsm1
 
+/*
+Tombstone file format:
+
+╔═══════════════════════════════════════════Tombstone File════════════════════════════════════════════╗
+║ ┌─────────────┐┌──────────────────────────────────────────────────────────────────────────────────┐ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │   Header    ││                                                                                  │ ║
+║ │   4 bytes   ││                                Tombstone Entries                                 │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ │             ││                                                                                  │ ║
+║ └─────────────┘└──────────────────────────────────────────────────────────────────────────────────┘ ║
+╚═════════════════════════════════════════════════════════════════════════════════════════════════════╝
+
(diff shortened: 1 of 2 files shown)


tsdb/tsm1/tombstone.go
```diff
@@ -84,6 +133,40 @@ func (t *Tombstoner) WithObserver(obs FileStoreObserver) {
 	t.obs = obs
 }
 
+// AddPrefix adds a prefix-based tomstone key.
```
benbjohnson: `s/tomstone/tombstone/`

tsdb/tsm1/tombstone.go
```diff
@@ -84,6 +133,40 @@ func (t *Tombstoner) WithObserver(obs FileStoreObserver) {
...
+func (t *Tombstoner) AddPrefix(key []byte) error {
+	return t.AddPrefixRange(key, math.MinInt64, math.MaxInt64)
+}
+
+// AddPrefixRange adds a prefix-based tomstone key with an explicit range.
```
benbjohnson: Another reference to https://en.wikipedia.org/wiki/Tom_Stone_(magician)

e-dard:
@benbjohnson addressed in 688a31a88

tsdb/tsm1/tombstone.go
```diff
@@ -309,18 +337,18 @@ func (t *Tombstoner) prepareV4() error {
 		os.Remove(tmp.Name())
 	}
 
-	// Copy the existing v4 file if it exists
+	// Copy the existing v5 file if it exists
 	f, err := os.Open(t.tombstonePath())
 	if !os.IsNotExist(err) {
```
zeebo: shouldn't this be "if err == nil"? like if it's some random I/O error, we shouldn't try to read and copy it...

e-dard:
@zeebo fixed in 9d8114ef8

PR #1993, opened 2018-12-17
fix(http): pass user service to handler New function

### LINKED ISSUE #1987: owner/member endpoints are timing out
Introduced by https://github.com/influxdata/platform/pull/1486:

`UserService` is being used in the `NewTaskHandler` method for `BucketService`, `DashboardService`, `OrganizationService` and `TaskService` before it's been assigned.

zhulongcheng:
Closes #1987

### DIFF
--- a/http/api_handler.go
+++ b/http/api_handler.go
@@ -72,30 +72,26 @@ func NewAPIHandler(b *APIBackend) *APIHandler {
 	h.SessionHandler.SessionService = b.SessionService
 	h.SessionHandler.Logger = b.Logger.With(zap.String("handler", "basicAuth"))
 
-	h.BucketHandler = NewBucketHandler(b.UserResourceMappingService, b.LabelService)
+	h.BucketHandler = NewBucketHandler(b.UserResourceMappingService, b.LabelService, b.UserService)
 	h.BucketHandler.BucketService = b.BucketService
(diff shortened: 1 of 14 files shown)


zhulongcheng:
Hi @jademcgough , Do you mind giving a review?

http/api_handler.go
```diff
@@ -72,30 +72,26 @@ func NewAPIHandler(b *APIBackend) *APIHandler {
 	h.SessionHandler.SessionService = b.SessionService
 	h.SessionHandler.Logger = b.Logger.With(zap.String("handler", "basicAuth"))
 
-	h.BucketHandler = NewBucketHandler(b.UserResourceMappingService, b.LabelService)
+	h.BucketHandler = NewBucketHandler(b.UserResourceMappingService, b.LabelService, b.UserService)
```
desa: What do you think about having a config struct for each of these handlers? So that when we want to add more services to a new handler we don't have to modify the function signature?

zhulongcheng: Great idea!  💯

zhulongcheng: Can I  finish those changes in a new PR?  A config for each handler.

desa: That works for me 👍

PR #2005, opened 2018-12-17
fix(storage): convert the storage table interface to use arrow buffers

jsternberg:
The table interface was modified to expose the arrow buffers. The
storage table has now been converted to use this interface with the same
fixes so that it exposes arrow buffers.

The influxql package has also been updated to use the `DoArrow` method
from the `flux.Table` interface.

PR #2008, opened 2018-12-18
fix(arrow): stub out Arrow support

nathanielc:
This change stubs out DoArrow with an erroring implementation. We will follow up with #2005 to fully implement it. For now nothing is calling DoArrow so this is safe.


nathanielc:
Merging even with dead tests, as https://github.com/influxdata/platform/pull/2009 fixes the broken test. The final integrations will be fixed in https://github.com/influxdata/platform/pull/2002

storage/reads/gen.go
```diff
@@ -4,4 +4,3 @@ package reads
 //go:generate env GO111MODULE=on go run github.com/benbjohnson/tmpl -data=@array_cursor.gen.go.tmpldata response_writer.gen.go.tmpl
 //go:generate env GO111MODULE=on go run github.com/benbjohnson/tmpl -data=@array_cursor.gen.go.tmpldata stream_reader.gen.go.tmpl
 //go:generate env GO111MODULE=on go run github.com/benbjohnson/tmpl -data=@array_cursor.gen.go.tmpldata stream_reader_gen_test.go.tmpl
-//go:generate stringer -type=readState -trimprefix=state
```
jsternberg: We can't merge this with removing this. `stringer` doesn't seem to work with go modules yet so this causes a problem, but we can't merge this as-is.

nathanielc:
Closing in favor of #2005

PR #2009, opened 2018-12-18
fix(http): remove /api/v2/query/plan

nathanielc:
BREAKING CHANGE: The API endpoint exposed internal details about the implementation of
the logical and physical plans. Those data structures need to be free to
change without breaking the API. For now since the endpoint is unused it
is being removed. A similar API may be added back in that correctly
abstracts away the internal details of plans.


nathanielc:
Merging even with dead tests, as #2008 fixes the broken test. The final integrations will be fixed in #2002

wolffcm:
My changes on `flux` master branch is what caused `TestFluxHandler_postFluxPlan` to fail here.  Sorry about that.  I should have run my changes against `platform` so I didn't block the upgrade---I will be sure to test against platform and idpe moving forward.

PR #2020, opened 2018-12-18
Create/update telegraf plugin at verify step of onboarding

### LINKED ISSUE #1976: telegraf config should be created at the beginning of going to the verify step
you shouldnt have to go through all of the configs to get to the verify step. when going to the verify, it should create the config instead of creating it at the end of the previous step and loading it in the verify one.

AlirieGray:
Closes #1976

_What was the problem?_
Previously, if the user selected streaming data source in the onboarding wizard, then a telegraf plugin was created during the configuration step. If one had already been created, a new one was instantiated anyway. 

_What was the solution?_
Move the creation of the telegraf plugin to the verify step. Also, check if there is already a telegraf plugin and update the existing one if so. (The update endpoint still needs to be updated on the backend.)

  - [ ] Rebased/mergeable
  - [ ] Tests pass


AlirieGray:
NOTE: The update endpoint still needs to be fixed on the backend. Currently, an error is logged to the console whenever the update method is called, but the tests are passing.

ui/src/onboarding/actions/dataLoaders.ts
```diff
@@ -239,7 +241,28 @@ export const createTelegrafConfigAsync = (authToken: string) => async (
...
+  let body = {
+    name: 'new config',
+    agent: {collectionInterval: DEFAULT_COLLECTION_INTERVAL},
+    plugins,
+  }
+
+  if (telegrafConfigsFromServer.length) {
+    const id = _.get(telegrafConfigsFromServer, '0.id', '')
+
+    await updateTelegrafConfig(id, body)
```
121watts: Should we notify the user if this update failed?

Issue #1928, opened 2018-12-13
Fix styling for array fields in plugin config forms


AlirieGray:
This is the current input field for string array fields in plugin configurations, using the `TagInput` component:

<img width="324" alt="screen shot 2018-12-13 at 10 47 33 am" src="https://user-images.githubusercontent.com/15273162/49960306-b34ccc80-fec4-11e8-9f13-3563961614aa.png">

We may want to style this component, or use a different input component in `ArrayFormElement` other than `TagInput`

PR #2029, opened 2018-12-18
Remove id prop from input component

Palakp41:
Removed id prop from input component. A component with form inputs had id as empty string was causing issues on non-unique ids. Added id to input prop in the last pr for fixing #1928 but now id is not necessary.


Palakp41:
@alexpaxton 

I changed and fixed it! Thanks :)

PR #2049, opened 2018-12-19
feat(session): platform errors

### LINKED ISSUE #1295: convert /signin /signout endpoint error to platform error
https://github.com/influxdata/platform/issues/897

leodido:
Closes #1295 

_Convert existing session services and tests to `platform.Error`s_

  - [ ] Rebased/mergeable
  - [x] Tests pass


session.go
```diff
@@ -20,7 +35,7 @@ type Session struct {
 // Expired returns an error if the session is expired.
 func (s *Session) Expired() error {
 	if time.Now().After(s.ExpiresAt) {
-		return fmt.Errorf("session has expired")
+		return fmt.Errorf(ErrSessionExpired)
```
kelwang: i am curious, shall we add a new code for expired

leodido: Atm I've chosen to wrap this error within a `platform.Error` with the default code in places where it is used ([for example here](https://github.com/influxdata/platform/pull/2049/files#diff-974b92c64e58571e6552314d0d698a9bR48)).

However, usually session expired are signaled with a 401 HTTP status code (I know this is another domain) plus a custom header (`WWW-Authenticate`), as RFC2617 [suggests](https://tools.ietf.org/html/rfc2617#section-1.2).
Therefore independently of creating a new (and specific) error code we at least (probably) know to what it should match to at HTTP side.

PR #2057, opened 2018-12-19
ci: ensure go.sum matches original file from checkout

### LINKED ISSUE #2056: ci: module cache isn't restored correctly on circle
See [this recent gotest build from a merge to master](https://circleci.com/gh/influxdata/platform/10346):

The restore module cache stage prints `No cache is found for key: platform-gomod-4N9tsxql49j0GtH+MZTpneDgvUgVvPOtIQocq+rxsts=`

But when we attempt to save the cache, we see output with a different checksum: `Skipping cache generation, cache already exists for key: platform-gomod-+fY8eoFk+iLFoFpL4gS4dKm2BBkiInFwHEZWJJNcdTY=`.

I haven't looked back in the build history to identify when this behavior was introduced.

It looks like the `GOFLAGS=-mod=readonly -p=4` should have prevented go.mod from being edited, but maybe there's a case where go.sum still changes? Or maybe the `GOFLAGS` environment variable doesn't propagate to one of the subcommands?

mark-rushakoff:
In go1.11, go mod vendor can modify go.sum when go.mod contains replace
directives. This behavior has been fixed in go1.12, so just work around
it for now.

Fixes #2056.


BondAnthony:
@mark-rushakoff I wonder if this could show up in other places. I'll have to look around for this.

PR #2074, opened 2018-12-20
chore(CircleCI): turn off deploy on commits to master

nathanielc:


nathanielc:
@BondAnthony Question, the current cron job builds at UTC midnight, which is still a time of active development for some on the west coast.  Should we move the time back a few hours so that its more likely to represent a whole unit of work? Some time before Europe wakes up but after the west coast is done.

BondAnthony:
Do we want to run `all` for each build, validation everything actually builds? I was thinking about creating a new target called `build: all` and having Circle call that instead of `nightly`

Yes we may want to move the time back, UTC-7?

nathanielc:
@BondAnthony Sounds good to me. We want to at least build it even if we do not deploy it. Makes sense.

PR #2078, opened 2018-12-20
Fix issue forcing sign in on page refresh

chnn:
Fixes an issue where a user would have to sign in again on every page refresh.


alexpaxton:
YAY

PR #2080, opened 2018-12-20
chore(Makefile): fix issue with go run not having modules enabled

nathanielc:


nathanielc:
@alexpaxton This is the fix for your issue.

PR #2085, opened 2018-12-20
Label properties

### LINKED ISSUE #1954: label need a description

jademcgough:
Closes #1954

Adds a "properties" bucket to labels, which can be used to store arbitrary key/value pairs for the UI.

### DIFF
--- a/bolt/label.go
+++ b/bolt/label.go
@@ -185,21 +185,17 @@ func (c *Client) updateLabel(ctx context.Context, tx *bolt.Tx, l *platform.Label
 
 	label := ls[0]
 
-	properties := make(map[string]string)
-
-	for k, v := range label.Properties {
-		properties[k] = v
+	if label.Properties == nil {
+		label.Properties = make(map[string]string)
 	}
 
 	for k, v := range upd.Properties {
 		if v == "" {
-			delete(properties, k)
+			delete(label.Properties, k)
 		} else {
-			properties[k] = v
+			label.Properties[k] = v
 		}
 	}
-
-	label.Properties = properties
 
 	if err := label.Validate(); err != nil {
 		return nil, &platform.Error{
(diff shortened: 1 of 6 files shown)


bolt/label.go
```diff
@@ -185,10 +185,22 @@ func (c *Client) updateLabel(ctx context.Context, tx *bolt.Tx, l *platform.Label
 
 	label := ls[0]
 
-	if upd.Color != nil {
-		label.Color = *upd.Color
+	properties := make(map[string]string)
+
+	for k, v := range label.Properties {
```
desa: What do you think about doing something like

```go
if label.Properties == nil {
   label.Properties = make(map[string]string)
}
```

this way we can avoid the unnecessary map copy? It's definitely not needed, so feel free to address/ignore it as you see fit.

http/swagger.yml
```diff
@@ -6574,8 +6574,7 @@ components:
...
           type: string
-        color:
-          type: string
-          description: A six-character hex color.
-          pattern: '^([A-Fa-f0-9]{6})$'
-          example: 'ffb3b3'
+        properties:
+          type: object
+          description: Key/Value pairs associated with this label. Keys can be removed by sending an update with an empty value.
+          example: {color: ffb3b3, description: "this is a description"}
```
goller: In this example color should be double quoted.

label.go
```diff
@@ -32,9 +29,9 @@ type LabelService interface {
 }
 
 type Label struct {
-	ResourceID ID     `json:"resource_id"`
-	Name       string `json:"name"`
-	Color      string `json:"color"`
+	ResourceID ID                `json:"resource_id"`
```
goller: I think this should be `resourceID` ... (I'm still not sure if I decided correctly so long ago to go camel case rather than snake case for this API)

PR #2108, opened 2018-12-21
fix(http): change /query to use org/orgID

### LINKED ISSUE #2106: api: /api/v2/query uses `organization` and `organizationID` parameter but swagger has `org`
https://github.com/influxdata/platform/blob/master/http/requests.go#L11-L14

This should be `org` and `orgID`.

Additionally, orgID needs to be added to swagger:
https://github.com/influxdata/platform/blob/master/http/swagger.yml#L2489-L2493

goller:
Closes #2106 

We should use `org` and `orgID` rather than `organization` and `organizationID` as HTTP query parameters to `/api/v2/query`

_What was the problem?_
Documentation in swagger had `org`.

_What was the solution?_
Update docs and add tests to show org/organizationID,

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/requests_test.go
+++ b/http/requests_test.go
@@ -2,7 +2,6 @@ package http
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
@@ -39,7 +38,10 @@ func Test_queryOrganization(t *testing.T) {
 								ID: platform.ID(1),
 							}, nil
 						}
-						return nil, fmt.Errorf("unknown ID")
+						return nil, &platform.Error{
+							Code: platform.EInvalid,
+							Msg:  "unknown org name",
+						}
 					},
 				},
 			},
@@ -69,7 +71,10 @@ func Test_queryOrganization(t *testing.T) {
 								Name: "org1",
 							}, nil
 						}
-						return nil, fmt.Errorf("unknown org name")
+						return nil, &platform.Error{
+							Code: platform.EInvalid,
+							Msg:  "unknown org name",
+						}
 					},
 				},
 			},
--- a/http/requests_test.go
+++ b/http/requests_test.go
@@ -1,0 +1,90 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"net/http"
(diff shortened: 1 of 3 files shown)


http/requests_test.go
```diff
@@ -0,0 +1,90 @@
...
+								ID: platform.ID(1),
+							}, nil
+						}
+						return nil, fmt.Errorf("unknown ID")
+					},
+				},
+			},
+		},
+		{
+			name:    "bad id returns errorn",
```
leodido: typo (`errorn) ?

http/requests_test.go
```diff
@@ -0,0 +1,90 @@
...
+				r:   httptest.NewRequest(http.MethodPost, "/api/v2/query?org=org1", nil),
+				svc: &mock.OrganizationService{
+					FindOrganizationF: func(ctx context.Context, filter platform.OrganizationFilter) (*platform.Organization, error) {
+						if *filter.Name == "org1" {
+							return &platform.Organization{
+								ID:   platform.ID(1),
+								Name: "org1",
+							}, nil
+						}
+						return nil, fmt.Errorf("unknown org name")
```
kelwang: use &platform.Error{
     Code: platform.EInvalid,
     Msg: "unknown org name",
}

http/requests_test.go
```diff
@@ -0,0 +1,90 @@
...
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := queryOrganization(tt.args.ctx, tt.args.r, tt.args.svc)
+			if (err != nil) != tt.wantErr {
```
kelwang: compare error code, and http status code

goller: Hey @kelwang, in this case this is a helper function that extracts information from the *http.Request, but, does not write a response:

https://github.com/influxdata/platform/blob/5297d1fee9f156af962700545b7ce0cc24a3be22/http/requests.go#L18

I'll check the error for its type.

PR #2109, opened 2018-12-21
feat(models): only use common SI time units for timestamp precision

### LINKED ISSUE #1959: remove the `u` option from write query precision path.

goller:
Closes #1959 

1. Use SI symbols `us` and `ns` rather than `u` and `n` for timestamp precision.
2. Remove `h` (hours) and `m` (minutes) from timestamp precision.

I don't think we need hours and minutes since the Unix epoch.

_What was the problem?_
We didn't have a consistent unit system for precision.

_What was the solution?
Use only `ns`, `us`, `ms`, and `s` of precision since the Unix epoch for timestamps.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -739,7 +739,6 @@ paths:
             enum:
               - ns
               - us
-              - u
               - ms
               - s
       responses:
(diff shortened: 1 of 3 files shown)


goller:
Hey @e-dard I'm going to ping you after the holidays about my removal of minutes and hours.  It's a bit aggressive to remove it, but, I figure this could be clean slate.

PR #2112, opened 2018-12-21
docs(http): remove unused org param from swagger

### LINKED ISSUE #2083: create bucket request is using org in a query parameters as well as in a body.
I just wanted to know why create bucket API is using `org` in both places: URL query parameters and request body. Swagger doc says: https://github.com/influxdata/platform/blob/master/http/swagger.yml#L2509-L2539
I tried both cases and both of them are working.
I am not an expert, but since my job is to question everything I am just curios - wouldn't it be better to either use query parameters or request body for POST requests?
In this particular case, I can pass `?org=NAME` as a parameter as well as a body request. If both names are used in a request, then the one in the body has the preference. And looks like the `org` name in the `parameters` is being ignored. 
Would it be fair to say that:
- do not use `?org=NAME` in the url query parameters.
- Make `organization` or `organizationID` a required key so only one of them is required. (I am not sure if it is possible to make either one required). 
- Update swagger doc to not use `org` in the query parameters.

zhulongcheng:
Closes #2083

_What was the problem?_
_What was the solution?_

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -113,12 +113,6 @@ paths:
       summary: Create a telegraf config
       parameters:
           - $ref: '#/components/parameters/TraceSpan'
-          - in: query
-            name: org
-            description: specifies the organization of the resource
-            required: true
-            schema:
-              type: string
       requestBody:
         description: telegraf config to create
         required: true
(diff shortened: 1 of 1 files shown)


desa:
@goller Will we need these? Or are we going to pass these things in explicitly on the request body? I think the request body makes more sense.

goller:
@desa @zhulongcheng We should pass them in the request body.

zhulongcheng:
🌈

Issue #1302, opened 2018-11-05
convert /write endpoint error to platform error


kelwang:
https://github.com/influxdata/platform/issues/897

PR #2130, opened 2018-12-21
fix(http): convert write endpoint error

kelwang:
Closes https://github.com/influxdata/platform/issues/1302

### DIFF
--- a/http/write_handler.go
+++ b/http/write_handler.go
@@ -32,7 +32,9 @@ type WriteHandler struct {
 }
 
 const (
-	writePath = "/api/v2/write"
+	writePath            = "/api/v2/write"
+	errInvalidGzipHeader = "gzipped HTTP body contains an invalid header"
+	errInvalidPrecision  = "invalid precision; valid precision units are ns, us, ms, and s"
 )
 
 // NewWriteHandler creates a new handler at /api/v2/write to receive line protocol.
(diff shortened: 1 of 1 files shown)


http/write_handler.go
```diff
@@ -56,7 +56,11 @@ func (h *WriteHandler) handleWrite(w http.ResponseWriter, r *http.Request) {
 		var err error
 		in, err = gzip.NewReader(r.Body)
 		if err != nil {
-			EncodeError(ctx, errors.Wrap(err, "invalid gzip", errors.InvalidData), w)
+			EncodeError(ctx, &platform.Error{
+				Code: platform.EInvalid,
```
goller: I think it should have a message such as, "gzipped HTTP body contains an invalid header"

http/write_handler.go
```diff
@@ -172,7 +181,11 @@ func decodeWriteRequest(ctx context.Context, r *http.Request) (*postWriteRequest
 	}
 
 	if !models.ValidPrecision(p) {
-		return nil, errors.InvalidDataf("invalid precision")
+		return nil, &platform.Error{
+			Code: platform.EInvalid,
+			Op:   "http/decodeWriteRequest",
+			Msg:  "invalid precision",
```
goller: perhaps add something to the Msg such as:
"invalid precision; valid precision units are ns, us, ms, and s"

http/write_handler.go
```diff
@@ -205,7 +218,11 @@ func (s *WriteService) Write(ctx context.Context, orgID, bucketID platform.ID, r
 	}
 
 	if !models.ValidPrecision(precision) {
-		return fmt.Errorf("invalid precision")
+		return &platform.Error{
+			Code: platform.EInvalid,
+			Op:   "http/Write",
+			Msg:  "invalid precision",
```
goller: perhaps the same error message here

Issue #2140, opened 2018-12-24
Use a config struct to construct all API handlers


zhulongcheng:
From @desa 's [idea/suggestion](https://github.com/influxdata/platform/pull/1993#discussion_r242705738).

Use a config struct to construct all API handlers.

Benefit:
When we want to add more services or config options to a new handler,
we don't have to modify the function signature and change args of function for all callers.

At present, there is already such a config struct exists, the [http.APIBackend](https://github.com/influxdata/platform/blob/master/http/api_handler.go#L34-L66).

Handlers:
 - [ ] BucketHandler        
 - [ ] UserHandler          
 - [ ] OrgHandler           
 - [ ] AuthorizationHandler 
 - [ ] DashboardHandler     
 - [ ] AssetHandler         
 - [ ] ChronografHandler    
 - [ ] ViewHandler          
 - [ ] SourceHandler        
 - [ ] MacroHandler         
 - [ ] TaskHandler          
 - [ ] TelegrafHandler      
 - [ ] QueryHandler         
 - [ ] WriteHandler         
 - [ ] SetupHandler         
 - [ ] SessionHandler

zhulongcheng:
```
// NewOrgHandler returns a new instance of OrgHandler.
func NewOrgHandler(mappingService platform.UserResourceMappingService,
	labelService platform.LabelService, userService platform.UserService) *OrgHandler {
	h := &OrgHandler{
		Router:                     NewRouter(),
		Logger:                     zap.NewNop(),
		UserResourceMappingService: mappingService,
		LabelService:               labelService,
		UserService:                userService,
	}

	h.HandlerFunc("POST", organizationsPath, h.handlePostOrg)
        // ...

	h.HandlerFunc("POST", organizationsIDMembersPath, newPostMemberHandler(h.UserResourceMappingService, h.UserService, platform.OrgResourceType, platform.Member))
        // ...

	return h
}
```

For `newGetMembersHandler` handler, how to new a custom-backend and pass to the handler?

`MemberBackend` as a field for `OrgBackend` or Make `MemberHander` independent of `OrgHander`?

desa:
What about something like the following

```go
// NewOrgHandler returns a new instance of OrgHandler.
func NewOrgHandler(mappingService platform.UserResourceMappingService,
	labelService platform.LabelService, userService platform.UserService) *OrgHandler {
	h := &OrgHandler{
		Router:                     NewRouter(),
		Logger:                     zap.NewNop(),
[...]

PR #2141, opened 2018-12-24
initialize api handlers by APIBackend

zhulongcheng:
Related #2140

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/api_handler_test.go
+++ b/http/api_handler_test.go
@@ -92,6 +92,7 @@ func NewMockAPIBackend() *APIBackend {
 		MacroService:               &mock.MacroService{},
 		BasicAuthService:           mock.NewBasicAuthService("", ""),
 		TaskService:                &mock.TaskService{},
+		SecretService:              mock.NewSecretService(),
 
 		// TODO impl mock service
 		//SourceService:                   mock.NewSourceService(),
--- a/http/api_handler_test.go
+++ b/http/api_handler_test.go
@@ -1,11 +1,13 @@
 package http
 
 import (
-	"go.uber.org/zap"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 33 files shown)


http/api_handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+		UserResourceMappingService: mock.NewUserResourceMappingService(),
+		LabelService:               mock.NewLabelService(),
+		DashboardService:           &mock.DashboardService{},
+		ViewService:                &mock.ViewService{},
+		MacroService:               &mock.MacroService{},
+		BasicAuthService:           mock.NewBasicAuthService("", ""),
+		TaskService:                &mock.TaskService{},
+		SecretService:              mock.NewSecretService(),
+
+		// TODO impl mock service
```
desa: Could you create an issue for these?

zhulongcheng: Of course.   #2145

zhulongcheng:
Rebased.

desa:
@zhulongcheng I really like this change, but when I started to implement it for some of the closed source pieces I hit a few panics. What do you think about having custom backends for each of the handlers. This way we know exactly which handlers require which services.

(Sorry for the early approval)

zhulongcheng:
@desa I think custom-backend is a better way to know that which services required by a handler.

What is the detail of those panics? custom-backend is an effective way to avoid/reduce panics?

desa:
👍 

> What is the detail of those panics? custom-backend is an effective way to avoid/reduce panics?

I could chase down each of the panics, but they were the result of the service not being set on the `APIBackend` and to figure out which things needed to be set I had to look at the constructor. I could have gone through and fixed the specific panic, but I think it might be easier if each handler had its own specific backend for the constructor. What do you think?

PR #2165, opened 2018-12-31
Label responses

jademcgough:
This PR:
- adds labels to all requests that return tasks or dashboards
- fixes label POST endpoint to receive a JSON object in the body that describes the label (previously, it ignored the body and just used the label name from the path)


http/dashboard_service.go
```diff
@@ -293,7 +299,7 @@ func (d getDashboardsResponse) toPlatform() []*platform.Dashboard {
 	return res
 }
 
-func newGetDashboardsResponse(dashboards []*platform.Dashboard) getDashboardsResponse {
+func newGetDashboardsResponse(dashboards []*platform.Dashboard, ctx context.Context, labelService platform.LabelService) getDashboardsResponse {
```
desa: *super nit* what do you think about having `ctx` as the first argument?

jademcgough: 👍

PR #2170, opened 2019-01-02
fix(gather): add write interface

kelwang:
Closes https://github.com/influxdata/platform/issues/1957


gather/metrics.go
```diff
@@ -1,9 +1,21 @@
...
+	"fmt"
+	"io"
+
 	"github.com/gogo/protobuf/proto"
+	"github.com/influxdata/platform"
 )
 
+// Collected is the struct including metrics and other requirements.
+type Collected struct {
+	OrgID        platform.ID  `json:"org_id"`
```
desa: JSON struct tags should be golang style camel cased

gather/metrics.go
```diff
@@ -13,6 +25,59 @@ type Metrics struct {
...
+			return nil, ew.err
+		}
+		if mi < len(ms)-1 && len(ms) > 1 {
+			ew.write("\n")
+		}
+	}
+	return buf, nil
+}
+
+type errBufWriter struct {
```
desa: What's the purpose of this? Why introduce another struct when you could just check errors in the `Reader` method?

gather/metrics.go
```diff
@@ -40,6 +105,11 @@ var metricTypeValue = map[string]int32{
 	"HISTOGRAM": 4,
 }
 
+// Valid returns whether the metrics is valid.
```
desa: What does it mean for a metric to be valid?

gather/storage.go
```diff
@@ -1,17 +1,43 @@
...
 	"encoding/json"
 	"fmt"
+	"time"
 
+	"github.com/influxdata/platform"
 	"github.com/influxdata/platform/nats"
 	"go.uber.org/zap"
 )
 
+// PlatformWriterStorage will use the writer interface to record the metrics
```
desa: Comment is missing a `.`

gather/storage.go
```diff
@@ -1,17 +1,43 @@
...
+type PlatformWriterStorage struct {
```
desa: What do you think about calling this a `PlatformWriter`, `PlatformStorage`, or `PlatformRecorder`?

gather/storage.go
```diff
@@ -1,17 +1,43 @@
...
+		collected.OrgID,
+		collected.BucketID,
+		r,
+	)
+
+	return nil
+}
+
 // Storage stores the metrics of a time based.
 type Storage interface {
```
desa: What do you think about renaming this to `Recorder` or `MetricRecorder`?

scraper.go
```diff
@@ -15,12 +15,12 @@ const (
...
-	Name       string      `json:"name"`
-	Type       ScraperType `json:"type"`
-	URL        string      `json:"url"`
-	OrgName    string      `json:"org"`
-	BucketName string      `json:"bucket"`
+	ID       ID          `json:"id,omitempty"`
+	Name     string      `json:"name"`
+	Type     ScraperType `json:"type"`
+	URL      string      `json:"url"`
+	OrgID    ID          `json:"org,omitempty"`
```
desa: Shouldn't the struct tag be `orgID` and `bucketID`?

gather/metrics.go
```diff
@@ -1,9 +1,21 @@
...
+	"bytes"
+	"fmt"
+	"io"
+
 	"github.com/gogo/protobuf/proto"
+	"github.com/influxdata/platform"
 )
 
+// Collected is the struct including metrics and other requirements.
+type Collected struct {
```
desa: The name `Collected` seems a bit off to me. At the surface, I'm not sure what a `gather.Collected` is just by the name.

gather/metrics.go
```diff
@@ -1,9 +1,21 @@
...
 import (
+	"bytes"
+	"fmt"
+	"io"
+
 	"github.com/gogo/protobuf/proto"
+	"github.com/influxdata/platform"
 )
 
+// Collected is the struct including metrics and other requirements.
```
desa: What are the other requirements for?

gather/metrics.go
kelwang: Yeah, this is an idea from go blogs, to prevent multiple err checking. https://blog.golang.org/errors-are-values

gather/metrics.go
kelwang: Yeah, I think i need some suggestions on this. This is basically metrics plus orgID, and bucketID.

gather/metrics.go
desa: Ah, interesting idea. I think I like that pattern, but I don't know that it's a perfect fit here. In any case, could you add a comment that links to that blog post and describe the pattern briefly?

kelwang: 👍

gather/metrics.go
desa: Is there a reason we're not using [models.Point](https://github.com/influxdata/influxdb/blob/38b07bbaab8616d26ce9e71eb9a7e62fd6463c6e/models/points.go#L65-L141) for `Metrics`?

Maybe a temporary work around could be `MetricsCollection`? or something of that sort?

kelwang: 👍  I like the models.Point actually, thanks

desa:
@kelwang is this ready for another review?

kelwang:
> @kelwang is this ready for another review?

not yet, still working it

gather/metrics.go
```diff
@@ -40,6 +82,11 @@ var metricTypeValue = map[string]int32{
 	"HISTOGRAM": 4,
 }
 
+// Valid returns whether the metrics type is valid.
+func (x MetricType) Valid() bool {
+	return x >= 0 && x < 5
```
leodido: I think using `MetricType` constants here would be preferable.

What do you think?

gather/metrics.go
```diff
@@ -1,18 +1,60 @@
 package gather
 
 import (
+	"bytes"
+	"io"
+	"time"
+
+	"github.com/influxdata/platform/models"
+
```
e-dard: Please remove this newline.

gather/metrics.go
```diff
@@ -1,18 +1,60 @@
...
 	Fields    map[string]interface{} `json:"fields"`
-	Timestamp int64                  `json:"timestamp"`
+	Timestamp time.Time              `json:"timestamp"`
 	Type      MetricType             `json:"type"`
 }
 
+// MetricsSlice is a slice of Metrics
+type MetricsSlice []Metrics
+
+// Reader will load the metrics slice string to reader
```
e-dard: How about something like:

> Reader returns an io.Reader that enumerates the metrics.
> 
> All metrics are allocated into the underlying buffer.

gather/recorder.go
```diff
@@ -0,0 +1,62 @@
...
+	Logger   *zap.Logger
+}
+
+// Process consumes job queue, and use recorder to record.
+func (h *RecorderHandler) Process(s nats.Subscription, m nats.Message) {
+	defer m.Ack()
+	collected := new(MetricsCollection)
+	err := json.Unmarshal(m.Data(), &collected)
+	if err != nil {
+		h.Logger.Error(fmt.Sprintf("recorder handler process err: %v", err))
```
e-dard: Alternatively: `h.Logger.Error("recorder handler error", zap.Error(err))`

e-dard:
I don't have any context on this PR (I was asked to review it), because the PR description isn't very descriptive. Further, #1957 is just a one liner.

Folks, can we please provide more background context in PR descriptions and issues? As the teams grow it's going to be even more important that we articulate proposed changes. You know what it's like when you start digging into other peoples' code 6 months down the line and there is nothing to explain why the chain was made 😄

gather/metrics.go
kelwang: yeah, good idea

gather/metrics.go
kelwang: 👍

gather/metrics.go
kelwang: > I don't have any context on this PR (I was asked to review it), because the PR description isn't very descriptive. Further, #1957 is just a one liner.
> 
> Folks, can we please provide more background context in PR descriptions and issues? As the teams grow it's going to be even more important that we articulate proposed changes. You know what it's like when you start digging into other peoples' code 6 months down the line and there is nothing to explain why the chain was made 😄

Hey @e-dard, thanks for the review. Scraper is the lib we could use to collect prometheus metrics. Currently by using Nats queue, we can successfully parse and collect the prometheus metrics. This pr is about how we store the metrics with the write interface. After merge of this pr, we plan to test it out in idpe.

kelwang:
> I don't have any context on this PR (I was asked to review it), because the PR description isn't very descriptive. Further, #1957 is just a one liner.
> 
> Folks, can we please provide more background context in PR descriptions and issues? As the teams grow it's going to be even more important that we articulate proposed changes. You know what it's like when you start digging into other peoples' code 6 months down the line and there is nothing to explain why the chain was made 😄

Hey @e-dard, thanks for the review. Scraper is the lib we could use to collect prometheus metrics. Currently by using Nats queue, we can successfully parse and collect the prometheus metrics. This pr is about how we store the metrics with the write interface. After merge of this pr, we plan to test it out in idpe.

gather/recorder.go
kelwang: 👍

gather/metrics.go
kelwang: 👍

mark-rushakoff:
Looks like this has been moved to influxdb properly. Thanks.

PR #2157, opened 2018-12-27
feat(auth): refactor authentication service

121watts:
Closes #2150
Closes #2151
Closes  #1881
Closes #2154
Closes #2155
Closes #2156

- Simplifies permission actions to `write` and `read`
- Added optional ID for a Permission resource
- Removed ID from resource type string
- Added different resource types
- UserID pulled from token.  No longer required in `POST`

PR #2185, opened 2019-01-02
Fix create authorization command

### LINKED ISSUE #2169: cli: authorization permissions invalid ID
### Observations

When trying to create a new authorization with `--delete-user` permissions I get an error:

```
influx  auth create -u yolo7  --delete-user
Error: invalid ID
Usage:
  influx auth create [flags]

Flags:
      --create-user                grants the permission to create users
      --delete-user                grants the permission to delete users
  -h, --help                       help for create
      --read-bucket stringArray    bucket id
  -u, --user string                user name (required)
      --write-bucket stringArray   bucket id

Global Flags:
      --host string    HTTP address of Influx (default "http://localhost:9999")
      --local          Run commands locally against the filesystem
  -t, --token string   API token to be used throughout client calls

invalid ID
```

### Analysis

Client side (cli) while creating a new `Permission` the function `NewPermission` doesn't set any ID on the newly created permission here:
https://github.com/influxdata/platform/blob/3429e8d0c6753e7f1dab6b080397580a5ba94205/authz.go#L166-L173

This leads to `ErrInvalidID` when the permission is validated server side:

https://github.com/influxdata/platform/blob/3429e8d0c6753e7f1dab6b080397580a5ba94205/http/auth_service.go#L218-L224

desa:
Closes #2169 

Previously, authorizations required the users id to be set explicitly on the authorization. As of https://github.com/influxdata/platform/pull/2157 the user id is retrieved off of the authorization or session used in the http request.

We now allow users to specify an explicit user when creating authorizations. If not id is provided, the user from the authorizer will be used.

Additionally, we now need to pass the organization id when creating an authorization.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/cmd/influx/authorization.go
+++ b/cmd/influx/authorization.go
@@ -23,7 +23,8 @@ var authorizationCmd = &cobra.Command{
 
 // AuthorizationCreateFlags are command line args used when creating a authorization
 type AuthorizationCreateFlags struct {
-	org string
+	user string
+	org  string
 
 	createUserPermission bool
 	deleteUserPermission bool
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
Looks like this has been moved to influxdb properly. Thanks.

PR #2187, opened 2019-01-02
Fix/org view

### LINKED ISSUE #2182: Use dashboard modified date in the UI
as a platform user
when i view the list of dashboards 
then i would like to see the last time they were updated

121watts:
Closes #2182


ui/src/dashboards/components/dashboard_index/Table.tsx
```diff
@@ -133,6 +128,13 @@ class DashboardsTable extends PureComponent<Props & WithRouterProps, State> {
     return null
   }
 
+  private get datedDashboards(): DatedDashboard[] {
+    return this.props.dashboards.map(d => ({
+      ...d,
+      modified: new Date(d.meta.updatedAt),
```
ebb-tide: is this not meant to be time since update? which is silly but.

PR #2196, opened 2019-01-03
Remove pipe operator when 'from' is clicked on explorer

### LINKED ISSUE #2116: clicking the from function should not add a `|>` in the script builder
when you add a function in the "Inputs" section, it should not add a `|>` in front of it

Palakp41:
Closes #2116 

Add parameter to the onclick for toolbar functions that conditions on the function name

### DIFF
--- a/ui/src/shared/components/flux_functions_toolbar/FluxFunctionsToolbar.tsx
+++ b/ui/src/shared/components/flux_functions_toolbar/FluxFunctionsToolbar.tsx
@@ -71,10 +71,14 @@ class FluxFunctionsToolbar extends PureComponent<Props, State> {
     this.setState({searchTerm})
   }
 
-  private handleUpdateScript = (funcExample: string) => {
+  private handleUpdateScript = (funcName: string, funcExample: string) => {
     const {activeQueryText, onSetActiveQueryText} = this.props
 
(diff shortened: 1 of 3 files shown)


ui/src/shared/components/flux_functions_toolbar/FluxFunctionsToolbar.tsx
```diff
@@ -71,10 +71,14 @@ class FluxFunctionsToolbar extends PureComponent<Props, State> {
     this.setState({searchTerm})
   }
 
-  private handleUpdateScript = (funcExample: string) => {
+  private handleUpdateScript = (funcName: string, funcExample: string) => {
     const {activeQueryText, onSetActiveQueryText} = this.props
 
-    onSetActiveQueryText(`${activeQueryText}\n  |> ${funcExample}`)
+    if (funcName === 'from') {
```
chnn: How about:
```suggestion
    if (funcName.startsWith('from')) {
```
This way it will handle the `fromRows` function too, which I think also should not have a pipe prepended. I think there will be other similar `from` functions in the future, like `fromCSV`, `fromSQL`, etc.

Palakp41: Okay, makes sense! Thanks :)

PR #2198, opened 2019-01-03
Update range function example constant to include now() instead of now

### LINKED ISSUE #2190: update the example in the range function to use now() instead of now
range example is currently `|> range(start: -15m, stop: now)`, should be   `|> range(start: -15m, stop: now())`, and documentation in the function dropdown should be changed as well to reflection

Palakp41:
Closes #2190

Updated the range function in the constants to include ()


Palakp41:
Sorry about this, I accidentally clicked wrong buttons.

PR #2226, opened 2019-01-04
Create dummy data for members and add filter box to the header

### LINKED ISSUE #1378: Add the ability to filter org members
as an org manager
when i view the list of members for a specific org
then i would like the ability to filter those members

Palakp41:
Closes #1378 

Created dummy data for members since there is no data in db for members. Added the filter box on the header of members tab and implemented the functionality to update state when user enters text
kapture 2019-01-03 at 16 31 49

### DIFF
--- a/ui/src/organizations/components/Members.tsx
+++ b/ui/src/organizations/components/Members.tsx
@@ -1,13 +1,17 @@
 // Libraries
-import React, {PureComponent} from 'react'
+import React, {PureComponent, ChangeEvent} from 'react'
 
 // Components
-import {ComponentSize, EmptyState} from 'src/clockface'
+import {ComponentSize, EmptyState, IconFont, Input} from 'src/clockface'
 import MemberList from 'src/organizations/components/MemberList'
 import FilterList from 'src/shared/components/Filter'
 
 // Types
 import {ResourceOwner} from 'src/api'
+
+// Constants
+import {resouceOwner} from 'src/organizations/dummyData'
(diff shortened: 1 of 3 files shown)


ui/src/organizations/components/Members.tsx
```diff
@@ -25,20 +29,40 @@ export default class Members extends PureComponent<Props, State> {
...
     )
   }
 
+  private handleFilterBlur = (e: ChangeEvent<HTMLInputElement>): void => {
+    this.setState({searchTerm: e.target.value})
+  }
+
+  private handleFilterChange = (e: ChangeEvent<HTMLInputElement>): void => {
+    this.setState({searchTerm: e.target.value})
+  }
```
ebb-tide: if these two functions are identical, maybe you could make one function and pass it to both onChange and onBlur?

ui/src/organizations/dummyData.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const resouceOwner: ResourceOwner[] = [
+  {
+    id: '1',
+    name: 'John',
+    status: User.StatusEnum.Active,
+    links: {
+      self: '/api/v2/users/1',
+    },
+    role: ResourceOwner.RoleEnum.Owner,
```
ebb-tide: nice use of enums!

ui/src/organizations/components/Members.tsx
Palakp41: Made it one function. Thanks for noticing!

PR #2227, opened 2019-01-04
Add timeRangeStart variable to queries as preamble

### LINKED ISSUE #2118: dashboardTime isn't working in platform
we need the equivalent of dashboardTime in platform. We should call it `timeRangeStart`.

bthesorceror:
Closes #2118 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/apis/v2/query.ts
+++ b/ui/src/shared/apis/v2/query.ts
@@ -1,3 +1,5 @@
+import _ from 'lodash'
+
 import Deferred from 'src/utils/Deferred'
 
 import {InfluxLanguage} from 'src/types/v2/dashboards'
@@ -19,8 +21,21 @@ interface XHRError extends Error {
 export const executeQuery = async (
   url: string,
   query: string,
-  language: InfluxLanguage = InfluxLanguage.Flux
+  language: InfluxLanguage = InfluxLanguage.Flux,
+  variables?: {[key: string]: string}
 ): Promise<ExecuteFluxQueryResult> => {
+  let preamble = ''
+
+  if (variables && language === InfluxLanguage.Flux) {
+    preamble = _.reduce(
+      variables,
+      (result, value, name) => `${result}${name} = ${value}\n`,
+      ''
+    )
+  }
+
+  query = `${preamble}${query}`
+
   // We're using `XMLHttpRequest` directly here rather than through `axios` so
   // that we can poll the response size as it comes back. If the response size
   // is greater than a predefined limit, we close the HTTP connection and
@@ -133,10 +148,11 @@ export const executeQuery = async (
 }
 
 export const executeQueries = async (
-  queries: URLQuery[]
+  queries: URLQuery[],
+  variables?: {[key: string]: string}
 ): Promise<ExecuteFluxQueryResult[]> => {
   const promise = Promise.all(
-    queries.map(({url, text, type}) => executeQuery(url, text, type))
+    queries.map(({url, text, type}) => executeQuery(url, text, type, variables))
   )
 
   return promise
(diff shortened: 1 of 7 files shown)


ui/src/shared/apis/v2/query.ts
```diff
@@ -19,8 +21,21 @@ interface XHRError extends Error {
...
+
+  if (variables) {
+    preamble = _.reduce(
+      variables,
+      (result, value, name) => `${result}${name} = ${value}\n`,
+      ''
+    )
+  }
+
+  query = `${preamble}${query}`
```
chnn: The language can also be `InfluxLanguage.InfluxQL`, in which case appending the `preamble` will result in a broken query.

ui/src/shared/utils/queryBuilder.ts
```diff
@@ -15,6 +15,18 @@ const WINDOW_INTERVALS = {
...
 }
 
+export const timeRangeVariables = (
+  timeRange: TimeRange
+): {[key: string]: string} => {
+  const result: {[key: string]: string} = {}
+
+  result.timeRangeStart = timeRange.lower
+    .replace('now()', '')
+    .replace(/\s/g, '')
```
chnn: some `TimeRange` objects have `duration` field. In particular, all the time ranges that are set by using the `TimeRangeDropdown` do:

https://github.com/influxdata/platform/blob/master/ui/src/shared/constants/timeRanges.ts#L9

Would it be better to use the duration and do something like
```
result.timeRangeStart = `-${timeRange.duration}`
```
instead of the regexes?

Kinda sketchy since `duration` is an optional field on the `TimeRange` interface...

ui/src/shared/components/cells/Cell.tsx
```diff
@@ -111,19 +112,17 @@ class CellComponent extends Component<Props> {
...
     return (
-      <ViewComponent
-        view={view}
-        onZoom={onZoom}
-        timeRange={timeRange}
-        autoRefresh={autoRefresh}
-        manualRefresh={manualRefresh}
-        onEditCell={onEditCell}
-      />
+      <Conditional isRendered={viewStatus !== RemoteDataState.Done}>
```
chnn: Is this much of an improvement over
```
{viewStatus !== RemoteDataState.Done && (
  <ViewComponent />
)}
```
For someone whose famliar with JSX and the `{bool && conditionalRender}` idiom it might be one more thing to learn about the codebase.

I know this is a nitpicky style thing, so I'm fine if you leave this in.

ui/src/shared/apis/v2/query.ts
```diff
@@ -19,8 +21,21 @@ interface XHRError extends Error {
...
+
+  if (variables && language === InfluxLanguage.Flux) {
+    preamble = _.reduce(
+      variables,
+      (result, value, name) => `${result}${name} = ${value}\n`,
+      ''
+    )
+  }
+
+  query = `${preamble}${query}`
```
chnn:
```suggestion
  query = `${preamble}\n${query}`
```
I am a huuuge fan of line breaks everywhere all of the time :)

ui/src/shared/components/RefreshingView.tsx
```diff
@@ -65,6 +66,7 @@ class RefreshingView extends PureComponent<Props, State> {
         submitToken={submitToken}
         queries={this.queries}
         key={manualRefresh}
+        variables={{...timeRangeVariables(timeRange)}}
```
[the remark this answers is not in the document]
bthesorceror: Just left it there as an indication other variables can go here as well, I was imagining there will be other variable to add soon.

ui/src/shared/components/cells/Cell.tsx
bthesorceror: I find this easier to parse personally, the way it was before always breaks my flow parsing code.

ui/src/shared/apis/v2/query.ts
bthesorceror: I left it without because when there is a preamble the preamble has a line break and when there is no preamble then the query is unchanged.

PR #2251, opened 2019-01-04
feat(query/functions/tests): end to end testing is configured to run …

aanthony1243:
…against a live storage engine.

all tests use a unique bucket based on the test file name.  copied all tests over from flux repo

added failing tests to skip list for now

  - [X] Rebased/mergeable
  - [X] Tests pass


query/functions/tests/end_to_end_test.go
```diff
@@ -0,0 +1,312 @@
...
+		testName := filename[:len(filename)-len(".flux")]
+		reason, skip := skipTests[testName]
+		t.Run(filename, func(t *testing.T) {
+			if skip {
+				t.Skip(reason)
+			}
+			if qs, err := ioutil.ReadFile(filename); err != nil {
+				return
+			} else {
+				// Query server to ensure write persists.
```
wolffcm: (nit) This `if` statement doesn't need an `else` since it will just return if `err` is not `nil`.  Removing one level of indention might make this method easier to read.

query/functions/tests/end_to_end_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+				req := &query.Request{
+					OrganizationID: l.Org.ID,
+					Compiler:       lang.FluxCompiler{Query: qy},
+				}
+				if _, err := l.FluxService().Query(ctx, req); err != nil {
+					t.Fatal(err)
+				}
+
+				// quirk: our execution engine doesn't guarantee the order of execution for disconnected DAGS
```
wolffcm: This quirk could be really confusing and I think it should be considered a bug.  Not sure what the fix is.  Maybe the move to an IR will fix it though?

wolffcm: Actually, won't executing the query twice cause the bucket to have twice as much data as expected sometimes?

query/functions/tests/end_to_end_test.go
```diff
@@ -0,0 +1,312 @@
...
+					})
+					if err != nil {
+						if assertionErr, ok := err.(AssertionError); ok {
+							t.Error(assertionErr)
+						} else {
+							t.Fatal(err)
+						}
+					}
+				}
+				if err := r.Err(); err != nil {
```
wolffcm: So if `assertEquals` finds differences in what was wanted and what was gotten, is this where we expect to find the error, or do we find it above from the call to `Tables().Do()`?

query/functions/tests/testdata/aggregate_window.flux
```diff
@@ -0,0 +1,39 @@
+inData = "
```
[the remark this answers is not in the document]
aanthony1243: current plan is to build this in as a flux package.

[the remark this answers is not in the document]
aanthony1243: current plan is to build this in as a flux package.

query/functions/tests/end_to_end_test.go
aanthony1243: from tables.Do

query/functions/tests/end_to_end_test.go
aanthony1243: if the input had no explicit time stamps this would be true.  perhaps this will explain some of the failures.  we will see.

query/functions/tests/end_to_end_test.go
```diff
@@ -0,0 +1,312 @@
...
+// loadData is a function that's referenced in all the transformation tests.  
+// it's registered here so that we can register a different loadData function for 
+// each platform/binary.  
+testLoadStorageHelper = (csv, bucket, org) => {fromCSV(csv: csv) |> to(bucket: bucket, org: org) return from(bucket: bucket) }
+testLoadMem = (csv) => fromCSV(csv: csv)
+`
+
+var skipTests = map[string]string{
+	// TODO(adam) determine the reason for these test failures.
+	"cov":                      "Reason TBD",
```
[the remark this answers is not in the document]
aanthony1243: the test also is flaky, I think. I have seen a few cases fail that previously succeeded.  but I want to see if we can get this merged to make it easier to attack multiple issues in parallel.

PR #2256, opened 2019-01-05
Do not truncate time fields in raw view

### LINKED ISSUE #2117: Improve column sizing in raw data view (time columns should not truncate if possible)
ensure that by default, you can see the entire timestamp for all time, _start, _stop columns in the raw data view

bthesorceror:
Closes #2117 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/RawFluxDataGrid.tsx
+++ b/ui/src/shared/components/RawFluxDataGrid.tsx
@@ -1,0 +1,116 @@
+import React, {PureComponent, CSSProperties} from 'react'
+import _ from 'lodash'
+import {Grid} from 'react-virtualized'
+
+const ROW_HEIGHT = 27
+const MIN_COLUMN_WIDTH = 150
+const TIME_COLUMN_WIDTH = 300
+
+interface Props {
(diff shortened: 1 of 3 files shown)


ui/src/shared/components/RawFluxDataGrid.tsx
```diff
@@ -0,0 +1,116 @@
...
+  scrollTop: number
+}
+
+interface State {
+  headerRows: number[]
+}
+
+export default class extends PureComponent<Props, State> {
+  public static getDerivedStateFromProps(nextProps: Props): Partial<State> {
+    const headerRows = _.reduce(
```
chnn: I think this is going to be called on every scroll event in the table, and might be pretty slow since it has to iterate through the entire data set. Could it be memoized via `memoizeOne` instead? So it would only change when the data set changes.

bthesorceror: tested on a fairly large dataset and didn't see a performance hit so will merge and investigate on Monday, thanks @chnn

PR #2279, opened 2019-01-07
Update swagger models that have labels

bthesorceror:
- [x] Rebased/mergeable
  - [x] Tests pass


http/cur_swagger.yml
```diff
@@ -3670,6 +3670,8 @@ components:
           type: string
         organizationID:
           type: string
+        labels:
+          $ref: "#/components/schemas/Labels"
```
OfTheDelmer: is the schema for labels up to date? I added one a while back while labels was still developing, so I'm just curious.

bthesorceror: Looks like it.

PR #2281, opened 2019-01-07
Fix issue plotting independent series

chnn:
Attempting to plot 2 time series with independent time values was resulting in visual discontinuities.

The issue was that dygraphs expects non-existant time values for a time series to be `null`, not `undefined`. 

I also refactored the `DygraphTransformation` component to memoize the parsing computation.

Before:

screen shot 2019-01-07 at 12 01 12 pm

After:

screen shot 2019-01-07 at 12 00 55 pm


ui/src/shared/components/DygraphTransformation.tsx
```diff
@@ -13,27 +16,14 @@ interface Props {
...
-    super(props)
-
-    this.state = {
-      labels: [],
-      dygraphsData: [],
-      seriesDescriptions: [],
-    }
-  }
+class DygraphTransformation extends PureComponent<Props> {
+  private fluxTablesToDygraph = memoizeOne(fluxTablesToDygraph)
```
Palakp41: Nice!

Issue #2172, opened 2019-01-02
Table graph field options sort errors


OfTheDelmer:
When using a source that errors and returns no data the table graphs errors due to a sorting on undefined.

![kapture 2019-01-02 at 11 26 07](https://user-images.githubusercontent.com/4994741/50601132-4e601300-0e81-11e9-80c0-a71c75b7ab9a.gif)

Steps to reproduce:

- querying autogen source with 

```
from(bucket: "telegraf/autogen")
  |> range(start: -1h)
  |> filter(fn: (r) => r._measurement == "cpu")
```

- responds with the following

```
#group,true,true
#datatype,string,string
#default,,
,error,reference
,"failed to execute query: failed to initialize execute state: could not find bucket ""telegraf/autogen""",
```

- the table graph then ends up with the following props

<img width="1118" alt="screen shot 2019-01-02 at 11 37 15 am" src="https://user-images.githubusercontent.com/4994741/50601672-e27eaa00-0e82-11e9-9bce-a38750078a31.png">

The error message I see is the following:

```
errors.tsx:50 Error: Sort cannot be performed
    at sortTableData (tableGraph.ts:220)
    at TableGraphTransform.transformTableData (tableGraph.ts:259)
    at TableGraphTransform.result [as memoizedTableTransform] (memoize-one.esm.js:24)
    at TableGraphTransform.render (TableGraphTransform.tsx:43)
    at finishClassComponent (react-dom.development.js:13538)
    at updateClassComponent (react-dom.development.js:13501)
    at beginWork (react-dom.development.js:14090)
    at performUnitOfWork (react-dom.development.js:16416)
    at workLoop (react-dom.development.js:16454)
    at renderRoot
```

Issue #2101, opened 2018-12-20
hoverTime is not highlighted in Table Graphs. 


ebb-tide:
As a user with editor privileges or more,
when I hover over another graph in the dashboard, 
I see that the row of the table that corresponds to the time I am hovering over, is highlighted, and the table scrolls to the appropriate row/column

ebb-tide:
May be related to #2094

PR #2286, opened 2019-01-07
feat(ui): add table graph hover time

### LINKED ISSUE #2094: Table Graph does not set hoverTime
As a user with editor privileges or more, 
when I mouseover a table graph in dashboards
I see that hover time is set,
and crosshairs are visible on other graphs in view.

OfTheDelmer:
Closes #2094 #2101 #2172 

Update the table graph in VEO to handle flux tables without a `_time` field and update hover time when viewing dashboards.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/dashboards/utils/tableGraph.ts
```diff
@@ -300,3 +300,44 @@ export const transformTableData = (
...
+}
+
+/**
+ * Get absolute mili seconds between a unix ms time number and iso utc string
+ *
+ * @param unixMs
+ * @param isoTime
+ */
+export const getUnixISODiff = (unixMs: number, isoTime: string | number) => {
+  return Math.abs(unixMs - new Date(isoTime).valueOf())
```
OfTheDelmer: This currently works, but I think I might need a separate issue to make this more robust

PR #2305, opened 2019-01-08
Add explicit submit button for query builder

### LINKED ISSUE #2294: add a submit button to the explorer
as a query explorer
when i select an option in the data explorer
then i would like to explicitly submit the query
then i would like some indication that the query is different than the option i have selected

bthesorceror:
Closes #2294

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/reducers/v2/timeMachines.ts
+++ b/ui/src/shared/reducers/v2/timeMachines.ts
@@ -509,7 +509,7 @@ export const timeMachineReducer = (
 
         draftQuery.builderConfig.tags[index].values = values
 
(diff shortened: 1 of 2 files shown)


ui/src/shared/reducers/v2/timeMachines.ts
```diff
@@ -509,7 +509,7 @@ export const timeMachineReducer = (
 
         draftQuery.builderConfig.tags[index].values = values
 
-        buildAndSubmitActiveQuery(draftState)
+        buildActiveQuery(draftState)
```
ischolten: noice

PR #2306, opened 2019-01-08
Display token on my account token overlay

### LINKED ISSUE #2285: Add the ability to display the token in the UI
as a new user
when i view the list of tokens
then i would like the ability to view the text of the token
then i would like the ability to copy the text of the token to my clipboard

Palakp41:
Closes #2285

Extracted copy to clipboard button from onboarding to it's separate component which takes in token or any snippet to be copied. CopyToken component also takes in notify prop to show notification that the text was copied.

feature-2285

### DIFF
--- a/ui/src/onboarding/components/verifyStep/TelegrafInstructions.tsx
+++ b/ui/src/onboarding/components/verifyStep/TelegrafInstructions.tsx
@@ -1,20 +1,15 @@
 // Libraries
-import React, {PureComponent, MouseEvent} from 'react'
+import React, {PureComponent} from 'react'
 import _ from 'lodash'
-import CopyToClipboard from 'react-copy-to-clipboard'
 
 // Decorator
 import {ErrorHandling} from 'src/shared/decorators/errors'
 
 // Components
-import {Button, ComponentSize, ComponentColor} from 'src/clockface'
+import CopyText from 'src/shared/components/CopyText'
 
 // Types
 import {NotificationAction} from 'src/types'
-import {
-  copyToClipboardSuccess,
-  copyToClipboardFailed,
-} from 'src/shared/copy/notifications'
 
 export interface Props {
   notify: NotificationAction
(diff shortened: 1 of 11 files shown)


ui/src/onboarding/components/verifyStep/TelegrafInstructions.tsx
```diff
@@ -47,61 +41,13 @@ class TelegrafInstructions extends PureComponent<Props> {
...
-                customClass="wizard-step--body-copybutton"
-                size={ComponentSize.Small}
-                color={ComponentColor.Default}
-                titleText="copy to clipboard"
-                text="Copy"
-                onClick={this.handleClickCopy}
-              />
-            </CopyToClipboard>
-          </p>
+          <CopyText copyText={configScript} notify={notify} />
```
OfTheDelmer: This component really helped remove a lot of boilerplate

ui/src/shared/components/CopyText.tsx
```diff
@@ -0,0 +1,67 @@
...
+// Actions
+import {NotificationAction} from 'src/types'
+import {
+  copyToClipboardSuccess,
+  copyToClipboardFailed,
+} from 'src/shared/copy/notifications'
+
+export interface Props {
+  copyText: string
+  notify?: NotificationAction
```
OfTheDelmer: Do we want this to be optional?

ui/src/shared/components/CopyText.tsx
```diff
@@ -0,0 +1,67 @@
...
+  private handleClickCopy = (e: MouseEvent<HTMLButtonElement>) => {
+    e.stopPropagation()
+    e.preventDefault()
+  }
+
+  private handleCopyAttempt = (
+    copiedText: string,
+    isSuccessful: boolean
+  ): void => {
+    const {notify} = this.props
```
OfTheDelmer: We may want a guard here if `notify` is not provided.

ui/src/shared/components/CopyText.tsx
Palakp41: Actually I will make it required prop. It makes sense to see a notification when text is copied.

ui/src/onboarding/components/verifyStep/TelegrafInstructions.tsx
Palakp41: Yes, it did :)

Issue #1837, opened 2018-12-11
Implement DROP BUCKET


e-dard:
This is a meta issue covering all the things we need to do to implement a performant `DROP BUCKET` feature.

Until we rework how we store TSM data on disk, multiple buckets can physically share TSM files. This makes enforcing retention periods, or dropping an entire bucket relatively expensive. In both cases they require a range-based delete, and can result in the creation of many tombstone entries for high cardinality data.

From a storage perspective, we need to expose an API that allows the API tier to provide a bucket and org name, which, when called, will apply an appropriate delete command to the `FileStore`, `Cache` and `WAL`.

```go 
func (e *Engine) DeleteBucket(orgID, bucketID platform.ID) error {
    ...
}
```

 As part of this work, we need to consider dealing with the following:

  - [ ] Expose `DeleteBucket` API in `storage.Engine`.
  - [x] Currentlly, tombstones are applied during startup, so that TSM data that has been deleted is not queryable. This process is expensive, and is causing performance issues when scaled out to dropping entire an buckets' worth of TSM data. We need to improve how tombstones are applied. This work will be tracked in #https://github.com/influxdata/platform/issues/1595
  - [x] In Platform, *all* series keys are prefixed with a measurement comprising an org and bucket id. Therefore, dropping a bucket involves deleting all series keys sharing the same prefix. Rather than creating tombstone entries for each and every key, we could create a new prefix-based tombstone entry, for a set of keys belonging to the same prefix. This work is tracked in https://github.com/influxdata/platform/issues/1833
  - [x] Teach the `tsm` engine to delete series keys based on series key prefixes. This work is tracked in #1836
  - [ ] Wire up retention enforcer to use new prefix-delete approach.

PR #2313, opened 2019-01-08
DeletePrefix on the tsm1 engine

### LINKED ISSUE #1836: Teach the storage engine to apply prefix-based tombstone entries
This work relies on #1833 and #1595 being completed first.

Once we can represent prefix-based deletes via tombstone entries, we need to teach the storage engine to apply them.

zeebo:
This PR adds a `DeletePrefix` method to the tsm1 engine and below. It is responsible for removing all of the data for keys matching a given prefix from every tsm file, the WAL and the cache, and removing from the index any series that were fully removed due to the delete.

It is only designed to work for the "happy path" right now in that there are still race conditions if writes happen while the delete is ongoing, data structures can get out of sync on disk in the event of process failures, and there are some algorithmic inefficiencies in determining which keys to delete.

It adds a tree to keep track of which prefixes have tombstones in the index. It does this by chunking on 8-byte sections of a prefix (since most prefixes will be multiples of 8 bytes) and falling back to 1-byte sections for the rest. Rather than store it in a flat `[N]*child` like a radix tree might, it uses a `map[key]*child` making it sparse. This allows for efficient queries of the form "what prefixes match this given key?" which is used when determining the set of tombstones for a key.

Closes #1836.
Updates #1837.


tsdb/tsm1/engine_delete_prefix.go
```diff
@@ -0,0 +1,86 @@
...
+
+	fs, err := e.index.RetainFileSet()
+	if err != nil {
+		return err
+	}
+	defer fs.Release()
+
+	// TODO(jeff): are the query language values still a thing?
+	// Min and max time in the engine are slightly different from the query language values.
+	if min == influxql.MinTime {
```
e-dard: IIRC the query-land min and max values are later/earlier than the engine ones by 1ns to allow for some special sentinal values.

tsdb/tsm1/reader_prefix_tree.go
```diff
@@ -0,0 +1,85 @@
...
+func newPrefixTree() *prefixTree {
+	return &prefixTree{
+		short: make(map[byte]*prefixTree),
+		long:  make(map[prefixTreeKey]*prefixTree),
+	}
+}
+
+func (p *prefixTree) Append(prefix []byte, values ...TimeRange) {
+	if len(prefix) >= prefixTreeKeySize {
+		lookup := *(*prefixTreeKey)(unsafe.Pointer(&prefix[0]))
```
benbjohnson: This `unsafe` makes me a little nervous but I can't believe there's not an easier way to convert slice data to an array. I suppose you could use a `copy()`.

tsdb/tsm1/engine_delete_prefix.go
e-dard: @zeebo correct. We don't.

tsdb/tsm1/reader_prefix_tree.go
zeebo: Copy is always a function call whereas this little trick compiles down to a single 64bit load instruction. That being said, there's only like a 2% difference between the two, so definitely gonna drop the unsafe.

PR #2321, opened 2019-01-08
store label properties in a separate bucket

jademcgough:
Each label that is created has separate properties, but all label mappings with the same name should actually share the same properties.

This PR implements shared label properties by adding another boltDB bucket for them.

### DIFF
--- a/bolt/label.go
+++ b/bolt/label.go
@@ -218,7 +218,11 @@ func (c *Client) updateLabel(ctx context.Context, tx *bolt.Tx, l *platform.Label
 
 	label := ls[0]
 
-	if label.Properties != nil {
+	if label.Properties == nil {
+		label.Properties = make(map[string]string)
+	}
+
+	if upd.Properties != nil {
 		for k, v := range upd.Properties {
 			if v == "" {
 				delete(label.Properties, k)
--- a/bolt/label.go
+++ b/bolt/label.go
@@ -10,11 +10,15 @@ import (
 )
 
 var (
-	labelBucket = []byte("labelsv1")
+	labelBucket      = []byte("labelsv1")
+	labelPropsBucket = []byte("labelpropertiesv1")
 )
 
 func (c *Client) initializeLabels(ctx context.Context, tx *bolt.Tx) error {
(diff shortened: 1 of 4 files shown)


bolt/label.go
```diff
@@ -128,6 +154,13 @@ func labelKey(l *platform.Label) ([]byte, error) {
 	return key, nil
 }
 
+func labelPropKey(l *platform.Label) []byte {
```
desa: What happens when two labels for different resources have the same name? Will they collide here?

bolt/label.go
```diff
@@ -189,11 +222,22 @@ func (c *Client) updateLabel(ctx context.Context, tx *bolt.Tx, l *platform.Label
...
+				label.Properties[k] = v
+			}
+		}
+
+		p, err := json.Marshal(label.Properties)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := tx.Bucket(labelPropsBucket).Put(labelPropKey(l), p); err != nil {
```
desa: Is it not possible to update a labels name?

inmem/label_service.go
```diff
@@ -12,6 +12,10 @@ func encodeLabelKey(resourceID platform.ID, name string) string {
 	return path.Join(resourceID.String(), name)
 }
 
+func encodeLabelPropKey(name string) string {
```
desa: ditto the previous question about collisions on labels that have the same name across organizations.

bolt/label.go
jademcgough: I should add that as part of this PR... would just need to delete it from the bucket and re-add it with a new name

mark-rushakoff:

1. Download the patch from GitHub, by adding .patch to the end of the pull request URL, e.g. wget -P /tmp https://github.com/influxdata/platform/pull/2321.patch to download 2321.patch to /tmp (so that you don't accidentally check in 2321.patch in your influxdb directory).
1. Use git-am to apply the commits in the patch: git am /tmp/2321.patch. Then push the branch to GitHub and open a PR like you would normally.

PR #2330, opened 2019-01-08
Display Labels in Dashboards Index

### LINKED ISSUE #1882: Display labels on the dashboards list
as a user
when i view the list of dashboard
then i would like to see the labels associated with that dashboard

### LINKED ISSUE #2346: Display description in dashboard list

alexpaxton:
Closes #1882 
Closes #2346 

screen shot 2019-01-08 at 2 09 57 pm

Each dashboard can have up to 4 labels before it truncates with a tooltip
Descriptions display below the name of the dashboard

Also updated `cur_swagger` and regenerated the client.
Had to manually add `resourceID` to the labels in swagger

  - [x] Rebased/mergeable
  - [ ] Tests pass


alexpaxton:
Not really sure how the snapshot test is failing. The timestamps are from mock data and shouldn't be changing

ui/src/clockface/components/index_views/IndexList.scss
```diff
@@ -193,4 +193,25 @@
...
\ No newline at end of file
+}
+
+/*
+  Margin for "Name" column in tables
+  ------------------------------------------------------------------------------
+*/
+
+.index-list--description {
+  user-select: none;
```
mavarius: What if someone wants to copy the description?

alexpaxton: ahh probs a good call. I was gonna build in the editable portion in a dif branch in which case that might change

PR #2354, opened 2019-01-08
Polish query builder tab states

### LINKED ISSUE #2301: Smaller input on rename tabs
when you edit the names of the tab, the edit box should be smaller so it doesn't expand the tab

chnn:
Closes #2301

- Use a smaller input when renaming a tab that will not cause the tab to resize:

   screen shot 2019-01-08 at 3 08 40 pm

- Overflow text in a tab properly:

   screen shot 2019-01-08 at 3 08 50 pm

- Do not refresh builder when clicking the active tab name


ui/src/shared/components/TimeMachineQueries.tsx
```diff
@@ -102,10 +102,10 @@ const TimeMachineQueries: SFC<Props> = props => {
 }
 
 const mstp = (state: AppState) => {
-  const {draftQueries, activeQueryIndex} = getActiveTimeMachine(state)
+  const {draftQueries} = getActiveTimeMachine(state)
```
ischolten: nice!

PR #2367, opened 2019-01-09
Add org admin and member specific permissions

### LINKED ISSUE #2252: Add Organization Admin privileges during onboarding/session creation
As of https://github.com/influxdata/platform/pull/2157, authorizations are broken for tasks. Specifically, the structure of permissions has changed in such a way that the user was not given the appropriate permissions during the onboarding set up.

To fix this, we should add OrgAdmin privileges like so

```go
// OrgResources is the list of all known resource types for an organization.
var OrgResources = []Resource{
	BucketsResource,    // 1
	DashboardsResource, // 2
	SourcesResource,    // 4
	TasksResource,      // 5
}

// OrgAdminPermissions are the default permissions for an admin in an organization.
func OrgAdminPermissions(id ID) []Permission {
	ps := []Permission{}
	for _, r := range OrgResources {
		for _, a := range actions {
			ps = append(ps, Permission{Action: a, Resource: r, ID: &id})
		}
	}

	return ps
}
```

And these privileges should be added during onboarding.

Additionally, these privileges need to be added on session create.

Link https://github.com/influxdata/platform/issues/2249

desa:
Closes #2252 

As https://github.com/influxdata/platform/pull/2157 introduced a new structure for permissions, we now need to grant org level permissions for organization owned resources. Specifically, we introduces permissions for dashboards, tasks, sources, buckets, telegrafs, and users for organizations. The permission for these resources now contains an `ID` that links to the organization.

**Note:** We still need a way to describe the write/read actions for an individual task. This may imply the need to add an `OrganizationID *ID` to the permissions struct.
@mark-rushakoff @lyondhill thoughts on the note above.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -30,7 +30,7 @@ func (a *Authorization) Allowed(p Permission) bool {
 		return false
 	}
 
-	return allowed(p, a.Permissions)
+	return PermissionAllowed(p, a.Permissions)
 }
 
 // IsActive is a stub for idpe.
--- a/authz.go
+++ b/authz.go
@@ -29,9 +29,25 @@ type Authorizer interface {
 	Kind() string
 }
 
-func allowed(p Permission, ps []Permission) bool {
+// PermissionAllowed
+func PermissionAllowed(p Permission, ps []Permission) bool {
+	pID := ID(0)
+	if p.ID != nil {
+		pID = *p.ID
+		if !pID.Valid() {
+			return false
+		}
+	}
+
 	for _, perm := range ps {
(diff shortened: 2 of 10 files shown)


mark-rushakoff:
> If a user has access to write any task they will have access to write all tasks.

I think this is fine for now.

> One idea would to differentiate be to have WriteAction TasksResource OrgID for the ability to create new tasks in general and WriteAction TaskResource TaskID for a specific task.

This would also be fine for post-alpha.

PR #2372, opened 2019-01-09
refactor: rename DoArrow to Do

jsternberg:
See influxdata/flux#783 for details.


query/functions/outputs/to.go
```diff
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/influxdata/platform/tsdb"
 	"time"
 
+	"github.com/influxdata/platform/tsdb"
+
 	"sort"
```
nathanielc: This is still odd, can you move the github one down with the rest and the sort up?

jsternberg: GoLand really needs to fix this import business.

PR #1009, opened 2018-10-09
add telegraf.go

kelwang:
Closes #https://github.com/influxdata/platform/issues/944

   - [x] Rebased/mergeable

PR #2376, opened 2019-01-09
Scope telegraf configs to an org

### LINKED ISSUE #2282: Scope telegraf configs to an organization
Currently, telegraf configs are not scoped to an organization. This means that organizations cannot differentiate between their configs.

We should add `OrganizationID` to configs to support filtering configs by organization.

leodido:
Closes #2282

Addition of `OrganizationID` to `TelegrafConfig` struct.
Addition of `TelegrafConfigFilter` which wraps `UserResourceMappingFilter` with a `*OrganizationID`.
Refactoring accordingly `inmem`, `bolt` and `http` (which misses some important tests atm and it seems to not be really in sync with swagger).

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [ ] http/swagger.yml updated (if modified Go structs or API)


kelwang:
besides we don't have to store OrganizationID in the telegraf config, everything else looks great!
https://github.com/influxdata/platform/pull/1009#discussion_r223868245

Since it's already merged, i think we can simple create a pr to remove the organizationID from telegraf config, then it will be perfect

PR #2385, opened 2019-01-09
Feature/update task options

### LINKED ISSUE #1996: task: support for updating options without directly editing flux script
https://github.com/influxdata/flux/pull/387 introduced functionality that should enable editing options on an AST, and https://github.com/influxdata/flux/pull/383 should enable translating AST into source code again. My understanding is that both of those changes should be available in the version of flux that is currently on platform.

Following the [current swagger for tasks](https://github.com/influxdata/platform/blob/49d098346d43dc366cf2a6d15703f84c8d82e8aa/http/swagger.yml#L4444-L4505), we should support overwriting options `name`, `every`, `cron`, or `offset`. 

Task service tests will need to be updated. Edge cases that come to mind:

- If any option is set and Flux is set too, the request should be rejected (`not allowed to update script and override options at the same time` or similar error).
- If existing script has `every`, and user updates option `cron`, ensure that existing `every` option is cleared. And vice versa.
- It is an error for the user to provide `cron` and `every` options at the same time.

docmerlin:
Closes #1996

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -146,7 +146,7 @@ require (
 	golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f
 	golang.org/x/sys v0.0.0-20181205085412-a5c9d58dba9a
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2
-	golang.org/x/tools v0.0.0-20181221154417-3ad2d988d5e2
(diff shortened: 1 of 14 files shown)


mark-rushakoff:
These changes are on https://github.com/influxdata/influxdb/pull/10954 now.

